PR #270, opened 2013-09-17
mu4e-headers.el: add ido-completing-read to narrow func

stsquad:
This is more of a RFC pull request as there is probably a more elegant
way to go about this. For example if someone adds the from: tag
additional completions could be offered based on what users are seen
in the current search stream.


stig:
I would love something like this. I frequently try to narrow to messages from a particular mailing list but fail to type the whole list name correctly.

djcb:
This has gone stale; closing this...

PR #330, opened 2013-12-21
Foldable threads in header view

xaiki:
The implementation is pretty silly:

+ put a marker everytime we come across a message that has a
 'thread-level' of 0.
+ look for that marker, the line that has it is the thread header,
  create an overlay for it.
+ the thread header is given different faces depending of it's folded
  state
+ all lines after that marker up to another marker or end-of-file are
  the subthreads, they are given another overlay.
+ the subthreads block overlay is given a face to customize it's look.
+ we only set-up overlays on demand (after rendering) so that we don't
  slow header-view rendering.
+ as a bonus track, a function to recurse all thread groups is provided.

Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -229,6 +229,9 @@ followed by the docid, followed by `mu4e~headers-docid-post'.")
 (defconst mu4e~headers-docid-post (purecopy "\377")
   "Each header starts (invisibly) with the `mu4e~headers-docid-pre',
 followed by the docid, followed by `mu4e~headers-docid-post'.")
+(defconst mu4e~headers-thread-start (purecopy "\378")
+  "Each thread starts (invisibly) with the `mu4e~headers-thread-start'
+ marker")
 
 (defvar mu4e~headers-view-win nil
   "The view window connected to this headers view.")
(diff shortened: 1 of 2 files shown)


gburch:
+1

dakrone:
This would be great to have, especially for very long threads so they can easily be folded. Any idea what's up with this PR?

avar:
I did a trivial rebase of this in my avar-xaiki/thread-foldable-headers branch to solve the conflicts, but the code doesn't work anymore if it ever did. @xaiki, is this something you still use or can you notice what's broken?

avar:
Actually. It sort of works if you set `mu4e-use-fancy-chars nil`, but doesn't work at all without it. But it's still buggy. I.e. when you collapse a thread from the first message in the thread the pointer ends up on the line above it.

If you collapse it in a message after the first the pointer correctly ends up at the first message, but you have to press `n` twice to navigate to the next line. Other than that it seems to work.

avar:
Oh, and it doesn't work at all for collapsing subcomponents of the thread, which would be neat. It just collapses the whole thing or expands it entirely.

djcb:
I like the idea, but it seems a slightly different approach is needed - so closing this PR.

avar:
@djcb For people interested in this, could you elaborate on that please?

I don't fully understand the patch, but from what I see it's basically placing an invisible marker at the start & end of threads, and then when you press `<TAB>` it checks if you're within one of those and collapses the thread. It seems this hack is used for some existing things in the header view.

What sort of approach do you think makes sense instead?

B.t.w. just to clarify the ambiguities of text-based communication the above is not a rhetorical question. I have very little experience with elisp mode programming and haven't seen how this sort of thing is usually solved, but would be interested in your thoughts on this if I find some time to try to implement this.

dakrone:
I applied this patch locally to play around with it to see what it was like, and it seems to work for the most part. It's missing something where threads can be automatically folded, but that would be pretty easy to add I think.

@djcb I'm curious about the same thing as @avar, if there is any plan for this in the future or what can be changed to get it in. I'm happy to try and contribute to get it to whatever approach is desired. For me, this is the only thing gnus does better than mu4e.

PR #350, opened 2014-01-19
Adds function for editing messages in mu4e-view

odi:
Adds function and keybinding to ',' in mu4e-view mode which provides the
possibility to edit and save a message in the mu4e-view mode. The
message could be saved with C-c C-c. This keybinding is hard coded.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -522,6 +522,7 @@ FUNC should be a function taking two arguments:
       (define-key map "E" 'mu4e-compose-edit)
 
       (define-key map "." 'mu4e-view-raw-message)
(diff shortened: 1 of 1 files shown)


Chris00:
How is it different from _Resend_ in the _View_ menu?

odi:
The difference is that the function ``mu4e-edit-raw-message`` does not send/resend the message. It only changes the file to the new content in the filesystem.

The intention of this function was that I have to edit a message with a crippled header or a wrong subject line every now and then. So you can see this function as manual washing step after getting some messages. You can see it as counterpart to the function ``gnus-summary-edit-article`` in gnus.

PR #400, opened 2014-03-20
* bugfix for issue 164

pyhrr0:
This patch ensures threads get sorted based on their newest messages instead of the first one.


jhelberg:
issue 164 seems to be back. no?

Issue #102, opened 2012-11-27
resend a message


Popsch:
Often when I don't get an answer to an email, I just resend the email (instead of reply). It would be good to have this feature (aka gnus-summary-resend-message or gnus-summary-resend-bounced-mail).

zakkak:
@djcb any comments on this issue and #449 ?

djcb:
I've commented on #449; would be interesting to get that in, but it requires a bit of work still.

djcb:
There's `mu4e-compose-resend` now.

PR #449, opened 2014-06-27
* add initial bounce functionality

pasja:
Shameless recombination from http://notmuchmail.org/emacstips/#index17h2 and #350 

I think it solves #102 

Sadly i can't make it to save bounced message to the sent folder. I use gmail and set `(setq mu4e-sent-messages-behavior 'delete)` I localized the variable, but the problem remain unsolved.


djcb:
This would be useful, probably use 'B' as the binding for this and relegate mu4e-headers-search-bookmark-edit to "C-u b". 

However, indeed we'd need to solve the "Sent" issue, and I think we'd want completion for the address.

djcb:
In any case, we have "Resend" now (for a while already) which can be used for bouncing; so closing this.

Issue #542, opened 2014-12-16
Option for automatically reply-to-sender vs. reply-to-all


toothbrush:
Hey! Recent mutt convert here. I've been digging through the mu4e source code, but it seems like i cannot do what i want to do:

When pressing some key, say `R`, i want to automatically reply to the person in the `From:` field. When i press some other key, say `g`, i want to automatically reply-to-all. I don't so much like prompts, you see :)

In the meantime perhaps i will try and write a pull request, but i'm not sure that my ELisp-fu is powerful enough to do it The Right Way™.

Thanks for a great MUA!

mallet:
Having the option to reply to the sender only or to all without the prompt would be very welcomed, indeed. @djcb did not want to give a quick access to the reply-to-all, but it does not have to be available by default. Just a function to bind in .emacs would do.

I find the current approach annoying: when you activate "mu4e-compose-dont-reply-to-self", sometimes you get a prompt, sometimes not, depending on whether you were the only receiver. In addition, like @toothbrush, I am not particularly keen on prompts, especially for such a basic operation as replying to a message.

semente:
What a about a "reply to mailing list only" feature? Some email clients has this handy feature.

cpaulik:
@semente This is possible by using the functions from https://groups.google.com/forum/#!topic/mu-discuss/9ZbUu0wGxU0 but I think it should be included in mu4e.

semente:
thank you very much @cpaulik

djcb:
It's a conscious decision, having been the victim of too much reply-to-all. So, you need to take the millisecond/1 keypress to make a decision. Perhaps it can be automated with a keyboard macro, but I'll leave it as it is in mu4e.

Issue #373, opened 2014-02-16
Add the function mu4e-compose-reply-all


philc:
It would be helpful to have a single shortcut in headers mode for "reply all". mu4e doesn't expose this as a single function. If you invoke `mu4e-compose-reply-all`, a menu is shown and you must type another character to select "reply all".

Having it available as a single function makes it easy to bind, which is helpful for people migrating from Gmail, where Reply All is a single keystroke.

incandescentman:
Agreed.

sineer:
Please :+1:

incandescentman:
This would be a super useful function to add.

edumerco:
:+1: 

Maybe with r to reply and R to reply to all. 
Or r & a (mnemonic for "a"ll). :)

gaydenko:
@edumerco , `r` is used for `refile` (that is to archive) already.

I don't understand so massive new +1s. On `R` we already get a choice to reply to sender (`s`) or to all (`a`) when appropriate.

PR #543, opened 2014-12-17
Add reply-to argument to mu4e-compose-reply

zakkak:
Make mu4e-compose-reply take an optional argument that shows whether
the reply should be to all recipients or only to the sender.
#373 and #542 related


adam353:
This functionality would be incredibly useful to those of us who greatly desire separate single-key bindings for "reply" and "reply all" from *mu4e-view* (as mutt does).  Please consider merging this request.

zakkak:
@djcb Any comments on this?

Ambrevar:
Any update?

djcb:
Don't really want to change so much code to save one keystroke... so am going to close this old PR.

PR #555, opened 2015-01-06
Make the speedbar feature also work for sr-speedbar

xuhdev:
sr-speedbar uses speedbar as backend, but displays it in the same frame.

I'm also curious that how speedbar knows to load `mu4e-*-speedbar-buttons`? I didn't find anywhere in the code `mu4e-*-speedbar-buttons` is called. Sorry I'm very new to elisp.

### DIFF
--- a/mu4e/mu4e-speedbar.el
+++ b/mu4e/mu4e-speedbar.el
@@ -32,6 +32,7 @@
 ;;; Code:
 
 (require 'speedbar)
+(require 'sr-speedbar nil t)
 (require 'mu4e-vars)
 (require 'mu4e-headers)
 


djcb:
Hmmm, wouldn't want to unconditionally use sr-speedbar if the user happens to have it installed.

xuhdev:
There would be no error if not installed. That's why I put the last param of require to `t`

djcb:
My concern is actually that if you have sr-speedbar installed, this proposed change would unconditionally turn it on. I think it's better to leave that up to the user.

xuhdev:
I don't understand what would change, `speedbar` will remain what it does, `sr-speedbar` will do what `speedbar` has been defined to do. There is no way to enable this feature besides hacking the code and adding this line.

zakkak:
What @djcb says is that with your patch upstream, any user having `sr-speedbar` installed will be using `sr-speedbar`, by default, with mu4e.  I agree with Dirk that this is not a welcomed behavior.

xuhdev:
I think you have misunderstood something here. sr-speedbar cannot be used "by default" in this case -- you have to explicitly run the `M-x sr-speedbar-open` to use sr-speedbar, `M-x speedbar` still launches the regular speedbar. `sr-speedbar` is NOT something which would override speedbar though.

zakkak:
So why do you need it to load in `mu4e/mu4e-speedbar.el` and not somewhere in your emacs configuration before loading mu4e (or even after it)?  It works for me.

xuhdev:
@zakkak  Hmm, I submitted this patch simply because the way you suggested did not work for me. I just tried it, and it works now :/

zakkak:
Nice that you got it working :)

djcb:
Ah, @xuhdev is right about {{sr-speedbar}} not overtaking {{speedbar}}; I think it did at some point.. Anyway, great that we don't even need it now!

xuhdev:
How about adding it to wiki or doc? People with terminal emacs may be looking for it   :)

PR #556, opened 2015-01-08
mu-cmd-server.c: quote msgid in query

abend:
Hi.  I received some messages with spaces in the message id.  mu4e was having problems handling the messages when attempting the "move" server command to set flags when I viewed the messages.  The query to return all docids for the msgid was returning no results.  Surrounding the msgid with quotes in the mu query fixed the problem.

If this seems like the correct approach, feel free to merge the fix.  If you'd like to see it reworked in some way, I'd be happy to make another attempt.

Thanks for this amazing mailreader!


djcb:
Merged -- thanks!

PR #558, opened 2015-01-15
mu4e-vars: fix reference to obsolete mu4e-my-email-addresses

stsquad:
This variable has been replaced with mu4e-user-mail-address-list.

Signed-off-by: Alex Bennée <alex.bennee@linaro.org>


djcb:
Merged, thanks!

PR #563, opened 2015-01-19
Improve queue-related actions of the main view

DamienCassou:
- The queue-related actions of `mu4e~main-view-real` have been extracted
  to their own `mu4e~main-view-queue` method.
- The queue toggling action now has the word "currently" to easily.
  distinguish between the current state and the toggle (e.g., "currently
  queued" instead of just "queued").
- The "[f]lush queued emails" is only visible if there is at least 1
  email in the queue.
- The number of emails in the queue is visible in the flushing
  action (e.g., "[f]lush 10 queued emails").


djcb:
Merged -- thanks!

PR #566, opened 2015-01-28
Fix org-mu4e bug with HTML in text/plain part

cpbotha:
When sending org mode mail, both the text/html and text/plain parts
contained the HTML version of the mail, instead of just the text/html
part. This patch puts the raw-body (org format) in the text part, and
the html with replaced images in the html part.


djcb:
Merged -- thanks for the fix!

PR #574, opened 2015-02-03
mu4e: Added action to import attachment in diary. Better handle encoding while sending html to html2text.

jeansch:
Let import an attachment to the emacs diary file with [d]
Especialy usefull when org-agenda is setup to display diary !


djcb:
Nice -- thanks!

PR #575, opened 2015-02-05
Let not be stuck into *mu4e-output* after rendering html. Reduce by one levels of (with-temp-buffer.

jeansch:


djcb:
Merged -- thanks!

PR #579, opened 2015-02-11
Add option for saving multiple attachments into same directory.

halbtuerke:
There was a discussion at the `mu-discuss` Google group about [How to easily save all attachments?](https://groups.google.com/d/topic/mu-discuss/LpwtWZ1WCyg/discussion).

I have added an option to allow saving all attachments into the same directory when using `C-u e`. By default this option is deactivated to not mess with existing workflows.

As I'm a complete Elisp newbie it would be great if you could give me some hints to improve the code in case there's something wrong with my implementation.


djcb:
Oh, looks good - thanks, merged

PR #580, opened 2015-02-13
Make sure C-S-u and C-c C-u work in main view

DamienCassou:
These 2 shortcuts update mail index in both the headers and view modes
but not in the main view.


djcb:
Merged -- thanks!

Issue #280, opened 2013-09-30
Can't verify signature on encrypted mail


pstch:
If I receive a mail that is encrypted AND signed with PGP/Mime, I can see the decrypted version without problems.

However, the signature does not appear anywhere. The signature seems to be combined with the encrypted version, and when I decrypt it using gpg, it gives a signature confirmation.

    ~ ❯ mu view me
    ...
    [No body found; message has encrypted parts]

mu view --decrypt works well, but does not show anything about the decryption process. 

Decrypting using gpg will work, and it will show signature verification directly after decryption.

I think this is bug.

I use 0.9.8.4-3+deb7u1, with libgmime2.6.

cocreature:
So I finally spent a bit of time looking into this and I think I found the problem:

There are two ways to sign & encrypt messages using gpg/mime. One is the one you described where you first sign it and then encrypt it. I don't have any message here using that to verify it, but as you describe not having problems, I guess that works. The other way (see [rfc](https://tools.ietf.org/html/rfc3156#section-6.2)) is a combined mode which seems to be what my emacs and afaik enigmail by default as well are using. So I guess the latter is not taken into account.

I'm not very good with C so if anybody else wants to tackle this I would be very happy.

zakkak:
OK, I see.

So how do we know it is a signed message after decrypting it? 
I can't find any hint in the headers you provide above (maybe it is dropped in the process).

If you find some more time to look into those specifications and provide more information, I can try to fix it.

cocreature:
[g_mime_multipart_encrypted_decrypt](https://developer.gnome.org/gmime/stable/GMimeMultipartEncrypted.html#g-mime-multipart-encrypted-decrypt) has a third option that is a [GMimeDecryptResult](https://developer.gnome.org/gmime/stable/GMimeCryptoContext.html#GMimeDecryptResult). This contains a list of signatures found during decrypting. From a quick look (throwing a printf in mu_msg_crypto_decrypt_part) it looks like this is the correct thing to look at. Not quite sure how to best fit this into the current architecture.

zakkak:
Could you please review https://github.com/zakkak/mu/commit/4efda2154da6af738779c27acce56396cd8a4d7a?w=1 .

I am using the decrypt result but I do not seem to get it working.  Any kind of help is welcomed.

cocreature:
Thanks, I'll look into it when I have time. (Exams coming up, so it might take me a bit.)

zakkak:
[...]

PR #584, opened 2015-02-16
Fix for #280

zakkak:
Since `parent` is not really used as a parent, I use it as the last
visited encrypted part while going down the parts-tree.

At the decryption of a part (`mu_msg_crypto_decrypt_part`) I check,
through the GMimeDecryptResult, for signatures (`check_decrypt_result`)
and add them to the part (`tag_with_sig_status`).  Any nested parts, hold
that encrypted part as their parent.  Finally at `handle_part`, for each
part I check if it a descendent of an encrypted part.  If so, I proceed
checking for signatures and adding them to the `msgpart`.

### DIFF
--- a/lib/mu-msg-crypto.c
+++ b/lib/mu-msg-crypto.c
@@ -115,7 +115,7 @@ get_gpg (GError **err)
 		 }
 		return g_strdup (envpath);
 	}
-	
+
 	if (!(path = g_find_program_in_path ("gpg")) &&
 	    !(path = g_find_program_in_path ("gpg2"))) {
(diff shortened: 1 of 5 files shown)


zakkak:
@djcb Please review this PR carefully.  I am not sure changing the behavior of `parent` does not break anything else.  Additionally we might need to reconsider the name of `parent` in case you decide to pull this upstream.  Thanks!

djcb:
It looks okay to me.... I'm going to merge it, I suppose that's the best way to give it some more testing!
Thanks for your work!

PR #593, opened 2015-02-28
Add option to delete the found duplicates

### LINKED ISSUE #592: find-dups script should have an option to delete duplicates found
Like --delete

hexmode:
Closes #592

### DIFF
--- a/guile/scripts/find-dups.scm
+++ b/guile/scripts/find-dups.scm
@@ -22,6 +22,7 @@ exec guile -e main -s $0 $@
 ;; INFO: find duplicate messages
 ;; INFO: options:
 ;; INFO:   --muhome=<muhome>: path to mu home dir
+;; INFO:   --delete: delete all but the first one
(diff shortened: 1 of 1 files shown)


djcb:
Hmmm, looks useful, but  I did not add delete support, because it runs the risk of deleting the wrong messages... Are we very certain that cannot happen? E.g. `(format #f "mu remove ~a\n" filepath)` does not escape the paths.

hexmode:
Dirk-Jan C. Binnema writes:

> Hmmm, looks useful, but  I did not add delete support, because it runs
> the risk of deleting the wrong messages... Are we very certain that
> cannot happen? E.g. `(format #f "mu remove ~a\n" filepath)` does not
> escape the paths.

The paths come from whatever tool manages the maildir, but I'll check
this over and add escaping if needed.

hexmode:
abandoning for something else.

PR #595, opened 2015-03-04
Fix a core dump under OpenBSD-current (ca. 5.7-beta, 04-MAR-2014) on amd64

StAlphonsos:
First, thanks for mu, it's great!

I'm a heavy mu/mu4e user on OpenBSD.  I just switched to running
-current on my amd64 thinkpad and mu was dumping core.  The reason
seems to be that c_str() was being used as the return value from
mu_msg_iter_get_{msgid,thread_id}.  At least in the case of
get_msgid() this resulted in an invalid pointer being passed to
g_strdup() in get_thread_ids() in mu-query.cc.  The same pattern was
used in mu_msg_iter_get_thread_ids().  In both cases the function is
only called once and the caller just calls g_strdup() so I moved the
call to g_strdup() into the functions in mu-msg-iter.cc.

### DIFF
--- a/lib/mu-msg-iter.cc
+++ b/lib/mu-msg-iter.cc
@@ -366,7 +366,7 @@ mu_msg_iter_get_msgid (MuMsgIter *iter)
 	g_return_val_if_fail (!mu_msg_iter_is_done(iter), NULL);
 
 	try {
-		return iter->msgid().c_str();
+		return g_strdup(iter->msgid().c_str());
 
 	} MU_XAPIAN_CATCH_BLOCK_RETURN (NULL);
 }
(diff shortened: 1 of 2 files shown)


djcb:
Oh, that makes sense! I took your patch/idea and updated the code, available in master. Does it work for you? Thanks!

StAlphonsos:
Sorry at my delay in responding: I haven't had a chance to uninstall mu and try again with your patched sources... I use mu too much it seems :-).  I looked at main and it looks right now, but I'll comment when I get a chance to test.  Thanks for the quick response.

djcb:
Okay, I'm closing this. Feel free to reopen if you can still reproduce -- thanks!

qbit:
:+1:  Thanks for doing this!

PR #596, opened 2015-03-07
Add option to delete the found duplicates

### LINKED ISSUE #592: find-dups script should have an option to delete duplicates found

hexmode:
Closes #592

Uses open-pipe* to ensure that escaping isn't needed (seen in md5sum)
since arguments are passed.

Also see this discussion: http://thread.gmane.org/gmane.lisp.guile.user/11777


djcb:
So, it seems this also changing a few files in mu4e, can you remove those changes?

hexmode:
fixed.  please merge now.

djcb:
Merged it -- thanks!

(I added a few extra warnings, hope people know what they're doing when they use this!)

PR #597, opened 2015-03-17
Extract mu4e-headers-mark-thread-using-markpair

danieroux:
This allows for a key binding to something like:

(mu4e-headers-mark-thread-using-markpair '(read))


djcb:
Oh, that looks useful; merged. Thanks!

PR #600, opened 2015-03-25
update docstring in mu4e-message-field-raw

jamestechnotes:
Just updates a docstring in mu4e-message-field-raw.  (changed :data to :date)


djcb:
Can you make a pull-request with just that change, against master?

Issue #215, opened 2013-05-19
Window #<window 0x2bfe9f0 on *mu4e-headers*> too small for splitting


mhofert:
1) In _mu4e-headers_, hit return while the point is on message A
2) hit 'R' to reply to A
3) put the point in _mu4e-headers_ again, now on message B. Hit return to display B
4) Instead of seeing B, I get the message in the title of this post (unfortunately not correctly displayed in this body)

tiagosab:
This has been biting me ever since I started using mu4e. It is not just with compose-mode, but everytime the message-view is not visible. I am submitting a pull request which traps the error and open the message replacing message-headers.

PR #602, opened 2015-03-28
mu4e: avoid error when splitting headers window

tiagosab:
This solves the problem in issue https://github.com/djcb/mu/issues/215. There was an error everytime the message-view was not visible and the message-headers was not large enough to be splitted. I am trapping this error and opening message-view on the same window message-headers was (effectively replacing it, just like with mu4e-split-view set to anything but 'horizontal' or 'vertical').


djcb:
Merged -- thanks!

PR #605, opened 2015-04-16
Update version number

pSub:


djcb:
Version has been updated, so this is no longer needed... thanks!

PR #610, opened 2015-04-27
mu4e: match truename of `mu4e-maildir`

ptpt:
If `mu4e-maildir` is a symlink, mu4e will
fail to determine drafts folder when you want
to edit drafts.


djcb:
Thanks for the patch, but `mu4e-maildir' should not be a symlink, see its docstring; allowing symlinks there would cause all kind of oddities with index.

PR #614, opened 2015-05-20
* mu4e/mu4e-actions.el: modify org-capture-templates

emon:
add " " between :EMAIL: and 'mail' because org-contacts needs the separator.
move the cursor to the left of 'name'


djcb:
Merged -- thanks!

PR #615, opened 2015-05-22
Add support of open compose window in new frame: Could resolve #554

### LINKED ISSUE #554: Compose mail in new frame
When I used the VM mailer, I found the `vm-mail-new-frame` function very useful. When composing a new email or writing a reply, this opens up a new frame for the composition buffer. When the email is sent, the frame is killed. 
I would find this particularly useful in mu4e. Very often I try replying to an email, which then occupies the pane previously occupied by the buffer of the email to which I'm replying. To look at this buffer then requires me to switch buffers, or open a new frame.
I've tried using `message-mail-other-frame` to achieve the desired effect, but this opens the buffer in the same frame, and doesn't display the mail buffer in another frame.
I've looked at the mu4e code, and come to the tentative conclusion that changes might be needed there - but my emacs-fu and time is too limited to make progress.

davidcsterratt:
I've cherry-picked the commit of @panjie , which seems to do the job.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -1,3 +1,4 @@
+;; -*-mode: emacs-lisp; tab-width: 8; indent-tabs-mode: t -*-
 ;; mu4e-compose.el -- part of mu4e, the mu mail user agent for emacs
 ;;
 ;; Copyright (C) 2011-2012 Dirk-Jan C. Binnema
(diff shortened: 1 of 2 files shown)


djcb:
Oh, that looks useful. Merged -- thanks!

zakkak:
It is really useful, however it still fails to close the frame after the e-mail is sent.

PR #616, opened 2015-05-25
add functions to handle spam with Bogofilter

gour:
Here is request adding some functions to handle spam messages with Bogofilter with the ability to configure support for SpamAssassin as well.


djcb:
Thank you! I've merged your patch, but didn't enable it to set up the variables and the keybindings etc. by default, since that might be confusing for other users.

 I also did not include the ``bf-`` prefixed functions yet; we can takes those in with a mu4e- prefix, and setting the bogofilter command through a defvar, like the existing ones. Thanks!

PR #626, opened 2015-06-07
Make `mu4e-compose-in-new-frame` always kill frame

zakkak:
When `message-kill-buffer-on-exit` was set to nil mu4e would not close
the new frame.  Closing the new frame should be independent of killing
the buffer.  This patch corrects this.

(relates to #615)


djcb:
Merged -- thanks

Issue #621, opened 2015-05-26
Rudimentary BBDB support for mu4e available


davidcsterratt:
BBDB lovers: [BBDB](http://savannah.nongnu.org/projects/bbdb/) now works with mu4e. The instructions below should set up the following:
1. `Tab` completion of addresses when composing emails
2. In mu4e-view mode, adding the sender of the email to BBDB with `C-u :`
3. Viewing the BBDB contact while viewing a message

N.B. Due to running an old version of Linux, I've not tested the building instructions; I've just copied the bbdb *.el files to my `load-path` and then used Method 3 to load bbdb. I'm enjoying using [Awesome Synk](http://asynk.io/) to keep my BBDB synchronised with Google Contacts.

## Installation
```
git clone git://git.savannah.nongnu.org/bbdb.git
cd bbdb
./autogen.sh
./configure --with-mu4e-dir=/path/to/mu4e
make
make install
```
## Setup
In your `.emacs`
```
;; Load BBDB (Method 1)
(require 'bbdb-loaddefs) 
;; OR (Method 2)
;; (require 'bbdb-loaddefs "/path/to/bbdb/lisp/bbdb-loaddefs.el")
;; OR (Method 3)
;; (autoload 'bbdb-insinuate-mu4e "bbdb-mu4e")
;; (bbdb-initialize 'message 'mu4e)
(setq bbdb-mail-user-agent (quote message-user-agent))
(setq mu4e-view-mode-hook (quote (bbdb-mua-auto-update visual-line-mode)))
(setq mu4e-compose-complete-addresses nil)
(setq bbdb-mua-pop-up t)
(setq bbdb-mua-pop-up-window-size 5)
```
## Covenience function to synk BBDB with Google Contacts (requires Asynk)
```
 (defun bbdb-sync()
  (interactive)
  (bbdb-save)
  (shell-command "sync-contacts")
  (bbdb-revert-buffer))
```

zakkak:
Great news, thanks!!!

I think we should include this in the manual (maybe after the BBDB release).

djcb:
Indeed -- in the manual there's an appendix 'Interaction with other tools', and something about interacting with bbdb would fit in nicely.

davidcsterratt:
I could something about BBDB in this section - I presume the file to edit is
https://github.com/djcb/mu/blob/master/mu4e/mu4e.texi ?

djcb:
Yes please!

djcb:
[...]

PR #630, opened 2015-06-23
#621 BBDB documentation

davidcsterratt:
Here is some basic documentation for BBDB. I'll come back to it once there is an official release of BBDB with the mu4e integration.


djcb:
Merged -- thanks!

PR #632, opened 2015-06-30
Fix core dump in mu-server (at least) for cmd:extract index:0 ...

StAlphonsos:
Under OpenBSD (i386-current) I just experienced a core dump from mu-server.
The issue was sending cmd:extract index:0 ...
This patch seems to fix it cleanly, now I just get an error (part 0 does not exist).

Thanks for writing mu! Great stuff.


djcb:
Merged -- thanks!

Issue #417, opened 2014-04-26
Various problems with mu4e-view-message-with-msgid.


tmalsburg:
When opening a message using the function `mu4e-view-message-with-msgid`, it's not possible to open the attachments of that message when there's no *mu4e-headers* buffer.  The error message is:

    mu4e-error: [mu4e] Not valid attachment. 

Typically there is a headers view because this is where messages are selected for viewing.  However, there are other situation when there's not necessarily a headers view, for instance, when clicking on a message link in an org mode file.

**Edit:** This description is not correct.  See my post from [29 May 2015](https://github.com/djcb/mu/issues/417#issuecomment-106892002) below for a correct description and for instruction for reproducing this problem.

tmalsburg:
This problem does not reliably reproduce.  It happens only sometimes and so far I couldn't pin down under which circumstances.

tmalsburg:
I finally found the trigger for this problem.  The absence of the headers-view is not the cause.  The problem occurs only when I open an email with `mu4e-view-message-with-msgid` when a mu4e-view with another email already exists.  Example:

1. Open first email: `(mu4e-view-message-with-msgid "117ef7b57667fb678b73b40d5bbaf270@server.de")`
2. Close mu4e-view.
3. Open second email: `(mu4e-view-message-with-msgid "901A2D07-E0A9-4F4D-AEE5-2F2F9507E1C6@posteo.de")`
4. Open attachment → no problem.

But:

1. Open first email: `(mu4e-view-message-with-msgid "117ef7b57667fb678b73b40d5bbaf270@server.de")`
2. Do **not close** mu4e-view.
3. Open second email: `(mu4e-view-message-with-msgid "901A2D07-E0A9-4F4D-AEE5-2F2F9507E1C6@posteo.de")`
4. Open attachment → attachment not found or attachment from email 1 is opened.

The problem is not only opening attachment but also replying and forwarding.  When I reply to or forward email 2 in the second scenario, the text of the first email is quoted.
 
This bug is really nasty because it can cause you to inadvertently disclose sensitive information to the recipient when forwarding or replying.

djcb:
mu4e-view-message-with-message-id now works just like a normal search/view.

tmalsburg:
Cool, thanks.

PR #635, opened 2015-07-03
mu4e: Added workaround for #417.

tmalsburg:
A fix of the underlying issue would of course be preferable but my understanding of mu4e is not good enough for that.


djcb:
Hmmm... what's this ``if-let``? My emacs does not have that...

tmalsburg:
Fixed it.   (`if-let` was introduced a year ago in subr-x.el.)

djcb:
Merged -- thanks!

PR #642, opened 2015-07-24
* mu4e-draft.el: fix adding User-agent header

eikek:
Hello,

I noticed that I don't have a User-agent header in my mails. With this change, it shows up in my outgoing mails.


djcb:
Merged -- thanks!

PR #648, opened 2015-08-12
Follow-up to commit #630: Clarify bbdb setup comes after mu4e setup

davidcsterratt:
I've hopefully made it clear that the BBDB setup should come after `(require 'mu4e)`.


djcb:
Merged, thanks!

PR #665, opened 2015-09-17
Make usage of fancy chars configurable.

floli:
mu4e-use-fancy-chars to can have the setting t (use fancy chars for marks and threads), marks, threads or nil. The change won't brake any config.


djcb:
Merged -- thanks!

PR #666, opened 2015-09-22
Do not convert dash to underscore when indexing

ptroja:
Mu converts dash (-) to underscore (_) when indexing. This makes it impossible to search for strings like "non-terminating" (instead one had to search for "non terminating" which is only an approximation).
    
This conversion also prevents me from searching for my company's internal ticket numbers, which contain dashes. Other mail indexers, like Thunderbird and notmuch, do not have a problem with searching for strings with dashes.

(Minor whitespace and comment fixes also included in this pull request).

### DIFF
--- a/lib/mu-flags.h
+++ b/lib/mu-flags.h
@@ -137,7 +137,7 @@ MuFlags mu_flags_from_str (const char *str, MuFlagType types,
  * @return concatenation of all non-standard flags, as a string; free
  * with g_free when done. If there are no such flags, return NULL.
  */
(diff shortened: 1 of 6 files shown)


djcb:
Hi Piotr - appreciated, but I won't merge this, as it's likely to break some other setup. In general, it is a bit of a mess trying to work around the way Xapian parses queries. 

The real fix for that is to use a mu-specific query parser rather than try to outsmart the default one. I'm actually working on one (see https://github.com/djcb/mux), but that'll a few months. Until that's ready, I suggest you use your forked mu.

ptroja:
Sure, I expected this. I am happy to live with my own fork until mu-x is ready. Thanks!

PR #669, opened 2015-09-23
Ensure tests have unique names.

atsampson:
glib 2.46 complains if this isn't the case:
(test-mu-str:28790): GLib-ERROR **: duplicate test case path: /mu-str/mu-str-esc-to-list


djcb:
Merged -- thanks!

PR #675, opened 2015-10-08
Fix issues reported by the byte-compiler.

tarsius:
The byte-compiler also complains about `mu4e-doc-dir` not being defined. I have not fixed that because I don't know if that variable is defined and set in a generated file. If so, you should add `(defvar mu4e-doc-dir)` to silence the compiler.

I have also noticed that you still use `cl` instead of `cl-lib` and that indentation often is very broken. Would you accept patches that fix those things?

### DIFF
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -304,11 +304,14 @@ If there is just one recipient of ORIGMSG do nothing."
 It looks something like
   <time>-<random>.<hostname>:2,
 You can append flags."
-  (let* ((hostname
-	   (downcase
-	     (save-match-data
-	       (substring system-name
-		 (string-match "^[^.]+" system-name) (match-end 0))))))
+  (let* ((sysname (if (fboundp 'system-name)
+		      (system-name)
+		    (with-no-warnings system-name)))
+	 (hostname (downcase
+		    (save-match-data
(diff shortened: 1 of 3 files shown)


djcb:
Merged -- thanks!

`mu4e-doc-dir` is indeed generated (`mu4e-meta.el`). As for the `cl-lib`... do you know what version of emacs first shipped with that? Don't want to break backward compatibility unnecessarily.

Indentation fixes are okay when they are for some function that is being modified already for some other reason.

tarsius:
> cl-lib

There is a `cl-lib` available from GNU Elpa which works with older Emacsen. It's a forward compatibility package and just defined `cl-` prefixed aliases for the old prefix-less symbols. It was created exactly so that packages wouldn't have to keep using the prefix-less symbols for backward compatibility reasons.

> Indentation fixes are okay when they are for some function that is being modified already for some other reason.

For my own projects I use exactly the opposite policy: cosmetics changes should be committed separately :-) But other people use different policies, that's why I asked.

djcb:
Separate commits for cosmetic changes are fine, but the main point is that big indentation cleanups make diffs harder to read, so I try to do any such cleanups step-by-step. 

Let me think a bit about cl-lib.

Thanks -- Dirk.

PR #678, opened 2015-10-14
Fix propertizing body with mu4e-view-body-face.

thierryvolpiatto:
* mu4e/mu4e-view.el (mu4e-view-message-text):
When using `propertize' all properties added by `mu4e-html2text-command'
in html message are overwrited by `mu4e-view-body-face', so use here
`add-face-text-property' if available, otherwise behave as before and return
body unchanged.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -240,7 +240,10 @@ found."
 		 (mu4e~view-custom-field msg field))))))
       mu4e-view-fields "")
     "\n"
-    (propertize (mu4e-message-body-text msg) 'face 'mu4e-view-body-face)))
+    (let ((body (mu4e-message-body-text msg)))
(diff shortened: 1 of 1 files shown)


djcb:
Thanks, merged!

thierryvolpiatto:
Hi,
after updating mu, I found that commit 1a33a71 is breaking the purpose of `mu4e-html2text-command`.
By using propertize, you are overwriting all other properties that may have had added by the call
to `mu4e-html2text-command`.
In this patch I use `add-face-text-property` with its `append` arg, which is a great improvement in emacs, but unfortunately not available in older emacs (24.3 I think).
So when this function is not present, I just return body unchanged.

thierryvolpiatto:
Ah! thanks, you were so fast that I didn't have the time to post further explanations in time :-)

PR #682, opened 2015-10-26
docs: mu-easy: summary to summary-len

jeroentbt:
This option has apparently changed somewhere down the line.


djcb:
Merged -- thanks!

PR #689, opened 2015-11-08
Proposed action: show this thread only, meta-data in mu4e~headers-last-query

dkogan:
This is the pull request from the mailing list thread

https://groups.google.com/forum/#!topic/mu-discuss/IqowWeupeMk

### DIFF
--- a/mu4e/mu4e-actions.el
+++ b/mu4e/mu4e-actions.el
@@ -268,6 +268,17 @@ store your org-contacts."
 
     (mu4e-message (concat "tagging: " (mapconcat 'identity taglist ", ")))
     (mu4e-refresh-message path maildir)))
+
+(defun mu4e-action-show-thread (msg)
+  "Show all the messages in the given thread."
+  (let ((msgid (mu4e-message-field msg :message-id)))
+    (when msgid
+      (let ((mu4e-headers-show-threads t)
+            (mu4e-headers-include-related t))
+        (mu4e-headers-search
+         (format "(%s) AND msgid:%s"
+                 mu4e~headers-last-query
+                 msgid))))))
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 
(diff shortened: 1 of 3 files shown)


djcb:
Oh - thanks! But could you make a slightly less intrusive change, ie. just the action and the adding it in the header/view? The other changes are a (possible) next step, but may affect other people's setups, so let's wait with that.

dkogan:

> Oh - thanks! But could you make a slightly less intrusive change, ie.
> just the action and the adding it in the header/view? The other
> changes are a (possible) next step, but may affect other people's
> setups, so let's wait with that.

I'm not sure I follow. Are you suggesting that the action should just
search for the msgid, and leave it up to the user to enable
mu4e-headers-show-threads and mu4e-headers-include-related?

djcb:
Yes; I suspect many users have include-related on by default already (and I'm going to make it the default behavior). show-threads is not strictly necessary (but also on by default).

dkogan:
OK, how about the current tree then (just updated)? It only leaves the action, but at least turns on include-related and show-threads for that one search. This is simple, and will do the right thing the first time (and the first time only).

Oh, and making the query "previous_query AND msgid:xxx" has an issue: if include-related is on, then search results can contain messages that do not match the query, and a new query of ""previous_query AND msgid:xxx" would thus yield no results. The tree in this request still has the "previous_query AND msgid:xxx", but I think it should be changed back to "msgid:xxx"

dkogan:
Hi again. Now that I've lived with this feature for a few days, I end up using it constantly, which is good. I'm also hitting the case very often where "previous_query AND msgid:xxx" yields no results. My workflow is to navigate to a particular folder (maildir: search) with include-related turned on, find an arbitrary message in a thread I'm interested in, and then invoke the new action in this patch. If I was lucky, the arbitrary message lives in the folder I searched for, and everything works. If I wasn't lucky, the message isn't there anymore, and I get no results. Let's go back to just "msgid:xxx".

djcb:
The changes to mu4e-view / mu4e-headers gave a compilation error, because  the "(" and ")" were in the wrong places. I've fixed that, and also changed the query back to just msgid:xxx, and pushed it to master.

PR #691, opened 2015-11-11
mu4e: Moving in headers should only change view if mu4e-split-view is non-nil.

### LINKED ISSUE #690: Deleting or flagging a message in header-view opens the next message even in single window mode.
I prefer to use single window mode, so I see either the header-view xor the message-view.  Since the last update from MELPA, I get a behaviour that is not desirable in that mode.  When I delete or flag a message in header-view, the next message is opened mu4e-view.  This did not happen in earlier versions of mu4e. I think 0956c096decd628b69336c564b9c08c1640f5b31 was supposed to solve this problem but missed some cases.

I think this issue only occurs when a mu4e-view exists somewhere in the background (not currently displayed).  Perhaps the problem is that `(window-live-p mu4e~headers-view-win)` is not a reliable way to detect whether the user runs split-view or not.  This could be fixed by additionally requiring that the mu4e-view window is visible but I would prefer if mu4e would not mess with my mu4e-views at all in single-window mode.  If I open a message in separate window next to the headers view, I really want it to be there.

tmalsburg:
Fixes #690.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1451,7 +1451,7 @@ docid. Otherwise, return nil."
       ;; attempt to highlight the new line, display the message
       (mu4e~headers-highlight docid)
       ;; update message view if it was already showing
-      (when (window-live-p mu4e~headers-view-win)
+      (when mu4e-split-view
 	(mu4e-headers-view-message))
       docid)))
 


djcb:
Merged -- thanks for fixing this!

PR #695, opened 2015-11-16
Include '-' in example regex

hinrik:
Dashes are pretty common in mailing list names.


hinrik:
Just updated it to include underscore as well.

djcb:
Merged, thanks!

ardpp:
On my laptop running Archlinux compilation fails with the following output. 
```
mu-maildir.c:771:1: Error: expected »=«, »,«, »;«, »asm« or »__attribute__« before »char«
char*
```
Let me know if I should provide more details.

PR #697, opened 2015-11-17
Add mu4e-headers-mark-all to execute certain action on all visible msg

j-hao:


djcb:
Ah, nice example. But can you make the docstring start with a capital, end with a '.'? And mention that it is about all the messages in the current query results, not just the messages which are visible. Thanks!

j-hao:
Hi, updated docstring and commit message. Let me know if this looks good to you. Thanks

djcb:
Merged -- thanks!

PR #701, opened 2015-11-19
mu-find(1): Document that you can search throug body:*

avar:
This is a long-standing feature, it works, but there was no mention of
it in the docs.

There's also bodyhtml:* but that doesn't seem to do anything for me,
maybe I just don't have enough HTML mail.

These also exist:

    path,l          ???
    refs,r          ???

But I don't know what they're for. They should probably be documented
too.

### DIFF
--- a/man/mu-find.1
+++ b/man/mu-find.1
@@ -104,6 +104,7 @@ search fields and their abbreviations:
 	from,f          Message sender
 	to,t            To: recipient(s)
 	subject,s       Message subject
+	body,b          Message body
 	maildir,m       Maildir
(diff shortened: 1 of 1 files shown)


djcb:
Merged -- thanks!

avar:
Thanks for merging it. B.t.w. what do you think about the latter half of the commit message, i.e. should `path:` and `refs:` be documented too?

djcb:
Perhaps `path:` could be useful for end-users, though `refs:` is more of an internal thing. But yeah, improvements to the docs are always welcome.

PR #704, opened 2015-11-22
mu4e-{view,headers}: Refactor a (lamba) kbd-bound function into a named function

avar:
Because this was a lambda C-h m would just show "??" instead of the
function name, and the documentation would be really confusing since it
showed the deparsed lambda function instead of the function name being
called.

Fix this by refactoring both the view & headers [ and ] functions into
named functions, and make their shared logic new internal
mu4u~{headers,view}-* functions.


avar:
I looked at these build failures and they seem to be some unrelated issue with the build system, not an issue with my patch.

djcb:
Merged -- thanks! Indeed, the build-failures seem to be a failure of the build-system itself.

PR #705, opened 2015-11-23
Query markpair in interactive declaration

myrjola:
My mail processing workflow is to browse through threads with the press of a button and mark the previous thread as read. Now the closest I can get is pressing `T!` to mark thread read. Unfortunately this can't be reduced to one keypress as `mu4e-headers-mark-thread` always queries the mark type separately.

I did some refactoring to the relevant mark-thread functions to make my workflow possible without using the private `mu4e~view-in-headers-context`. So my view-mark-thread-as-read keybinding becomes:

```
(lambda ()
  (interactive)
  (mu4e-view-mark-thread '(read)))
```

Relevant issue: https://github.com/syl20bnr/spacemacs/pull/3922/.

If there's no reason to have  `mu4e~view-in-headers-context` as private, then I think it is cleaner to just make it public API that hopefully won't change in the future.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1153,21 +1153,23 @@ descendants."
       (goto-char last-marked-point)
       (mu4e-headers-next))))
 
-(defun mu4e-headers-mark-thread (&optional subthread)
+(defun mu4e-headers-mark-thread (&optional subthread markpair)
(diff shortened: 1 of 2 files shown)


djcb:
Thanks -- that looks useful! I've merged it (locally). 

[ I consider `mu4e~view-in-headers-context` a bit of a hack, so I'm not planning to make it public right now. ]

PR #710, opened 2015-11-28
Cleaned up some trailing white space(s)

jjasghar:
- Walked through the NEWS, README, and about org-pages and removed some
  white spaces that where sticking around.
- :metal:


djcb:
Merged -- thanks!

PR #714, opened 2015-12-08
Add support for mu4e-use-fancy-chars to the marking in the headers view

avar:
The mu4e-use-fancy-chars customization wasn't supported for the headers view's main marking of messages. Now it is, I've bent over backwards here to not make this break anyone's existing customization at the expense of some slight complexity.

I just added non-ASCII fancy characters for two of the actions, I couldn't find topical ones for the rest.

### DIFF
--- a/mu4e/mu4e-mark.el
+++ b/mu4e/mu4e-mark.el
@@ -124,7 +124,13 @@ is either a headers or view buffer."
   "The list of all the possible marks.
 This is an alist mapping mark symbols to their properties.  The
 properties are:
-  :char (string)  The character to display in the headers view
+  :char (string) or (basic . fancy) The character to display in
+    the headers view. Either a single-character string, or a
+    dotted-pair cons cell where the second item will be used if
+    `mu4e-use-fancy-chars' is `t' or `marks', otherwise we'll use
+    the first one. It can also be a plain string for backwards
+    compatibility since we didn't always support
+    `mu4e-use-fancy-chars' here.
   :prompt (string) The prompt to use when asking for marks (used for
      example when marking a whole thread)
   :ask-target (function returning a string) Get the target.  This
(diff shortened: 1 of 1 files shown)


djcb:
Oh, that looks nice - thanks!

ardpp:
May I ask which fonts show those symbols? I'm using DejaVu and the other fancy-chars are shown.

avar:
@ardpp On Debian I'm using GNOME Terminal & "Monospace Regular" and all these characters appear for me, but of course if this new stuff I've hacked up uses symbols that aren't available where the existing fancy ones are we should reconsider having this on under the same switch by default.

What's your system setup like?

Also if you copy/paste the table from [this Wikipedia page](https://en.wikipedia.org/wiki/Miscellaneous_Symbols_and_Pictographs) into your scratch buffer how many of those symbols can you see in your Emacs? With the above setup I'd say roughly 90-95% of the symbols are shown correctly in my terminal.

ardpp:
@avar I'm on Arch Linux and use DejaVu fonts in different flavours for almost everything. When I opened the Wikipedia page you linked to almost none of the symbols were shown, not in firefox and not in emacs. I investigated a bit and after installing the package `ttf-symbola` almost all symbols are shown everywhere.

zobi8225:
@ardpp :+1:

avar:
@ardpp Ah, my fonts on Debian are:

    $ dpkg -l | grep ttf|awk '{print $2}' | tr '\n' ' '
    fonts-freefont-ttf ttf-dejavu ttf-dejavu-core ttf-dejavu-extra ttf-unifont

I'd installed `ttf-dejavu` and `ttf-unifont` manually before, so maybe this isn't shown in the default install.

avar:
@zobi8225 What does that thumbs up mean? You like his comment, or you have the same issue, or something else?

ardpp:
@avar so I suppose that in your case `ttf-unifont` is doing what `ttf-symbola` does for me. `ttf-unifont` is also available in for Arch Linux but it is not installed on my laptop: 
```
$ pacman -Q | grep ttf
ttf-dejavu 2.34-2
ttf-freefont 20120503-4
ttf-inconsolata-g 20090213-3
ttf-ms-fonts 2.0-10
ttf-symbola 8.00-1
```

avar:
So, what do we want to do here? Should I add a `mu4e-use-fancier-chars` in addition to `mu4e-use-fancy-chars` ?:)

ardpp:
I was naively thinking that emacs is only using the font that I set for it. That's why I asked which font you use for the symbols. So the only thing I would probably do is to mention (somewhere in the manual?) the example font families (e.g. `ttf-unifont` and `ttf-symbola`)  that need to be installed.

djcb:
I think `mu4e-use-fancy-chars` should be enough; but (as @ardpp mentioned) a note in the manual would be good.

avar:
On Mon, Dec 14, 2015 at 6:34 AM, Dirk-Jan C. Binnema <
notifications@github.com> wrote:

> I think mu4e-use-fancy-chars should be enough; but (as @ardpp
> <https://github.com/ardpp> mentioned) a note in the manual would be good.
>

I can add that, but I'm unsure what the current status is vis-a-vis
https://github.com/djcb/mu/issues/733

I.e. are the characters you used like the scissor icon instead of the
trashcan not causing the same font issues for users?

Issue #716, opened 2015-12-08
Make the status reporting hardcoded at every 1k messages customizable


avar:
See also issue #715. This is currently hardcoded at 1000 in the source, and both this and showing this at all could be arguments to cmd:index.

avar:
This is solved with pull request #722.

djcb:
Thanks for that, but it seems like a lot of extra complexity / changes for little improvement, or?

avar:
I wanted to hack on this before working on the more general issue of issue #709. This makes the debug log a lot more usable for me, and I think in general if mu4e doesn't need something from mu its should pass an option along to suppress that.

djcb:
As discussed in the PR, I think it's a bit too much extra complexity for little gain; still, appreciate the work, thank you!

avar:
If you don't want this pull request do you think you wouldn't like if I implement issue #709 either? The patch for that will likely look very much the same.

I don't really need to customize this, I just want to quite it, do you think you'd take a patch that would just make the hardcoded 1000 number a ./configure time define so I could just pass --report-processed-every=0 or something?

Issue #715, opened 2015-12-08
Is there a reason for why :status messages are sent when mu4e-hide-index-messages is t?


avar:
It seems like a lot of chatter in the protocol could be eliminated if the `:status' messages weren't sent when `mu4e-hide-index-messages' is on.

I noticed this because I'm watching what's going on in the `*mu4e-log*` buffer to debug something, and it'll get a lot of noise every `mu4e-update-interval'.

avar:
This is solved with pull request #722.

djcb:
Thank you & appreciate the work, but not going to take the PR... closing this.

PR #717, opened 2015-12-08
mu-server man page: Correct documentation about status reporting

avar:
This 500 number was correct in v0.9.7-pre-302-g858552a, but since
v0.9.8.2-41-gc2e3eac it's been hardcoded to 1000 in the server code.

See also issue #715 and issue #716.


djcb:
Merged, thanks!

Issue #75, opened 2012-10-15
add thread fold/unfold


zabbal:
Ability to fold/unfold thread via single click and hotkey will make UI much cleaner and convenient to use - especially in case of ML where long heated threads are not uncommon thing.

sachdevaprash:
I was wondering where this is at and if I can help complete this. Do we have a good idea on the strategy to employ for this feature?

zakkak:
@sachdevaprash please follow #783

zabbal:
https://gitlab.com/ambrevar/mu4e-conversation - this seems to be related.

seanfarley:
Damn, this notification really got my hopes up. Much like the rest of 2020, it was a disappointment.

zabbal:
Pardon, wasn't my intention to give false hopes :)

There's incomplete implementation in PR#783 and there's rather complete implementation of a related thing in mu4e-conversation - I don't know enough elisp to bring those together but some ideas in the latter (using org-mode code for folding for example) seem like they could be useful to fill in gaps in the former.

seanfarley:
I think the two features are orthogonal: this issue is, to my understanding, about folding a thread in the header view (something similar to what `notmuch` does). `mu4e-conversation` implements threading in the main view (the body of the emails). Both are useful, for sure, but I'm more interested in folding threads in the header view.

djcb:
The headers view is quite complicated already, I think it would benefit from being split in a generic tree-view-like widget, and then the mu4e-specific logic.

Perhaps something like https://github.com/kiwanami/emacs-ctable or something else.

seanfarley:
That makes sense, @djcb. I assume the goal would be something like `emacs-ctable` would handle the folding, correct?

rougier:
I've setup a small project at https://github.com/rougier/mu4e-thread-folding for folding/unfoldong threads in headers view.
It is based on overlay. Note this is not foolproof, but usable until  the ctable version. One problem is with `(mu4e~headers-get-thread-info msg 'thread-id)` that does not seem to return a unique id when there are a lot of thread.

djcb:
Thread-folding has been merged in master, so finally closing this.

stefanv:
This is huge! Thank you @djcb, @zabbal, @rougier!

djcb:
> > Thread-folding has been merged in master, so finally closing this.
> 
[...]

PR #722, opened 2015-12-09
Partially solve issue #75 & add a :report-processed-every argument

avar:
This argument can be report-processed-every for report no progress, or N
where N >1 for report every N. In mu4e we just use this to pass 0 if
`mu4e-hide-index-messages' is t. This makes the protocol a lot less
chatty during indexing.

We could add something to make the "send us an update every N" to mu4e,
that's supported by mu-server now, but I'm not doing that, this solves
my use-case of having a less chatty *mu4e-log*.

I was trying to do something with (apply) and (mapconcat) in
mu4e~proc-index that would make the if/else branches less copy/pasted,
but couldn't come up with something that wasn't even more verbose and
unwieldy.

This resolves issue #715 and issue #716.

### DIFF
--- a/lib/mu-index.c
+++ b/lib/mu-index.c
@@ -95,6 +95,7 @@ struct _MuIndexCallbackData {
 	gboolean		_reindex;
 	time_t			_dirstamp;
 	guint			_max_filesize;
+	unsigned                _reportevern;
 };
 typedef struct _MuIndexCallbackData	MuIndexCallbackData;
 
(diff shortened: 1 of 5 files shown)


avar:
@zabbal This is the pull request for that code to upstream. It's still working just fine for me FWIW.

avar:
Any update on what you think about this / whether you'd prefer some other approach? This works for me and I think it's sensible to pass this sort of thing to the server. See also your comment on issue #716.

djcb:
Hi @avar, I guess the #716 comment applies here as well. I'm esp. a bit reluctant to change the protocol between mu/mu4e.

avar:
Isn't that fine if it's just adding options or new commands? I can't see
how it would break something.

avar:
Update: Poke @djcb for notification...

djcb:
There's a certain cost to any change; and here we add a new code path etc. Nothing big, but not really needed imho. 

I am working on some bigger overhaul of the mu4e <-> mu connection (see the mu-x repo), but that 't a long term project. In the mean time, I'd like to keep this part of mu4e as it is, as far as possible.

djcb:
As discussed elsewhere, I think it's a bit too much extra complexity, so, closing this; anyhow, I appreciate the work, thank you!

PR #730, opened 2015-12-15
Refactor mu4e~headers-header-handler

cpitclaudel:
Hi @djcb!

I'm starting to like mu4e a lot :) It's really nice. I've been playing around with the code a bit, attempting to understand a few things. While doing so, I noticed two small issues in the header list:

* Header lines always have a trailing space
* Header lines are constructed by repeated concatenation

This PR fixes these two issue by slightly refactoring the `header-handler` function. It also implements two more small changes:

* I removed the `(when str` check. It probably should be added again, but I don't really understand its purpose. In the original code, if `str` is nil, then all subsequent fields will be improperly aligned, won't they?
* I changed `propertize` to `add-face-text-property`, which works in place instead of making a copy, and does not erase the existing faces (which allows me in particular to display the unicode status indicators in a different font.

Speaking of the last point; what would you think of assigning a different face to each column of the header list, or at least to the subject field? It would open the way for plenty of cool customizations.

As a side effect, the change yields a speedup of about 15% on my machine :)

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -466,62 +466,63 @@ found."
 		  (mu4e-error "no :function defined for field %S %S" field (cdr item)))))
     (funcall func msg)))
 
+(defun mu4e~headers-field-handler (f-w msg)
+  "Create a description of the field of MSG described by F-W."
+  (let* ((field (car f-w))
+         (width (cdr f-w))
+         (val (mu4e-message-field msg (car f-w))) (str)
+         (str (case field
+                (:subject
+                 (concat ;; prefix subject with a thread indicator
+                  (mu4e~headers-thread-prefix (mu4e-message-field msg :thread))
+                  ;;  "["(plist-get (mu4e-message-field msg :thread) :path) "] "
+                  ;; work-around: emacs' display gets really slow when lines are too long;
+                  ;; so limit subject length to 600
+                  (truncate-string-to-width val 600)))
+                (:thread-subject (mu4e~headers-thread-subject msg))
+                ((:maildir :path :message-id) val)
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Readable diff: https://github.com/djcb/mu/pull/730/files?w=1

djcb:
Oh, nice work, that's an impressive speedup! Merged this, and will check the 'when str' path.

djcb:
Adding different colors to different fields could be nice; wonder what's the best to do this.

Perhaps use some `:color` property in `mu4e-header-info` and for its default value, extract the foreground color from various font-lock-faces (we use those already as defaults for other faces in mu4e, and they should give us some reasonable default regardless of theme etc., background color etc.).

Then, perhaps a callback in the headers-view (if it's not too slow), so users can customize things further (e.g. think special colors for certain recipients).

cpitclaudel:
Good question; finding a good solution performance-wise seems crucial indeed.
Here's a proposal that should be flexible, reasonably fast, and rather easy to maintain:

 Suppose that instead of the pattern matching we currently do on headers with `cond` we called a user-customizable function. In fact, suppose that there's a list of such functions, and that we called them with `field, message, val, and width`, each in succession, passing the return value of each as the `val` of the next. We would have another sequence of such functions for whole-line properties, too.

In the common case of no customization, we would only incur the cost of an extra function call per field, plus one per line, and this should give us all the flexibility in the world :) Since I made the code use `add-face-text-property` already, we can stack faces, so the successive function applications should work nicely.

I'll experiment with this idea and report :)

cpitclaudel:
Ok, I've opened #735 for this :)

PR #731, opened 2015-12-15
Extend contexts example

zakkak:
Extends the contexts' example to be more functional. It sets a default context to be used and sets `mu4e-user-mail-address-list` to reflect the e-mail addresses in the different contexts.


djcb:
Merged -- thanks!

PR #732, opened 2015-12-15
Don't use real (or potentially real) domains in examples

avar:
Let's not spam the poor sods who own foo.com and bar.com, or home.com
etc. Went through the various examples and changed them to use
example.com or subdomains of example.com where appropriate, even when
the domain or TLD didn't exist (yet!) for completeness.

See-Also: https://tools.ietf.org/html/rfc2606


djcb:
Merged. Don't think those domains were too much of a problem, but anyway, good to be careful -- thanks!

PR #735, opened 2015-12-15
Initial implementation of customizable headers

cpitclaudel:
As discussed in #730, this adds two lists of functions (I'll make them into defcustoms, with proper documentation, if we settle on this implementation) which are applied in sequence to produce the headers line.

* `mu4e~headers-field-handler-functions` are applied one by one to produce each header field.
* `mu4e~headers-line-handler-functions`  are applied one by one to produce the full header line.

This could allow users to display the Unicode status icons in a different font, or to make display the subject line in a variable-pitch font, or to color specific recipients differently, or to add a background to every odd column, and so on.

Similarly, with an implementation of a line handling function that keeps one bit of state, we could have a zebra pattern on lines (though this leads me to ask: should we pass the field or line number to these functions?).

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -461,52 +461,74 @@ found."
 		  (mu4e-error "no :function defined for field %S %S" field (cdr item)))))
     (funcall func msg)))
 
+(defun mu4e~headers-field-apply-basic-properties (msg field val width)
+  (case field
+    (:subject
+     (concat ;; prefix subject with a thread indicator
+      (mu4e~headers-thread-prefix (mu4e-message-field msg :thread))
+      ;;  "["(plist-get (mu4e-message-field msg :thread) :path) "] "
+      ;; work-around: emacs' display gets really slow when lines are too long;
+      ;; so limit subject length to 600
+      (truncate-string-to-width val 600)))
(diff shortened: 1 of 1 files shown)


djcb:
Oh, looks nice & clean, merged. Let's experiment a bit with this! As for the zebra lines, I think `line-number-at-pos` should work, or?

cpitclaudel:
Glad you like the implementation :) Regarding zebra lines: `line-number-at-pos` won't work; it's much too slow (it works by counting lines, so calling it on line 500 is 500 times slower than calling it on line 0 :/). I can make a new PR to add a record id to the list of arguments, or we could add a hook that get's called before display begin, to give functions a chance to clean up their state.

cpitclaudel:
I'm really happy with what this change allows, actually :) I made restricted the application of flag fonts to just message titles, and used a new set of Unicode markers for flags (along with a custom font), and used a variable-width font for message subjects and bodies (which are rendered using shr).

screenshot from 2015-12-15 15 16 39

I like it! Thanks for your cool work on mu4e @djcb. I'll submit a PR that documents these two `defvars` soon, and then we can consider making them `defcustoms`.

djcb:
Oh, that looks very nice!

zakkak:
:+1:

PR #737, opened 2015-12-16
Fix mu4e-context-determine to match documentation

zakkak:
The documentation of `mu4e-context-determine` states that

>If there are contexts but none match, return nil, unless PICK-FIRST is non-nil, in which case return the first context."

but it never actually checks `pick-first`.


djcb:
Merged -- thanks!

PR #738, opened 2015-12-16
When mu4e-context-determine fails, prompt user

zakkak:
When `mu4e-context-determine` fails to pick a context (i.e. returns nil) when invoked from `mu4e~compose-handler` and there are available contexts, prompt the user to pick one.

This PR depends on #737, so please review/pull that one first.


djcb:
Ah yes, that makes sense... Thanks, merged.

PR #744, opened 2015-12-20
mu4e: Only switch context when really necessary.

magthe:
It's unnecessary to call the leave and enter functions when not really switching context.

This will allow use of destructive functions (.e.g. clearing of caches) in the leave function.

### DIFF
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -68,24 +68,28 @@ for the message replied to or forwarded, and nil otherwise. Before composing a n
   (let* ((names (map 'list (lambda (context)
 			     (cons (mu4e-context-name context) context))
 		  mu4e-contexts))
+         (current-name (if mu4e~context-current
+                           (mu4e-context-name mu4e~context-current)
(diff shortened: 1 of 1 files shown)


magthe:
I'm guessing this can be closed with <https://github.com/djcb/mu/commit/60192a60d2dbb9e8b6ee29a8b467241024ad7377>, right?

djcb:
Ah, thanks for this, appreciated. 

I solved the same in a slightly different way though (hope you're not offended!), defaulting to not doing anything when new context is the same as the current one, but allow to force it, using a prefix-argument.

Note that this slightly changes mu4e-context-switch interface.

magthe:
Your way is better :)

djcb:
Thanks :-)  There some customization now, using mu4e-context-policy and mu4e-compose-context-policy.

magthe:
I had a look at those... I can't seem to get things to work the way I want, I'm *always* asked about what context I want.  I'll have to look into it a bit more before raising any issue for it though :)

PR #746, opened 2015-12-23
fix typo

zw963:
Just fix some typo. Thanks


djcb:
Thank you! 

There was indeed a typo in the mu4e-user-mailing-lists docstring, but it was a slightly different one. And I fixed the other type as part of some other doc changes.

zw963:
Be glad to could help a little.

djcb:
Yes, thank you!

PR #750, opened 2015-12-29
View in browser with images

sachdevaprash:
mu4e-actions.el: Adding function to view email in browser with inline images.

Signed-off-by: Prashant Sachdeva <sachdevp@cs.ubc.ca>


djcb:
That's a nice idea, thanks! 

Would be great if this could be combined with #758; then we can replace the current view-in-browser action with one that handles inline and attached images as well.

sachdevaprash:
OK. Just an update, this works for inline images. I'll add those for other attachments too at some point and post a new pull request soon.

In it's current state, I have just noticed some bugs that I will need to fix.

djcb:
Oh - cool!  WRT writing the tmp file, would probably be good to write in an some tempdir (subdir of temporary-file-directory) that is only readable by user. Cheers.

PR #751, opened 2016-01-02
Fix racing condition with context in mu4e~compose-handler

thierryvolpiatto:
* mu4e/mu4e-compose.el (mu4e~compose-handler):
Switch the context after `mu4e-compose-mode', otherwise mu4e in the meaning time
switch back to view-mode buffer and calculate context there.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -403,6 +403,9 @@ tempfile)."
   (set (make-local-variable 'mu4e-compose-parent-message) original-msg)
   (put 'mu4e-compose-parent-message 'permanent-local t)
   (run-hooks 'mu4e-compose-pre-hook)
+  ;; maybe switch the context.
+  (mu4e~context-autoswitch mu4e-compose-parent-message
+			   mu4e-compose-context-policy)
   ;; this opens (or re-opens) a messages with all the basic headers set.
   (condition-case nil
       (mu4e-draft-open compose-type original-msg)
@@ -431,12 +434,7 @@ tempfile)."
    ;; hide some headers
   (mu4e~compose-hide-headers)
   ;; switch on the mode
-  (mu4e-compose-mode)
-  ;; maybe switch the context, this should happen
-  ;; after `mu4e-compose-mode', otherwise mu4e in the meaning time
-  ;; switch back to view-mode buffer and calculate context there.
-  (mu4e~context-autoswitch mu4e-compose-parent-message
-			   mu4e-compose-context-policy))
+  (mu4e-compose-mode))
 
 (defun mu4e-sent-handler (docid path)
   "Handler function, called with DOCID and PATH for the just-sent
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -402,11 +402,7 @@ tempfile)."
   ;; message being forwarded or replied to, otherwise it is nil.
   (set (make-local-variable 'mu4e-compose-parent-message) original-msg)
   (put 'mu4e-compose-parent-message 'permanent-local t)
-  ;; maybe switch the context
-  (mu4e~context-autoswitch mu4e-compose-parent-message
-			   mu4e-compose-context-policy)
   (run-hooks 'mu4e-compose-pre-hook)
-
   ;; this opens (or re-opens) a messages with all the basic headers set.
   (condition-case nil
       (mu4e-draft-open compose-type original-msg)
@@ -435,7 +431,12 @@ tempfile)."
    ;; hide some headers
   (mu4e~compose-hide-headers)
   ;; switch on the mode
-  (mu4e-compose-mode))
+  (mu4e-compose-mode)
+  ;; maybe switch the context, this should happen
+  ;; after `mu4e-compose-mode', otherwise mu4e in the meaning time
+  ;; switch back to view-mode buffer and calculate context there.
+  (mu4e~context-autoswitch mu4e-compose-parent-message
+			   mu4e-compose-context-policy))
 
 (defun mu4e-sent-handler (docid path)
   "Handler function, called with DOCID and PATH for the just-sent
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -144,8 +144,6 @@ to do:
 		     (and (mu4e-context-match-func context)
 		       (funcall (mu4e-context-match-func context) msg)))
 	  mu4e-contexts)
-	;; no matching one; but is there a current one?
-	(mu4e-context-current) 	
 	;; no context found yet; consult policy
 	(case policy
 	  (pick-first (car mu4e-contexts))


thierryvolpiatto:
This describe the problem I face, but I am for now unsure it is the right fix, please wait before merging.
Thanks.

djcb:
Hmmm, I think we should switch the context _before_ creating the draft message, since e.g. the From: field is determined by the context.

Do you have some steps to reproduce the problem? There's not really any reason why we'd switch the focus in view-mode; only at mu4e startup, when composing a message or explicitly by the user.

thierryvolpiatto:
I found the problem, it is from your recent commit a085c88, reverting it fixed this.
Using `mu4e-current-context` doesn' help when switching context back and forth, it should be recalculated in this case.

thierryvolpiatto:
Probably this PR can be closed, and a bug should be open about commit a085c88.

djcb:
Can you describe what you were expecting, and what actually happened? Thanks!

thierryvolpiatto:

> Can you describe what you were expecting, and what actually happened? Thanks!

I have two accounts, "Gmail" and "Yahoo".
The default account (the first in `mu4e-contexts`) is "Gmail", when I reply
to e.g here on github "Gmail" is used, ok good, then I open a mail sent to
my "yahoo" address and reply to it, the context is switched correctly,
then I open again a github mail and reply, mu4e is reusing the
current-context which is "Yahoo"; This is wrong, it should use the context
corresponding to the "To" header or use the car of `mu4e-contexts` if not
found something matching.

IOW Always use the car of `mu4e-contexts` when don't know what to use
instead of expecting the current account is the one to use (Which is
true on startup but maybe not true anymore later).

Thanks.

-- 
Thierry

djcb:
Hmm, I can't reproduce what you're seeing. It seems my contexts switch back and forth as expected. But I'll investigate further.

Note that what happens depends on `mu4e-context-policy` and `mu4e-compose-context-policy`. But I guess you have those at their default values.

thierryvolpiatto:

> Hmm, I can't reproduce what you're seeing. It seems my contexts switch
> back and forth as expected. But I'll investigate further.

Reading the code:

      (or ;; is there a matching one?
	(find-if (lambda (context)
		     (and (mu4e-context-match-func context)
		       (funcall (mu4e-context-match-func context) msg)))
	  mu4e-contexts)
        ;; no matching one; but is there a current one?
        (mu4e-context-current)<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< Here
	;; no context found yet; consult policy
	(case policy
	  (pick-first (car mu4e-contexts))
	  (ask (mu4e~context-ask-user "Select context: "))
	  (otherwise nil))))))


I don't see how it is possible if you have a "To" header such
"mu@noreply@github.com" which is not matched (unless you have added it
to your matching regexps in `:match-func`) by e.g "djcb@you.com".
In this case, the next rule is tried which is `(mu4e-context-current)`
and success because you are coming from your last context.
 
> Note that what happens depends on mu4e-context-policy and
> mu4e-compose-context-policy. But I guess you have those at their
> default values.

Yes I use default values.


Thanks for your time and also for mu4e which comes better and better
over the time.

-- 
Thierry

thierryvolpiatto:
Thierry Volpiatto <thierry.volpiatto@gmail.com> writes:

> Dirk-Jan C. Binnema <notifications@github.com> writes:
>
>> Hmm, I can't reproduce what you're seeing. It seems my contexts switch
>> back and forth as expected. But I'll investigate further.
>
> Reading the code:
>
>       (or ;; is there a matching one?
> 	(find-if (lambda (context)
> 		     (and (mu4e-context-match-func context)
> 		       (funcall (mu4e-context-match-func context) msg)))
> 	  mu4e-contexts)
>         ;; no matching one; but is there a current one?
>         (mu4e-context-current)<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< Here

And BTW it seems this defeat the purpose of
`mu4e-compose-context-policy` when set.

-- 
Thierry

djcb:
Thanks!

So, the intended/documented behavior is that we first try the match-function, if that doesn't find anything, we try if there already was a context, and if that didn't work either, we check the policy on what to do. 

From you initial report, it seems that it happens as described here; however, your preferred behavior would be to pick the default (first) whenever we can't find a matching context, rather than stay where we are. IOW, mu4e prefers the *current* over the *default* when it can't find anything, where you'd expect the opposite.

Now, something could be said for both behaviors - I was thinking of the case where you have a couple of contexts, say, 'work' and 'private', but don't necessarily want to set match functions for those. In that case, it's nice if they can just manually switch to 'work' or 'private' without mu4e flipping back to whatever was the first (default) account.

But, your default-means-default-always behavior makes sense as well. Let me see how I can accommodate both...

thierryvolpiatto:

> Thanks!
>
> So, the intended/documented behavior is that we first try the
> match-function, if that doesn't find anything, we try if there already
> was a context, and if that didn't work either, we check the policy on
> what to do.
>
> From you initial report, it seems that it happens as described here;
> however, your preferred behavior would be to pick the default (first)
> whenever we can't find a matching context, rather than stay where we
> are. IOW, mu4e prefers the current over the default when it can't find
> anything, where you'd expect the opposite.
>
> Now, something could be said for both behaviors - I was thinking of
> the case where you have a couple of contexts, say, 'work' and
> 'private', but don't necessarily want to set match functions for
> those. In that case, it's nice if they can just manually switch to
> 'work' or 'private' without mu4e flipping back to whatever was the
> first (default) account.
>
> But, your default-means-default-always behavior makes sense as
> well. Let me see how I can accommodate both...

Great, thanks.

BTW I have some code ready for showing the context in main view when
toggling with ";", will send PR as soon this is clarified.

-- 
Thierry

thierryvolpiatto:

> But, your default-means-default-always behavior makes sense as
> well. Let me see how I can accommodate both...

What about asking user when current context is different of default one?

      (or ;; is there a matching one?
	(find-if (lambda (context)
		     (and (mu4e-context-match-func context)
		       (funcall (mu4e-context-match-func context) msg)))
	  mu4e-contexts)
	;; no matching one; but is there a current one?
	(let ((cur (mu4e-context-current)))
          (when cur
            (if (string= (mu4e-context-name cur)
                         (mu4e-context-name (car mu4e-contexts)))
                cur (prog1 nil (setq policy 'ask)))))
          
	;; no context found yet; consult policy
	(case policy
	  (pick-first (car mu4e-contexts))
	  (ask (mu4e~context-ask-user "Select context: "))
	  (otherwise nil))))))


-- 
Thierry

djcb:
I've made some changes... it's hard to come up with a good system that is both powerful and simple enough for casual users!

Anyway, I think your desired behavior (as discussed above), would be to set `mu4e-compose-context-policy` to `pick-first`. It's not the default, as explained in the commit message.

thierryvolpiatto:

> I've made some changes... it's hard to come up with a good system that
> is both powerful and simple enough for casual users!
>
> Anyway, I think your desired behavior (as discussed above), would be
> to set mu4e-compose-context-policy to pick-first.

Yes, that's what I need.

> It's not the default, as explained in the commit message.

Good, your changes are working except one thing, it is when
`mu4e-compose-mode` runs in `mu4e~compose-handler` it runs again
`mu4e~start` which in turn runs `mu4e~context-autoswitch` with
`mu4e-context-policy` as argument, which works if this one is
`ask-if-none` but fail if it is `pick-first`.

-- 
Thierry

thierryvolpiatto:
This change in `mu4e-compose-mode` is working for me:

```diff
@@ -337 +337 @@ message-thread by removing the In-Reply-To header."
-    (mu4e~start) ;; start mu4e in background, if needed
+    (unless (mu4e-running-p) (mu4e~start)) ;; start mu4e in background, if needed
```

djcb:
Oh, good catch, thanks! I've solved it slightly differently (on the `mu4e~start` side).

thierryvolpiatto:

> Oh, good catch, thanks! I've solved it slightly differently (on the mu4e~start side).

Great, thanks, works great now.

-- 
Thierry

djcb:
Great - closing this one. Thanks for your help!

PR #753, opened 2016-01-03
Show context in main view

thierryvolpiatto:
This allow to show current context in main view buffer.

### DIFF
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -83,32 +83,35 @@ first context for which `match-func' return t."
 If the new context is the same and the current context, only
 switch (run associated functions) when prefix argument FORCE is
 non-nil."
-  (interactive "P")
-  (unless mu4e-contexts
-    (mu4e-error "No contexts defined"))
-  (let* ((names (map 'list (lambda (context)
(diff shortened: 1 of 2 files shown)


djcb:
I had something like this before, but instead opted for the modeline, so we can always accurately show the current context, no matter how the user set it. If we set it in the main-view like this, we might show an outdated context (fixing this would require various extra magic).

Also, the idea would be that we don't bother users with the contexts if they haven't set any up. Which will be the majority for quite some time I think.

thierryvolpiatto:

> I had something like this before, but instead opted for the modeline,
> so we can always accurately show the current context, no matter how
> the user set it. If we set it in the main-view like this, we might
> show an outdated context (fixing this would require various extra
> magic).
>
> Also, the idea would be that we don't bother users with the contexts
> if they haven't set any up. Which will be the majority for quite some
> time I think.

Ok, no problems I will keep showing context in main-view for now for my
own usage as it works fine for me, showing it also in mode-line is nice
too as we can see it in over views.

You have the PR anyway, if you need the code for any reason you can
reopen it.

Thanks.

-- 
Thierry

djcb:
Thanks!

PR #771, opened 2016-01-15
Fixes #764 for CYGWIN systems

### LINKED ISSUE #764: Mu4e and Cygwin: Contacts-List over 64K length
I'm using Cygwin with emacs-w32 and current git revision d9fcfc6ddcbd5edc142ea85a0a0c2dde9d682029 . Recently, my contact list was not updated anymore. I found out the following:

When mu server sends answers that are above 64K in length to mu4e, it seems to me that the part above 64K is just omitted. I believe this is due to some buffering issue (maybe of cygwin), where emacs does not read out the server's IO buffer fast enough and hence data gets lost. 

I have worked around the behaviour by writing out the server answers in chunks of 1K size and waiting some milliseconds in between the runs. I know, this is not the perfect way to do it, but at least it works for me now. 

Here's the new code for the corresponding write function in mu-cmd-server.c:
```
static void G_GNUC_PRINTF(1, 2)
print_expr (const char* frm, ...)
{
	char *expr, *expPtr;
	va_list ap;
	ssize_t rv;
	size_t exprlen, lenlen;
	char cookie[16];
	static int outfd = 0;

	const int writeStep = 1024;
	int bytesToWrite = 0;

	if (outfd == 0)
		outfd = fileno (stdout);

	expr    = NULL;

	va_start (ap, frm);
	exprlen = g_vasprintf (&expr, frm, ap);
	va_end (ap);

	/* this cookie tells the frontend where to expect the next
	 * expression */

	cookie[0] = COOKIE_PRE;
	lenlen = sprintf(cookie + 1, "%x",
			 (unsigned)exprlen + 1); /* + 1 for \n */
	cookie[lenlen + 1] = COOKIE_POST;

	/* write the cookie, ie.
	 *   COOKIE_PRE <len-of-following-sexp-in-hex> COOKIE_POST
	 */
	rv = write (outfd, cookie, lenlen + 2);
	if (rv != -1) {
		expPtr = expr;
		while(exprlen > 0) {
			bytesToWrite = exprlen > writeStep ? writeStep : exprlen;
			rv = write(outfd, expPtr, bytesToWrite);
			expPtr += bytesToWrite;
			exprlen -= bytesToWrite;
			usleep(10000);
			}
		g_free (expr);
	}
	if (rv != -1)
		rv = write (outfd, "\n", 1);
	if (rv == -1) {
		g_critical ("%s: write() failed: %s",
			   __func__, strerror(errno));
		/* terminate ourselves */
		raise (SIGTERM);
	}
}
```

Thank you for producing this great piece of software and I would be glad, if I can help any further with this issue.

mmatthe:
when using mu server, write only 64K bytes at maximum with one write, Split longer strings into several write calls.

### DIFF
--- a/mu/mu-cmd-server.c
+++ b/mu/mu-cmd-server.c
@@ -98,12 +98,17 @@ install_sig_handler (void)
 static void G_GNUC_PRINTF(1, 2)
 print_expr (const char* frm, ...)
 {
-	char *expr;
+	char *expr, *expr_orig;
 	va_list ap;
 	ssize_t rv;
 	size_t exprlen, lenlen;
 	char cookie[16];
 	static int outfd = 0;
+
+#if defined(__CYGWIN__ )&& !defined (_WIN32)
+	const size_t writestep = 4096 * 16;
+	size_t bytestowrite = 0;
+#endif
 
 	if (outfd == 0)
 		outfd = fileno (stdout);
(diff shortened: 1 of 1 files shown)


djcb:
Merged your PR, thanks!

djcb:
Merged your PR, thanks!

PR #777, opened 2016-01-27
fix documentation inconsistency on mu4e-completing-read-function

mmatthe:
The variable is named mu4e-completing-read-function, but the documentation stated mu4e-completing-read. 
Fixed in this commit.

My emacs deletes trailing white-space from each line, so there are some more changed lines.

### DIFF
--- a/mu4e/mu4e.texi
+++ b/mu4e/mu4e.texi
@@ -429,7 +429,7 @@ with @t{MU4E-PATH} replaced with the actual path.
 @section Folders
 
 The next step is to tell @t{mu4e} where it can find your Maildir, and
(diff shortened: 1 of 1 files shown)


djcb:
Appreciated, but please create a PR without the spurious whitespace change, that makes it easier to see what changed (without needing to pass extra options to git).

(note, git actually allows you to create commit out of only selected parts of a file; magit makes this easy). Thanks!

mmatthe:
I have removed the white-space changes, just commiting the necessary change.

djcb:
Excellent, I've merged it!

PR #780, opened 2016-01-30
Add a hook to execution of marks

magthe:
I find it useful to be able to run a hook function on executing marks.

The latter patch is the actual code change, the former is just a change of indentation.

### DIFF
--- a/mu4e/mu4e-mark.el
+++ b/mu4e/mu4e-mark.el
@@ -386,29 +386,29 @@ work well.
 If NO-CONFIRMATION is non-nil, don't ask user for confirmation."
   (interactive)
   (mu4e~mark-in-context
-    (let ((marknum (hash-table-count mu4e~mark-map)))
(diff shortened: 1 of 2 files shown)


djcb:
Hmmm... what's the use-case for this?

magthe:
For me, to run `mu4e-action-retag-message` on `'trash` and `'refile` to make Gmail behave.

djcb:
Thanks - that makes sense. Can you include some update to the 'Marking' section of the manual?

magthe:
Sure, I'll see what I can come up with :)

magthe:
Is that enough for documentation, or should I add something more?

djcb:
No -- looks good! I have merged it, with a few minor changes - most important one, the hook is called `mu4e-mark-execute-pre-hook` now. Anyway, thanks!

magthe:
Excellent! :+1:

PR #784, opened 2016-02-01
fix compose cancel orphaning a buffer

jtmoulia:
This changes `mu4e-draft-open` to create the buffer's initial contents before opening the file. As for the why: if the read prompt is cancelled, e.g. _Reply to..._, there is no hanging buffer.

Might want to double check the indentation .. .


djcb:
Merged -- thanks!

Issue #627, opened 2015-06-10
When mu4e-compose-in-new-frame is set to `t`, should `message-kill-buffer` kill the frame?


zakkak:
When `mu4e-compose-in-new-frame` is set to `t` I would expect/want `message-kill-buffer` to close the new frame as well.

That said, I believe we need a wrapper for `message-kill-buffer` that will allow us to kill the frame as well.

What is your opinion?

panjie:
You can use something like:
```elisp
(defadvice message-kill-buffer (after kill-frame activate)
  (when mu4e-compose-in-new-frame
    (delete-frame)))
```

djcb:
The defadvice could be an good solution for this -- but we should somehow limit it to only mu4e (so it does not affect other possible uses of message-kill-buffer);  perhaps check for the buffer-mode?

zakkak:
That's why I was thinking about a wrapper (with a `mu4e-*` name, e.g., `mu4e-message-kill-buffer`) that we could bind to `C-c C-k` in the `mu4e-compose-mode`.

djcb:
Okay, we're closing the frame in slightly different way now, by adding `delete-frame' as a message-mode action that is executed with killing/postponing/sending a message.

PR #786, opened 2016-02-01
Restore window-layout after `message-kill-buffer'

akermu:
Replaces `message-kill-buffer' with `mu4e-message-kill-buffer'. This new
function is largely a copy of the original and has the same
functionaliy, but it restores the window-layout of mu4e after cancelling
a message in mu4e-compose-mode.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -314,6 +314,7 @@ message-thread by removing the In-Reply-To header."
     (let ((map (make-sparse-keymap)))
       (define-key map (kbd "C-S-u")   'mu4e-update-mail-and-index)
       (define-key map (kbd "C-c C-u") 'mu4e-update-mail-and-index)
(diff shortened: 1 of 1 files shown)


akermu:
Should fix #627.

djcb:
Hmmm, that seems useful, but would it be possible to wrap or advise (rather than copy) the original `message-kill-buffer`? Wouldn't want to duplicate so much code

akermu:
Yeah i thought about that too. A advice would definitly be nicer. But by using an advice it's impossible to check the outcome of the yes-or-no-query in the original function..
Do you have an idea?

akermu:
I updated the PR. Now mu4e-message-kill-buffer is a wrapper around message-kill-buffer.

zakkak:
:+1:

djcb:
Merged -- thanks!

PR #792, opened 2016-02-09
Update mu4e-about.org

sje30:
Add  "showall" to orgmode so that all of the buffer contents are visible.
This matches the NEWS.org file.


djcb:
Merged -- thanks!

PR #793, opened 2016-02-10
mu: escape the name of the part

YoungFrog:
I had a bug where mu4e wouldn't display a message, and instead kept "Waiting for message..." in the buffer. It was a bit annoying to debug because of the async nature of mu4e and because of an `ignore-errors` in `mu4e~proc-eat-sexp-from-buf`. The problem was that mu didn't properly quote some backslashes (in the mail, the name of the attachments had backslashes!), causing `read-from-string` to produce an error, which was ignored because of `ignore-errors`. This PR is a potential fix but I'm not sure it's the right place to do the escaping. Of course feel free to ignore and fix by other means.

[makes me wonder if `mu4e~proc-eat-sexp-from-buf` should perhaps be a tiny bit smarter about which error to ignore...]


djcb:
Oh, good catch! Merged, thanks!

PR #797, opened 2016-02-14
mu4e: escape % in queries when updating the mode-string

wavexx:
Prevent stray % in queries to be interpreted by the modeline formatting.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -998,7 +998,9 @@ the query history stack."
 	global-mode-string
 	'(:eval
 	   (concat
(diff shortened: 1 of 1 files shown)


djcb:
Merged. Good catch, thanks!

wavexx:
Hrrm, there is also the same issue with the context label.
Is there already a built-in function to escape strings for the modeline? (maybe I'm just missing it).

djcb:
On Monday Feb 15 2016, wavexx wrote:

> Hrrm, there is also the same issue with the context label.
> Is there already a built-in function to escape strings for the modeline? (maybe I'm just missing it).

Don't think there is such a function. But feel free to create another PR
:-) Thanks.

PR #800, opened 2016-02-15
mu4e: also quote the context label in the modeline

wavexx:
Couldn't find a better place for the helper function.


djcb:
Merged -- thanks!

PR #801, opened 2016-02-16
mu4e/org-mu4e.el: use message subject for search link

stsquad:
Embedding the actual search string is a little dry especially if your
search is a message-id for the top of thread.

### DIFF
--- a/mu4e/org-mu4e.el
+++ b/mu4e/org-mu4e.el
@@ -78,8 +78,8 @@ Example usage:
 	      desc link)
 	(org-store-link-props :type "mu4e" :query query)
 	(setq
-	  desc (concat "mu4e:query:" query)
-	  link desc)
+          desc (mu4e-message-field-at-point :subject)
+          link (concat "mu4e:query:" query))
 	(org-add-link-props :link link :description desc)
 	link)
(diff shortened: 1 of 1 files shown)


djcb:
Hmmm, but the subject of the message you are currently on is not a very good indicator (it might not even match the query when you execute, e.g. when it's time-based)

stsquad:
@djcb: It does seem a bit redundant to repeat the search terms in the link title and certainly most of my links are to mail threads where the search terms provide no human readable information. However I understand there are other links people use ;-)

Two options I can think of:
  - heuristics in the link function to decide what to use
  - prefix detection to use the current message-at-point subject for the link

What do you think?

djcb:
I think the query itself is the best default.  When you use it e.g. with org-capture / org-store-link, you can change the description to something else.

stsquad:
Is there any way the description can be seeded from the previous view, otherwise it's very fiddly to type the whole thing back in again.

djcb:
I rarely use the link-to-query (instead just link to a specific message), but you could stick the subject of the message-at-point in the kill-buffer, and use that in some org-capture template.

Issue #802, opened 2016-02-16
add API to toggle autoupdating


gaydenko:
Hi! It would be handy to toggle autoupdating without restarting mu4e. Say, during break autoupdating needs to be turned off, but all other Emacs frames/buffers/windows must be kept.
Emacs daemon is in use, and direct `mu4e-quit` with `mu4e` resulted in error related to mu server lock. Code I have tried:
```lisp
(defvar my-mu4e-autoupdate-mark " mu ")
(push '(:eval my-mu4e-autoupdate-mark) mode-line-misc-info)

(defun my-mu4e-set-autoupdate-mark (mark)
  "Set mode line autoupdate mark to MARK."
  (setq my-mu4e-autoupdate-mark mark)
  (force-mode-line-update t))

(my-mu4e-set-autoupdate-mark " mu ")

(defun my-mu4e-autoupdate-toggle ()
  "Toggle mu4e autoupdate mode."
  (interactive)
  (if (integerp mu4e-update-interval)
      (progn (my-mu4e-set-autoupdate-mark " mu ")
	     (setq mu4e-update-interval nil)
	     (mu4e-quit)
	     (mu4e)
	     (mu4e-message "auto-update is off"))
    (progn (my-mu4e-set-autoupdate-mark " MU ")
	   (setq mu4e-update-interval 300)
	   (mu4e-quit)
	   (mu4e)
	   (mu4e-message "auto-update is on"))))

(global-set-key (kbd "C-S-M-u") 'my-mu4e-autoupdate-toggle)
```

djcb:
I think it's a bit too much added complexity for little gain, and the risk of forgetting to re-enable the auto-updating. I would simply quite mu4e for a while. 

Cheers,
Dirk.

PR #810, opened 2016-02-27
mu4e: add mu4e-index-update-in-background

j-hao:
Give users choice to decide if they want to run automatic mu4e-index-update in background or not

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -828,7 +828,7 @@ successful, call FUNC (if non-nil) afterwards."
 		(setq mu4e~update-timer
 		  (run-at-time
 		    0 mu4e-update-interval
-		    (lambda () (mu4e-update-mail-and-index t)))))
+		    (lambda () (mu4e-update-mail-and-index mu4e-index-update-in-background)))))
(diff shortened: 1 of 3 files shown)


gaydenko:
@j-hao , will it resolve https://github.com/djcb/mu/issues/802 ?

j-hao:
@gaydenko, no, I don't think so. This is just let user to be able to choose the auto update to run in background or foreground.
Personally, I use a dedicate emacs instance for mu4e and all other works to emacs server. I have following config in my .emacs:
```
  (if (member "-m" command-line-args)
      (progn
        (delete "-m" command-line-args)
        (message "mu4e mode")
        (mu4e))))
```
And run ```emacs -m``` to start mu4e in a dedicate window

gaydenko:
@j-hao , aha, I see, thanks. So, to get what I want that `lambda` must be wrapped with predicate using another custom var :)

djcb:
@j-hao: looks useful... can you include a little update for the manual, ie. section "Retrieval and indexing with mu4e"?

Thanks.

j-hao:
Updated manual

djcb:
Merged. Thank you!

PR #813, opened 2016-03-05
Update calls to mu:get-parts after name change

jsitnicki:
In commit 8e3fbe380ec0 mu:get-parts was renamed to mu:c:get-parts but
get-parts method was not updated. Fix it.


djcb:
Merged -- thanks!

PR #816, opened 2016-03-14
mu4e: quote user-full-name according to rfc822 when construct from field

j-hao:
If user-full-name is set to this format: ```Last Name, First Name```, without proper quotation, SMTP will use ```Last Name``` for from field which is wrong.


djcb:
Looks goods, merged. Thanks!

PR #821, opened 2016-03-22
Align table of marks in docs

hinrik:
Small aesthetic fix.


djcb:
Ah, thanks for noticing. There's should be a `+` at the beginning, but I've fixed the table.

PR #831, opened 2016-04-16
Handle attached images when write msg body to html

j-hao:
This will make view in browse works with attached images


djcb:
Looks good, but the "cid" in "each_part" must be g_free'd. 

I actually noticed that some of the other variables need freeing as well, so I've updated the code.

j-hao:
Updated to g_free cid

djcb:
Merged, thanks!

PR #833, opened 2016-04-19
Add new face for subject in mu4e-view

odi:
I would like to have a different face for the subject line in the mu4e-view.

This adds a new face for the subject in the mu4e-view. The default configuration inherits from mu4e-header-value-face, that means that there is no visual difference to the old configuration.

### DIFF
--- a/mu4e/mu4e-vars.el
+++ b/mu4e/mu4e-vars.el
@@ -623,6 +623,11 @@ mu4e-compose-mode."
   "Face for highlighting marked region in mu4e-view buffer."
   :group 'mu4e-faces)
 
+(defface mu4e-header-subject-face
+  '((t :inherit mu4e-header-value-face))
+  "Face for subject header."
+  :group 'mu4e-faces)
+
 ;; headers info
 (defconst mu4e-header-info
   '( (:attachments .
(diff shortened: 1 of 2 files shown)


davidcsterratt:
Thanks for the work, odi. This would be a feature I would like to see.

Ambrevar:
Anything blocking a merge?

antoine-levitt:
Can you explain the close? It looks like this is something quite a few people want, and I don't think it's implemented in master?

djcb:
It's obsolete for the gnus-based view.

antoine-levitt:
Is the old view going to be removed in future versions then? I like it better (and don't want to take the time to configure the gnus view to my liking)

djcb:
Yes.

antoine-levitt:
:(

Chris00:
@antoine-levitt What don't you like in the Gnus based view?  It supports many messages that were improperly decoded by the old version.

antoine-levitt:
Oh it's fine but I don't like the faces and it messes up with my hardwired patterns. More generally I feel like mu4e is breaking a lot of stuff and it's taking a lot of my energy to try to keep up - it's nice to see it move forward, but it does take time to adapt.

PR #834, opened 2016-04-19
Add spam-filtering functions suitable for the view mode

Chris00:
In `mu4e-contrib` there were spam-filtering functions suitable for the `headers` mode but not for the `view` mode.  This alleviate this asymmetry...


djcb:
Merged -- thanks!

PR #841, opened 2016-05-03
Only ignore the "/mu/mu" executable

liweitianux:
Currently the whole directory "/mu" are ignored, thus we have to force git to track the files in that directory.


djcb:
Merged -- thanks!

Issue #837, opened 2016-04-26
Remove draft backup files


Chris00:
When `message-kill-buffer-on-exit` is set to `t`, the backup file of the draft should be removed as well otherwise there will be many backup files (`*~`) in the draft folder.

djcb:
Please always mentions the versions of mu/emacs etc.

Seems they're deleted for me.... perhaps some other variable involved? Can you reproduce with just running emacs -Q and mu4e?

Chris00:
Sorry, I forgot.  I use 0.9.16.

Chris00:
On 2016-04-26, Dirk-Jan C. Binnema wrote:
> 
> Seems they're deleted for me.... perhaps some other variable involved? 
> Can you reproduce with just running emacs -Q and mu4e?

I removed the *~ files from the draft folder and will see if it is OK 
now (I am thinking they may be leftovers prior to me activating 
message-kill-buffer-on-exit).

Chris00:
I tried with the Git version of mu4e and `--Q` but when sending a message the backup file in `mu4e-drafts-folder` is *not* removed.

Chris00:
One possibility to fix this is to disable backup files for drafts with `(setq (make-local-variable 'make-backup-files) nil)`.

Chris00:
See the [pull request](https://github.com/djcb/mu/pull/842) for another — more satisfactory — solution.

djcb:
mu4e depends on the emacs settings for backups/auto-save files for this, you could configure those. In particular, `backup-directory-alist` and `auto-save-file-name-transforms`.

PR #842, opened 2016-05-06
Remove the backup file of a draft message when sending it

Chris00:
Fixes https://github.com/djcb/mu/issues/837

The solution is a bit unsatisfactory because it only removes the most recent backup file (while one would like to remove all associated backups to the file) but it works well in practice.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -532,10 +532,10 @@ the appropriate flag at the message forwarded or replied-to."
     (mu4e~proc-remove docid))
   ;; kill any remaining buffers for the draft file, or they will hang around...
   ;; this seems a bit hamfisted...
-  (dolist (buf (buffer-list))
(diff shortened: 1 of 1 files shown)


djcb:
I think backup of e-mail message can be treated like other files and honor emacs settings for that, I'd be a bit hesitant to work around that.

The only things is the question where it keeps those backups/auto-saves; you might want to set `backup-directory-alist` and `auto-save-file-name-transforms` for that.

Chris00:
> honor emacs settings for that, I'd be a bit hesitant to work around that.

I agree that the above proposed solution is not the ideal one but the difference here with other buffers is that Emacs will remove the underlying file upon sending the email (thus, if `mu4e` is responsible for the existence of the file, shouldn't it also be responsible for its siblings?).  Maybe one way out is to create the auto-save and backup files for draft files into the trash folder (maybe prefixing them with `draft-`)?

PR #844, opened 2016-05-06
Compose mails with format=flowed by default

Chris00:
The default is expressed by the variable `mu4e-compose-format-flowed`, so it is easy to change.  However, given the variety of devices people use to read mails, I've considered that a format that respect that is better (however some important services — gmail, github — do not understand format=flowed so maybe we want the "long lines" mode to be the default).

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -340,15 +340,6 @@ simply executes `fill-paragraph'."
 	(fill-paragraph nil region))
     (fill-paragraph nil region)))
 
-(define-key-after
-  (lookup-key text-mode-map [menu-bar text])
-  [mu4e-hard-newlines]
-  '(menu-item "Format=flowed" mu4e-toggle-use-hard-newlines
-	      :button (:toggle . use-hard-newlines)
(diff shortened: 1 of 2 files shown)


djcb:
Looks good -- but could you add a little update for the manual as well? (`mu4e.texi`). There's an existing FAQ about that, so I suppose you could simply update that. Thanks!

Chris00:
I wonder whether `M-q` shouldn't unwrap paragraphs into a single line regardless of the values of `use-hard-newlines`.  Or maybe, when `use-hard-newlines` evals to false, `M-q` should be something like [`compact-uncompact-block`](http://ergoemacs.org/emacs/modernization_fill-paragraph.html)?

djcb:
One more change request -- at least for now, I'd like to have `mu4e-compose-format-flowed` default to `nil`.

format=flowed is a behavioral change that makes things work less well in some common cases (like gmail), and I'd like to avoid unpleasantyl surprising users.

Chris00:
Done.

djcb:
Thanks -- merged!

PR #845, opened 2016-05-06
mu4e-actions: improve mbox handling for multiple marks

stsquad:
This avoids prompting the user over and over if the directory hasn't
changed.


djcb:
Ah, looks useful -- but can you quote the paths passed to the shell, ie. `shell-quote-argument`? And MSG doesn't need quoting in the docstring. Thanks.

stsquad:
I've submitted a new pull request with the update and another fix.

PR #847, opened 2016-05-10
Add a menu entry to attach captured messages

Chris00:
Without this menu entry, one has to remember the command name which is
not convenient.  This entry also helps to discover the "capture" feature.


djcb:
Merged -- thanks!

Issue #843, opened 2016-05-06
Shouldn't mu4e-message-kill-buffer remove the draft file?


Chris00:
When `C-c C-k` (`mu4e-message-kill-buffer`) on a draft message, its file is not deleted and so it remains in the draft folder.  Is this the desired behavior??

djcb:
I wouldn't expect anything else, no; the function simply kills the buffer as its name suggests.

Chris00:
While I agree for the function, my question was more about the action of the binding `C-c C-k` which is described in the manual as
```
C-c C-k      kill the message
```
So maybe the manual should be more precise about the action of that binding.  Alternatively, `C-c C-k` could ask whether to also remove the message from the draft folder.

djcb:
`mu4e-message-kill-buffer` is just some window restoration and  `message-kill-buffer` (its docstring mentions this) which it reuses from gnus/message-mode, so we're just following its behavior. Of course, we could do something different, but in this case I think it's quite okay.

PR #850, opened 2016-05-15
Improve the description of C-c C-k in compose mode

Chris00:
Fixes https://github.com/djcb/mu/issues/843
But not the only possible way out — please comment.


djcb:
Merged -- thanks.

PR #852, opened 2016-05-19
Improve behavior of mu4e-action-show-thread

jschwab:
This patch improves the behavior of mu4e-action-show-thread.  This
action now leaves the point on the message where the action was invoked,
which helps prevent losing ones' place in a long thread.  When invoked
in view mode, it continues to display the message that was being viewed,
instead of returning to a header-only view.


djcb:
Merged -- thanks!

PR #853, opened 2016-05-23
Use Unicode characters when cleaning up attachment filename

liweitianux:
Improve the function ``cleanup_filename()`` of ``lib/mu-msg-part.c`` to use Unicode characters when replacing the control characters, slashes and colons with ``-``.

Originally, this function just use plain C characters (i.e., assuming ASCII string) when checking each character is or not a control character, slash or colon.  However, when the attachment filename contains non-ASCII (e.g., Chinese characters), all the non-ASCII characters are replaced with ``-``.

For example:
* Before:
```
> mu view test_chinese_attachment_filename.eml
From: Tester <tester@example.com>
To: Example <example@example.com>
Subject: Test email with attachment of Chinese filename
Date: Mon 23 May 2016 05:22:09 PM CST
Attachments: 'attachment-test.txt', '------------.txt', '-------test.txt'
Hello,

This is a simple test email with three attachments:

1. `attachment:test.txt`: filename is all English;
2. `测试附件.txt`: filename is all Chinese (exclude the extension);
3. `附件-test.txt`: filename mixes Chinese and English.
```

* After:
```
> ./build/mu/mu/mu view test_chinese_attachment_filename.eml
From: Tester <tester@example.com>
To: Example <example@example.com>
Subject: Test email with attachment of Chinese filename
Date: Mon 23 May 2016 05:22:09 PM CST
Attachments: 'attachment-test.txt', '测试附件.txt', '附件-test.txt'
Hello,

This is a simple test email with three attachments:

1. `attachment:test.txt`: filename is all English;
2. `测试附件.txt`: filename is all Chinese (exclude the extension);
3. `附件-test.txt`: filename mixes Chinese and English.
```

### DIFF
--- a/lib/mu-msg-part.c
+++ b/lib/mu-msg-part.c
@@ -239,15 +239,26 @@ get_part_size (GMimePart *part)
 }
 
 
-static void
+static char*
 cleanup_filename (char *fname)
 {
+	GString *gstr;
 	gchar *cur;
+	gunichar uc;
 
-	/* replace control characters, slashes,colons by a '-' */
-	for (cur = fname; *cur; ++cur)
-		if (*cur < ' ' || *cur == '/' || *cur == ':')
-			*cur = '-';
+	gstr = g_string_sized_new (strlen (fname));
+
+	/* replace control characters, slashes, and colons by '-' */
+	for (cur = fname; cur && *cur; cur = g_utf8_next_char (cur)) {
+		uc = g_utf8_get_char (cur);
+		if (g_unichar_iscntrl (uc) || uc == '/' || uc == ':')
+			g_string_append_unichar (gstr, '-');
+		else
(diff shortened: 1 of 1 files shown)


liweitianux:
Hello, @djcb 

How about my pull request? Is there any problem that needs improvement?  Wish this can be merged, as I'm a Chinese user and it is annoying that all the Chinese characters in the attachment filename are replaced by the ``-``.  Thanks!

P.S.: I checked the continuous integration test failure, and it seems due to the environment bootstrap.  I think this pull request can be cleanly merged.  (I do not know how to re-trigger the continuous integration test.)

Regards!

djcb:
Hi @liweitianux: thanks for your PR -- was a bit busy, so only got to it now!

Anyway, looks like a useful improvement, thanks! So let's merge it.

PR #858, opened 2016-05-30
mu4e: Make dvipng optional when convert from org to html

j-hao:
For users don't have latex installed but still want to use org-mu4e-compose-org-mode


djcb:
Looks good, except that "latex" should be spelled "LaTeX" in the docstring. Thanks!

j-hao:
Updated

djcb:
Merged, thanks!

PR #867, opened 2016-06-09
Fix window restoration when C-g'ing on reply.

thierryvolpiatto:
* mu4e/mu4e-compose.el (mu4e~compose-handler):
Save and restore window config.


thierryvolpiatto:
Hi,
when replying to a mail (R) and then aborting with C-g the header buffer is duplicated, this patch fix this problem.

djcb:
Merged -- thanks!

PR #871, opened 2016-06-21
[FIX] references header dropped bug

arthurl:
Hi Dirk,

I was wondering why threading in `mu` suddenly stopped working sporadically since about 2 weeks ago… so after some bisecting, here it the fix. Basically, the “References” header have been erroneously dropped for all messages composed with less than `mu4e~max-reference-num` number of reference.

I guess there are really [too many parenthesis](https://xkcd.com/297/) in lisp :p

Arthur

### DIFF
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -141,8 +141,8 @@ References. If both are empty, return nil."
     ;; remove some refs when there are too many
     (when (> refnum mu4e~max-reference-num)
       (let ((surplus (- refnum mu4e~max-reference-num)))
-	(mu4e~shorten-1 refs cut surplus))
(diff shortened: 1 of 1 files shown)


djcb:
Merged -- thanks! What a silly bug...

arthurl:
Haha it happens. Thanks for the good work!

Arthur

mankoff:
If we now have a month of messages with threads broken because of this bug, is there any way to recover the thread? Upgrading and sending a new email it seems the bug is fixed for the new email, but not for existing ones :(.

arthurl:
Unfortunately, the problem occurs when composing messages (specifically, the creation of the references header). It's not simply a problem with display.

As such, I'm not even sure "fixing" it makes sense, because it will involve changing historic data. Keep in mind that the email is also mangled for your recipient, and you certainly can't fix it on their end.

Issue #827, opened 2016-04-06
mu4e: update progress buffer doesn't escape ansi code


seanfarley:
## Expected or desired behavior

Ideally, I'd like the output of the update buffer to escape ansi codes (if possible).

## Actual behavior

In the update buffer, this is the output I see with using mbsync:

`
C: 0/1  B: 0/0  M: +0/0 *0/0 #0/0  S: +0/0 *0/0 #0/0^MC: 0/1  B: 0/10  M: +0/0 *0/0 #0/0  S: +0/0 *0/0 #0/0^MC: 0/1  B: 1/10  M: +0/0 *0/0 #0/0  S: +0/0 *0/0 #0/0^MC: 0/1  B: 2/10  M: +0/0 *0/0 #0/0  S: +0/0 *0/0 #0/0^M
`

(notice the ^M characters)

## Steps to reproduce

Use mbsync as the `mu4e-get-mail-command`.

## Versions of mu, mu4e/emacs, operating system etc.

Latest commit `69780d2e02110dcb05f200b6cb7f441ca527ec46`, Mac OS X El Capitan, Emacs 24.5.1

## Any other detail

Or maybe there is a way to use lisp to parse this output so that it can match the text seen in the minibuffer that mu4e outputs (e.g. "Retrieving mail ... /-\|" etc)?

djcb:
That doesn't seem related to ansi-codes; `^M` is typically used to represent Windows-style line ending, ie. '\r\n' rather than '\n'. I think the question is why the newlines are not seen or not handled correctly in the process output buffer.

Can you try to evaluate: `(start-process-shell-command "my-test" "*my-test*" "mbsync -a")`, and check what the "*my-test*" buffer look like after that?

seanfarley:
Yeah, it displays the same ^M characters. I guess if I want pretty output I should wrap the mbsync call with some lisp and parse the output?

seanfarley:
Digging in further, mbsync outputs `\r` which resets the cursor to the beginning of the line in the shell but in an emacs buffer `\r` is represented as `^M` (mostly just putting this here for future archeologists).

djcb:
No lisp required, I think. You can put some shell expression in `mu4e-get-mail-command` (e.g. with `tr`) to clear up the output. Or you can write a little shell-script that calls mbsync, and sanitizes the output.

seanfarley:
I actually did try that but it caused buffering problems :-(

seanfarley:
[...]

PR #877, opened 2016-07-08
Handle mbsync output in mu4e-update buffer

remvee:
The mbsync command uses carriage return characters to go to the beginning of the line.  This change does the same for the mu4e-update buffer.  See also #827.


djcb:
Merged, looks useful -- thanks!

PR #886, opened 2016-07-20
mu4e-proc: Add a hook before we call start-command in mu4e~proc-start

avar:
I'm using this to "killall mu" before I start mu again, see
<CACBZZX6ntW2uF5MPr-LaPM5HMiQzTgbUHyu6njJ1iX0pT+mGfA@mail.gmail.com> (The
dreaded "Waiting for message..." message and what to do about it")
message on the mu mailing list for more details.

The hook I'm using is:

  (add-hook
   'mu4e~proc-start-hook ; (setq mu4e~proc-start-hook nil)
   '(lambda ()
      (message "Now running the 'killall mu' hook!")
      (shell-command "killall mu")
      (sleep-for 0 250)))

Coupled with:

  (setq mu4e-update-interval nil)

And this monitoring script:

  https://github.com/avar/dotemacs/blob/master/scripts/mu-monitor.pl

While running these two concurrently in a screen session:

    while sleep 1; do time sudo perl ~/g/dotemacs/scripts/mu-monitor.pl; done
    while sleep 30; do time mu index --maildir=~/Maildir/work --max-msg-size=500000000; done

### DIFF
--- a/mu4e/mu4e-proc.el
+++ b/mu4e/mu4e-proc.el
@@ -50,6 +50,9 @@ a length cookie:
   (concat mu4e~cookie-pre "\\([[:xdigit:]]+\\)" mu4e~cookie-post)
   "Regular expression matching the length cookie.
 Match 1 will be the length (in hex).")
+(defvar mu4e~proc-start-hook nil
+  "Hook run just before `mu4e~proc-start' calls `start-process'
+our 'mu server' process.")
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 (defsubst mu4e~proc-send-command (frm &rest args)
@@ -70,6 +73,7 @@ Start the process if needed."
 	  (args (append args (when mu4e-mu-home
 			       (list (concat "--muhome=" mu4e-mu-home))))))
     (setq mu4e~proc-buf "")
+    (run-hooks 'mu4e~proc-start-hook)
     (setq mu4e~proc-process (apply 'start-process
 			      mu4e~proc-name mu4e~proc-name
 			      mu4e-mu-binary args))


djcb:
`mu4e-get-mail-command` already is the hook-mechanism you can use for that -- use a shellscript that does whatever killing / mail fetching you need.

avar:
I may be missing something, but isn't `mu4e-get-mail-command` different? That's the thing that's invoked every N seconds where N = mu4e-update-interval.

For this "index in background but be responsive" approach to work I can't be killing mu from that, I need to be killing it from the bit in mu4e that goes "hang on, I don't have a mu-server process, better make one", which is why I'm hooking into mu4e~proc-start.

djcb:
Hmmm... such hacks really shouldn't be necessary... Did you try the recent tricks for making indexing a bit quicker? Ie. `--nocleanup` and `--lazycheck` (or their mu4e counterparts).

If you really want to hook into arbitrary functions, I'd recommend using the "advise" mechanism, which exists there for such things.

avar:
Yeah let's just close this pull request. I've since stopped using my hack since while it worked most of the time there were some extra nasty race conditions where a mu process would be left locked.

I didn't try the new --nocleanup or --lazycheck but I read the patch for the latter. It looks like it's just using stat() on the directory, in my case I have one huge INBOX folder which is always going to be modified by the time mu4e runs (I get a lot of mail) so it wouldn't do anything as an optimization for me, but I'll check out `--nocleanup`.

One thing that would make this really much more performant would be if mu-index could be fed a list of files that have been removed/added/changed since last time, which e.g. offlineimap could be patched to emit.

avar:
FWIW I'm asking that this gets implemented in offlineimap: https://github.com/OfflineIMAP/offlineimap/issues/367 / might implement it myself.

djcb:
Okay, closing this. Thanks for the effort - hope you get it fixed in offlineimap.

Indeed, the stat-shortcut doesn't help much for the huge-inbox case. No-cleanup should still give you some speedups. I actually try to keep my inbox as small as possible, and move messages elsewhere when I've dealt with them.

avar:
Yeah thanks, I don't think this should be included. I *do* think there's probably room for some functionality in mu-daemon similar to e.g. emacs-daemon where multiple things can connect to it, and it'll go "oh a background index is running and someone wants a foreground command? Screw that, let's kill the background thing and serve the interactive command'.

What I had was a very bad buggy hack to make that happen, but I think the general concept makes sense.

I'll probably experiment with indexing via `run-with-idle-timer` as a workaround.

PR #889, opened 2016-07-25
Add --expr flag to constrain search when looking for duplicates

sje30:
(mu:for-each-message ...) allows an optional expression to be passed
so that only messages matching that expression are searched when
looking for duplicates.

This patch adds a --expr flag so that e.g.

    find-dups.scm --expr d:6m..3m

will reduce the range of messages to those between 3 and 6 months ago
to search for duplicates.  I found this useful when using expressions
to find messages in a particular year, rather than searching the whole
database.  This is often quicker and less worrying than searching the
whole database.

If --expr is not provided, expr should default to #t so that the whole
database is searched as before.

### DIFF
--- a/guile/scripts/find-dups.scm
+++ b/guile/scripts/find-dups.scm
@@ -34,7 +34,7 @@ exec guile -e main -s $0 $@
     (close-pipe port)
     md5))
  
-(define (find-dups delete)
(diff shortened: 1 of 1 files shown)


djcb:
Merged -- thanks!

PR #891, opened 2016-07-27
Fix kill ring clobbering by get mail filter

remvee:
The kill ring fills up with lines like:

  C: 0/1  B: 28/29  M: +0/0 *0/0 #0/0  S: +2/2 *1/1 #0/0

when using mbsync or another tool using carriage return for progress.


djcb:
Merged -- thanks!

PR #893, opened 2016-07-29
mu4e: Allow header fields with more than one lines to fold to one

j-hao:
Press `TAB` within header field key to toggle


djcb:
Finally had some time to check this out... works great! Thanks a lot.

Issue #894, opened 2016-07-31
A regression pull request #831 broke mu4e-action-view-in-browser


avar:
There's a regression after pull request #831 from @j-hao. On the current master 49bc960 if you follow the "FAQ -> Reading messages" documentation and do:

      (add-to-list
       'mu4e-view-actions
       '("View in browser" . mu4e-action-view-in-browser) t)

Reverting bd0fd4b & d16957d fixes the issue. I reverted the former purely because it's needed due to conflicts with the second, but d16957d introduced the issue according to a bisect, opening the HTML in the browser before that worked just fine. Now I get the "Format specifier doesn't match argument type" error.

djcb:
Works fine for me... can you generate a backtrace?  I.e.  M-x toggle-debug-on-error, then reproduce.

avar:
@djcb: I'm bad at elisp, didn't know about `toggle-debug-on-error` the issue became obvious once I did, fixed in pull request #895.

djcb:
Thanks, it has been merged.

PR #895, opened 2016-07-31
Fix a regression in pull request #831

avar:
The mu4e-message-field function was called in a way that would never
work, fix that by calling it correctly.

There's the additional follow-up TODO here that the mu4e-message-field
function itself should probably die on this sort of invocation, but I
don't know enough about elisp idioms to know how that should look.

This fixes my issue #894.


djcb:
Yes, that's clearly a good fix. Thanks a lot!

PR #896, opened 2016-08-01
Actually make the "write attachments" feature in d16957d work

avar:
This is a follow-up to my pull request #895 which fixes another bug in
pull request #831 (d16957d).

The code to write out the attachments would never work, for what it's
worth it's clear from the issue I fixed in #895 that the codepath had
never been executed as-is.

It would find the attachments and try to write them out to /tmp/, just
that, no /tmp/NAME, just the directory itself. That would yield an
error of trying to write to a directory.

Fix that, now we create a temporary name as a function of the
attachment and both save it and extract it.

This makes the mu4e-action-view-in-browser function finally work for
me. It'll now write out the attachments to /tmp, and rewrite the HTML
so that I'll see the attachments in my browser.


j-hao:
Nice finding. I was wondering why sometimes this fails for me :+1:

djcb:
Merged -- nice fix, thanks.

PR #897, opened 2016-08-02
mu4e: use builtin function "shr-render-region" for simplicity

liweitianux:
Hi,

Emacs already has the function ``shr-render-region`` which requires Emacs been built with ``libxml2`` option.  This built-in function also reports better error message if the ``libxml2`` option was disabled.

Also some minor space fixes.

Please note, for unknown reason, Github does **NOT** show my changes correctly.   Please ``diff`` by yourself.  Thanks!

Cheers,
Aly

### DIFF
--- a/mu4e/mu4e-message.el
+++ b/mu4e/mu4e-message.el
@@ -36,7 +36,7 @@
 
 (defcustom mu4e-html2text-command
   (if (fboundp 'shr-insert-document) 'mu4e-shr2text 'html2text)
-  
+
   "Either a shell command or a function that converts from html to plain text.
 
 If it is a shell-command, the command reads html from standard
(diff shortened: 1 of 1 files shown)


Chris00:
P.S. I wrote to GitHub support about the diff problem but, after a couple of mails exchanged with them, I have not had a satisfactory answer (they do not even seem to *see* the problem...).

Chris00:
FYI, I got an answer.  Here is their explanation:
> It seems the issue is coming from the carriage return at the line below:
> https://github.com/liweitianux/mu/blob/9b54c2745eaedb11ab14a66bbe5b7b8227e60c1f/mu4e/mu4e-message.el#L204-L205
> 
> Currently, we are treating the carriage return as a line ending, which is throwing the line numbers off when displaying the diff.

djcb:
Merged -- thanks!

PR #906, opened 2016-08-11
Add a face for displaying the context in the mode-line

Chris00:
When the background is dark, the mode-line background may not be.  That means that the default color chosen to display the context (`mu4e-title-face`) may not be the best one (and may not be easily customizable because it is tied to the way titles are displayed on dark background).  This PR adds a new color for the context (initialized with the old face so nothing changes by default).

emacs1


djcb:
Looks good; merged. Thanks!

PR #907, opened 2016-08-16
Compare email addresses case insensitively

Chris00:
The local part of email addresses is supposed to be case sensitive according to the standard but this is typically not respected (for good reasons).  The domain part, on the other hand, should be matched case insensitively.

This PR takes the stand that the whole address should be matched case insensitively because this is the common practice.


djcb:
Oh, that looks useful... but can you make it lambda instead of a separate function?

Chris00:
Done.

djcb:
Merged -- thanks

PR #909, opened 2016-08-17
Adding eshell/mu4e-attach allowing attachments from eshell.

sachdevaprash:
Another way to attach files that I find useful.


djcb:
This could be useful, but currently the code seems to be dependent on gnus... can you make it independent of that?

sachdevaprash:
Updated. No more gnus dependencies. I had to get a util function from gnus to get active composition buffers, so a new function has been added to mu4e-utils.el.

djcb:
Thank you! Merged.

PR #911, opened 2016-08-22
add rudimentary completion support for retag action

aroig:
This adds a variable mu4e-action-tags-completion-list, that contains a list of commonly used tags to suggest as completion terms during a retag actions.

Along the way, the retag action accepts as argument a comma-separated list of +tag and -tag keywords, instead of a space-separated one, removing the need to quote tags with spaces in them, and making it consistent with the behaviour of completing-read-multiple.


djcb:
Nice - thanks.

PR #914, opened 2016-08-28
mu4e: show '...' when header field folded and keep text-properties

j-hao:
Suggested by Christophe from https://github.com/djcb/mu/commit/250793317669f415385623082cf91d35c28c16f5#commitcomment-18804074


djcb:
Merged, thanks!

PR #917, opened 2016-09-04
For the upcoming org-mode version 9 org-add-link-type is obsolete, so…

kkholst:
… if org-link-set-parameters is bound we will use that instead.

### DIFF
--- a/mu4e/org-mu4e.el
+++ b/mu4e/org-mu4e.el
@@ -120,6 +120,14 @@ Example usage:
 (org-add-link-type "mu4e" 'org-mu4e-open)
 (add-hook 'org-store-link-functions 'org-mu4e-store-link)
 
+;; org-add-link-type is obsolete as of org-mode 9.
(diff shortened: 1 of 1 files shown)


kkholst:
Can see this has also been raised as an issue #918

djcb:
Merged -- thanks!

andersjohansson:
Should it really do both `org-add-link-type` and `org-link-set-parameters`? Shouldn't it be:
````Elisp
;; org-add-link-type is obsolete as of org-mode 9.
;; Instead we will use the org-link-set-parameters method
(if (fboundp 'org-link-set-parameters)
    (org-link-set-parameters "mu4e"
                             :follow #'org-mu4e-open
                             :store #'org-mu4e-store-link)
  (org-add-link-type "mu4e" 'org-mu4e-open)
  (add-hook 'org-store-link-functions 'org-mu4e-store-link))
````

PR #920, opened 2016-09-13
Only call old org-link API if new unavailable

andersjohansson:
Fix for #917


djcb:
Looks good -- thanks.

Issue #929, opened 2016-10-16
mu4e eats display-time-mode


antoine-levitt:
It's me again with yet another cosmetic patch ;-)

When display-time-mode is on, mu4e doesn't display time in any of its buffers. This possibly applies to other modes as well. That's because it (setq)s global-mode-line, which overwrites everything.

Here's a patch against master that uses add-to-list instead:

```
diff --git a/mu4e-compose.el b/mu4e-compose.el
index a2ec9ed..31e545b 100644
--- a/mu4e-compose.el
+++ b/mu4e-compose.el
@@ -402,7 +402,8 @@ buffers; lets remap its faces so it uses the ones for mu4e."
 \\{message-mode-map}."
   (progn
     (use-local-map mu4e-compose-mode-map)
-    (set (make-local-variable 'global-mode-string) '(:eval (mu4e-context-label)))
+    (make-local-variable 'global-mode-string)
+    (add-to-list 'global-mode-string '(:eval (mu4e-context-label)))
     (set (make-local-variable 'message-signature) mu4e-compose-signature)
     ;; set this to allow mu4e to work when gnus-agent is unplugged in gnus
     (set (make-local-variable 'message-send-mail-real-function) nil)
diff --git a/mu4e-headers.el b/mu4e-headers.el
index 73bba68..c711cdf 100644
--- a/mu4e-headers.el
+++ b/mu4e-headers.el
@@ -1082,15 +1082,15 @@ the query history stack."
       (setq
    mu4e~headers-buffer buf
    mode-name "mu4e-headers"
-   mu4e~headers-last-query expr
-   global-mode-string
-   '(:eval
-      (concat
-       (propertize
-        (mu4e~quote-for-modeline mu4e~headers-last-query)
-        'face 'mu4e-modeline-face)
-        " "
-        (mu4e-context-label)))))
+   mu4e~headers-last-query expr)
+      (add-to-list 'global-mode-string
+                   '(:eval
+                     (concat
+                      (propertize
+                       (mu4e~quote-for-modeline mu4e~headers-last-query)
+                       'face 'mu4e-modeline-face)
+                      " "
+                      (mu4e-context-label)))))

     (switch-to-buffer buf)
     (run-hook-with-args 'mu4e-headers-search-hook expr)
diff --git a/mu4e-main.el b/mu4e-main.el
index a647a2d..72b5288 100644
--- a/mu4e-main.el
+++ b/mu4e-main.el
@@ -75,7 +75,8 @@
[...]

PR #934, opened 2016-10-26
Don't erase global-mode-string with context label, but add to it

antoine-levitt:
See #929


antoine-levitt:
Ah, sorry, I added to this pull request instead of doing another one. Not very familiar with these yet. 

Anyway, the second patch is because I kept wondering why, when I saved my draft titled "hi", I got "hi<2>"  and "hi" alternatively. This fixes that, and it should not introduce any side effect.

djcb:
Merged -- thanks!

PR #943, opened 2016-10-31
Support Emacs 26 in configure.ac

xuchunyang:
The version under development is now 26. I installed Emacs from source
via Git recently, it gives:

emacs-version
     => "26.0.50.2"


j0ni:
I hit this and had to install mu from source with this tweak to get it to build and install mu4e. This appears to be a workable solution for now - is there a better one? Or can you merge this so the downstream Arch AUR package `mu-git` works again?

djcb:
Merged -- thanks!

PR #952, opened 2016-11-18
Org mu4e compose stabilization

titaniumbones:
Hi Dirk-Jan,

Two small patches which I think fix some of the stability issuesi n org-mu4e-compose-org-mode.  With these patches, I can now add a hook to compose in html by default. I know this is not best practice, but I think there are many use cases for it, and while it would be better to fix org-mu4e-compose-org-mode entirely, this may help a fair number of people.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,7 +130,7 @@ AC_SUBST(GLIB_LIBS)
 glib_version="`$PKG_CONFIG --modversion glib-2.0`"
 
 # gmime, some late-2012 version
-PKG_CHECK_MODULES(GMIME,gmime-2.6 >= 2.6.12)
+PKG_CHECK_MODULES(GMIME,gmime-2.6 >= 2.6.7)
 AC_SUBST(GMIME_CFLAGS)
 AC_SUBST(GMIME_LIBS)
 gmime_version="`$PKG_CONFIG --modversion gmime-2.6`"
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -780,7 +780,10 @@ draft message."
 
   ;; add any other headers specified
   (when other-headers
-    (message-add-header other-headers))
+    (dolist (h other-headers other-headers)
+      (if (symbolp (car h)) (setcar h (symbol-name (car h))))
+      (message-add-header (concat (capitalize (car h)) ": " (cdr h) "\n"  ))
+      ))
 
   ;; yank message
   (if (bufferp yank-action)
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -184,6 +184,20 @@ place to do that."
   :type 'hook
   :group 'mu4e-compose)
 
+
+(defcustom mu4e-compose-post-hook nil
+  "Hook run *after* message composition buffer is set up.
+If the compose-type is either 'reply' or 'forward', the variable
+`mu4e-compose-parent-message' points to the message replied to /
+being forwarded / edited, and `mu4e-compose-type' contains the
+type of message to be composed.
+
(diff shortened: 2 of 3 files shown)


stig:
Hi @titaniumbones, I manually applied https://github.com/djcb/mu/pull/952/commits/d60f58e8887ecec1d6692c846b69b80e91ec6a66 to my setup (I don't use any hooks, so didn't bother applying that changeset) but I don't think it has made a difference. I'm still seeing the "Message already sent, send again?" question. I selected no, and went to the header part and hit `C-c C-c` and it sent without errors.

titaniumbones:
Maybe that change is irrelevant. I was thinking it might be.

However, I also rewrote John's htmlize-and-send function -- see my blog
here:

http://matt.hackinghistory.ca/2016/11/18/sending-html-mail-with-mu4e/

What are you using to send your mail? Can you try my code and see if it
helps?

stig:
> What are you using to send your mail? Can you try my code and 
> see if it helps?

My current setup is here:
https://github.com/stig/dot-files/blob/master/emacs.d/Email.org

I don't tend to actually send HTML mail, I just like to edit it in 
org mode
(for footnotes etc). Because of that, moving to the header section 
and
pressing C-c C-c works fine for me. Actually sending from the body 
works fine
too, I just get that annoying warning.

As I don't want to send HTML emails automatically, I don't think 
your code is
exactly what I need.

titaniumbones:
I think that you will need to make sure org-mu4e-org-compose-mode has been turned OFF before you send the message.  One option might be to rewrite the toggle function with an optional parameter that allows you to specify turning it off, and add that as a hook to the send function. Or something like that. Anyway, try turning the mode off manually before sending and see if it works?


Stig Brautaset writes:

>> What are you using to send your mail? Can you try my code and 
>> see if it helps?
>
> My current setup is here:
> https://github.com/stig/dot-files/blob/master/emacs.d/Email.org
>
> I don't tend to actually send HTML mail, I just like to edit it in 
> org mode
> (for footnotes etc). Because of that, moving to the header section 
> and
> pressing C-c C-c works fine for me. Actually sending from the body 
> works fine
> too, I just get that annoying warning.
>
> As I don't want to send HTML emails automatically, I don't think 
> your code is
> exactly what I need.


--

averter:
For some reason I cannot get attachments right with this solution. Is there any particular trick to it? When I do `M-x mml-attach` I get the typical mu4e code
````
<#part type="image/jpeg" filename="~/example.jpeg" disposition=attachment>
<#/part>
````

**Comment by Chris00** [comment]
I think [OrgMsg](https://github.com/jeremy-compostella/org-msg) is a good alternative to compose HTML emails.  I supports attachments.

**Comment by djcb** [comment]
mu4e/org/compose has been deprecated for many years, closing this PR... perhaps OrgMsg can help?

**Comment by Chris00** [comment]
> perhaps OrgMsg can help?

Absolutely (one may even write maths and have images generated automatically).  Have not yet found the time to write something about it in the manual though.

**Comment by peterhoeg** [comment]
orgmsg is for writing in org and sending html. orgmsg also only kicks in when replying to an existing html mail (or when creating a new email).

My request was with regards to writing in org and sending either markdown or plain text (or html but orgmsg would handle that).

**Comment by Chris00** [comment]
> My request was with regards to writing in org and sending either markdown or plain text (or html but orgmsg would handle that).

Sending in plain text is no problem.  Just adapt the following to your liking:
```elisp
  (setq org-msg-default-alternatives '((new		. (text html))
				       (reply-to-html	. (text html))
				       (reply-to-text	. (text))))
```
Sending in `org` format is also possible.  It may be possible to register the [markdown exporter](https://www.orgmode.org/manual/Markdown-Export.html) in `org-msg-alternative-exporters`.

PR #956, opened 2016-11-23
logical error of choose text or html part of mail fixed

panjie:
The logic to decide whether text or html part of email will be used is wrong so that text part will always be choosed, even if html is forced to be used(using `h` in message view).


djcb:
Hmm, this doesn't seem quite right... with html-only messages I get:
```
error in process filter: mu4e-view-message-text: Wrong type argument: char-or-string-p, nil
error in process filter: Wrong type argument: char-or-string-p, nil
```

djcb:
I've fixed it now in a slightly different way.

PR #957, opened 2016-11-26
typo

gambhiro:


djcb:
Ah, thanks -- but already fixed that one.

PR #960, opened 2016-11-30
mu4e: call enter-func after setting vars during context switch

stig:
With this you can change bookmarks during context switch and have the menu redrawn by making your enter-func contain a call to mu4e~main-view-real.


djcb:
Thanks, 

The enter-func should be before setting those variables though, since want to allow for changing things before that kicks in.

If there's some problem with the menu, I think we should handle that in a different way.
(and I recommend not to use any functions/variables that start with `mu4e~`, those are internal to mu4e and can change at any moment),

stig:
No problems; thanks for the comments. I've opened #963 with an alternative solution that should solve the issue. (I like that one better, to be honest!)

PR #961, opened 2016-11-30
mu4e: smarter initialization of mu4e-msg2pdf

mssola:
Before this commit the initialization of the mu4e-msg2pdf variable
assumed that the build directory was available. However, that's not the
case when installing mu from a distro package (in my case from OBS).

With this change, mu4e will look for the `msg2pdf` binary on the
`exec-path` first. If the binary exists, then this will be the initial
value. If the binary doesn't exist, then we fallback to the
`mu4e-builddir` approach.

Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>

### DIFF
--- a/mu4e/mu4e-actions.el
+++ b/mu4e/mu4e-actions.el
@@ -49,7 +49,11 @@ Works for headers view and message-view."
 
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-(defvar mu4e-msg2pdf (concat mu4e-builddir "/toys/msg2pdf/msg2pdf")
+(defvar mu4e-msg2pdf
+  (let ((exec (locate-file "msg2pdf" exec-path exec-suffixes)))
+    (if exec
+        exec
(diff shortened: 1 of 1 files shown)


djcb:
`msg2pdf` is explicitly marked as `noinst_`, so it shouldn't be installed at all, I think.

mssola:
As far as I can see, it's not marked as `noinst`. Only test executables and test libraries are marked like that. If you think that it shouldn't be installed at all, maybe I can change this PR to make it so.

djcb:
https://github.com/djcb/mu/blob/master/toys/msg2pdf/Makefile.am#L31 says `noinst_`, or?

Perhaps it should be installed though.

mssola:
> https://github.com/djcb/mu/blob/master/toys/msg2pdf/Makefile.am#L31 says noinst_, or?

Sigh... I was looking at the wrong `Makefile.am` ... 

> Perhaps it should be installed though.

Makes sense to me. I can update this PR so it's installed.

djcb:
I made change (similar to your change so that it will look outside the build-path... but I don't think msg2pdf is ready yet to be installed. Anyway, thanks!

mssola:
Oh, I can see this in commit 5a0973304853eb59e70c4744f0b3d885ace0dd4a. Thanks!

PR #962, opened 2016-11-30
force use of strings for other-header in mu4e~compose-mail

titaniumbones:
org-mime and potentially other users of `compose-mail' pass a parameter
`other-headers', an alist of the form ((symbol . string)).
`mu4e~compose-mail' sends this parameter to message-add-header, which
appears to expect a string instead.  We therefore loop through
other-headers and instead pass a string to message-add-header for each
element of other-headers.


djcb:
Looks good! Thanks, merged.

PR #963, opened 2016-12-03
mu4e: redraw menu after context switch

stig:
This is useful if you change e.g. bookmarks in the context vars and want your menu to reflect that.

Would this be an acceptable alternative to https://github.com/djcb/mu/pull/960?


djcb:
Looks good - merged. Thanks!

Issue #964, opened 2016-12-03
mu: no matches for search expression (4) -- for size search with leading digit of lower bound bigger than leading digit of upper bound


stig:
## Expected or desired behavior

Results of `mu find size:3M..100M` should be a strict superset of `mu find size:3M..50M`.

## Actual behavior

The bigger range, which *should* be a superset of the smaller, actually returns 0 messages and prints an error where the smaller range finds 96 messages.

```
22:36:06 ~> mu find size:3M..50M | wc -l
      96
22:36:11 ~> mu find size:3M..100M | wc -l
mu: no matches for search expression (4)
       0
```

Bizarrely, the behaviour seems to be related to the leading digit of the lower bound. 0 and 1 is OK:

```
22:47:53 ~> mu find size:0..100M | wc -l
   16419
22:47:57 ~> mu find size:1..100M | wc -l
   16419
```

But 2-9 is not:

```
22:48:01 ~> mu find size:2..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:29 ~> mu find size:3..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:29 ~> mu find size:4..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:39 ~> mu find size:5..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:43 ~> mu find size:6..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:51 ~> mu find size:7..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:55 ~> mu find size:8..100M | wc -l
mu: no matches for search expression (4)
       0
22:51:59 ~> mu find size:9..100M | wc -l
mu: no matches for search expression (4)
       0
```

However, 10 is fine again:

```
22:52:03 ~> mu find size:10..100M | wc -l
   16419
```

As is using 50M as the upper bound:

```
22:48:06 ~> mu find size:2..50M | wc -l
   16419
```

## Steps to reproduce

Hopefully the above transcripts should be enough. I'm happy to provide more detail if required.

## Versions of mu, mu4e/emacs, operating system etc.

- macOS Sierra 10.12.1
- mu built from git master today (3rd December)

## Any other detail

I thought perhaps my index was corrupted and re-created it with `mu index --rebuild` before capturing the above sessions.

stig:
I have a bit more info. It's not the absolute size of the upper bound, it seems:

```
> mu find size:1K..10K | wc -l
   10327
[...]

PR #965, opened 2016-12-03
mu: swap begin and end if begin is _numerically_ bigger than end

stig:
The old behaviour would compare the strings, so would swap the start and end ranges if start was 3K and end was 2M.

Fixes: #964


djcb:
Oh, good catch! Thanks, merged.

PR #969, opened 2016-12-08
Child processes should be session leaders.

xuhdev:
If not, when the session of mu is killed, these child processes are also killed. This scenario shows up when using mu4e: a PDF attachment, for example, is opened by Evince, but as soon as Emacs exits, Evince is also killed.


djcb:
This h as been merged -- thanks!

PR #972, opened 2016-12-11
mu4e: add mu4e-msg-changed-hook to mu4e-index-updated-hook globally

j-hao:
`mu4e-msg-changed-hook` should add to `mu4e-index-updated-hook` globally instead of header mode only. Otherwise, it causes issue like https://github.com/agpchil/mu4e-maildirs-extension/issues/34


djcb:
Makes sense -- merged, thanks!

PR #976, opened 2016-12-13
Allow `mu4e-html2text-command' to be given `msg' as argument

Chris00:
As explained in `mu4e.texi`, this is the only way to change the "cid:" links to links to actual images.  The change is backward compatible.


Chris00:
The build is failing but this is [unrelated to my change](https://travis-ci.org/djcb/mu/jobs/183683390#L310).

Chris00:
See https://github.com/djcb/mu/pull/977 to make it pass the tests.

Chris00:
I may propose an alternative where attachments are always rewritten — so there should not be any need for `msg` as an argument...

Chris00:
The alternative is provided by PR https://github.com/djcb/mu/pull/978

Chris00:
I think this PR is more versatile than https://github.com/djcb/mu/pull/978 and should be preferred to it.  It allows to extract cid images through `shr-content-function` (I have additional code using this — at the moment, it goes through a file but it would be nice to get the image data directly into a buffer if `mu` allows it).

mu4e/mu4e-vars.el
```diff
@@ -242,13 +242,21 @@ and `mu4e-headers-visible-columns'."
...
+view buffer.  If `mu4e-view-show-cid-inline' is t, only images not
+appearing as cid:// links are displayed."
   :type 'boolean
   :group 'mu4e-view)
 
 (make-obsolete-variable 'mu4e-show-images
   'mu4e-view-show-images "0.9.9.x")
 
+(defcustom mu4e-view-show-cid-inline nil
+  "Whether to to embed cid:// images in the message view buffer (and
```
stig: "Whether to to" in docstring is one to too many!

Chris00: Thanks.

djcb:
So, there's a bit of movement here; mu4e-message now _has_ a mu4e-html2text-command that can take `msg` as a parameter (and uses introspection for backward compatibility).

Apart from that, I guess we don't need a special `mu4e-view-show-cid-inline` and use `mu4e-view-show-images` instead. Can you update your PR for that? Thanks.

Chris00:
So one assumes that, if `mu4e-view-show-images` is true, then `mu4e-html2text-command` displayed and we do not need to show them again at the end of the message?

djcb:
Oh, that's a good point, I think that for cid:-images, indeed they shouldn't be shown (again) at the end of the message. Normal attached images should though, I think, or?

Chris00:
On 2017-01-25, Dirk-Jan C. Binnema wrote:
>
> Normal attached images should though, I think, or?

That’s what I think.

djcb:
This PR is superseded; closing.

PR #977, opened 2016-12-14
Travis: use recent autoconf 2.69 and recent autoconf-archive

Chris00:


Chris00:
P.S. Note that the `-bin` are [supposed to be deprecated](https://github.com/rejeep/evm#deprecation-warning).

djcb:
Thanks!

djcb:
BTW, Travis broke since mu now requires c++14...  @Chris00: do you know how to fix this?

Chris00:
Does it help if you add the package that provides c++14 to `BUILD_PKGS` (in https://github.com/djcb/mu/blob/master/.travis.yml)?

PR #978, opened 2016-12-14
Convert cid: → file:// files before mu4e-html2text-command

Chris00:
The cid: attachments are saved to disk and the file:// links point to
these files.


Chris00:
Of course Travis fails — see https://github.com/djcb/mu/pull/977

Chris00:
I was favoring this PR initially but it did not play well with `shr` — images did not show up.  I have code that shows images in `shr` using https://github.com/djcb/mu/pull/976, thus the latter is favored.

PR #979, opened 2016-12-14
Defer compilation instructions to mu4e manual

jschwab:
The compilation information appeared in two places.  It will be easier
to keep up to date if it is only in one place.


djcb:
I'm not going to merge the two documents, they have a different purpose / target audience, but indeed, HACKING needed an update. Thanks!

PR #983, opened 2016-12-18
Ignore xua

stig:


stig:
There was a whole lot of extra commits on this PR for some reason... I've fixed it now, so it only contains the .gitignore addition.

djcb:
Merged, thanks.

PR #985, opened 2016-12-19
Add some improvements in attachment actions.

thierryvolpiatto:
* mu4e/mu4e-view.el (mu4e-view-attachment-action):
Handle multi save in mu4e-view-attachment-action.

* mu4e/mu4e-view.el (mu4e-view-attachment-actions):
Make mu4e-view-attachment-actions a defcustom and add some options.

New user variable mu4e-view-attachment-assoc.

* mu4e/mu4e-view.el (mu4e-view-attachment-assoc): New.
(mu4e-view-open-attachment-with): Use it.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -111,8 +111,17 @@ where:
 
 The first letter of NAME is used as a shortcut character.")
 
-(defvar mu4e-view-attachment-actions
-  '( ("wopen-with" . mu4e-view-open-attachment-with)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
`mu4e-view-attachment-assoc` allow configuring which program to use by default with extension.
I added it as initial input in the `read-string` but could be added instead to the default arg and accessible with `M-n` (in this case more than one program can be added).

djcb:
Looks good, merged, thanks!

Issue #808, opened 2016-02-23
Support for long lines for incoming emails


wavexx:
Incoming messages using quoted-printable or format=flowed encoding can distinguish between hard or soft newlines. I guess quoted-printable is currently handled by gmime, which just unwraps the message.

However, if a message is text/plain, it can still be format=flowed (it's in the content-type). In substance, f=f can distinguish between a hard and a soft newline as well, but just using trailing whitespace in a way that's completely invisible to the end user. A f=f message looks wrapped and can be replied as-is (preserving it's structure), or can be logically unwrapped without breaking quoting.

Even if the RFC discourages it, a f=f message transmitted in QP still works correctly, for example.

The decoding algorithm is trivial, and it could be implemented completely in mu4e-view if the content-type of the text part was available.

andersjohansson:
:+1:  With the new options for sending format=flowed, it seems silly to not be able to display it (as flowed)

djcb:
Okay, I think this works now with the new (experimental) view mode -- `(setq mu4e-view-use-gnus t)` (before starting mu4e). Finally!

PR #986, opened 2016-12-19
POC: Respect format=flowed and delsp=yes for viewing plain-text messages

stig:
This is my initial attempt at addressing #808, supporting `format=flowed` (and `delsp=yes`) when viewing mail.

While this does rewrap some test messages I've tried nicely, I wouldn't recommend merging it in its current form:

- ~There are almost certainly may be memory leaks here.~ Update: may not be true any more, after the last few commits on this branch.
- ~I also encountered a situation where some mails didn't load.~ Update: I was able to track down why, and plugged that particular hole. It's running fine for me now.
- I suspect we would also want a toggle to say whether to honour reflow, much as we do for sending.

I fear this approach is too simplistic because of the way mu accumulates plain-text bodies. This approach could produce very nasty results if some plain-text bodies are marked `format=flowed` and others are not, but then everything is format-flowed. This body accumulation happens in C, but I don't relish the idea of re-implementing `flow-fill.el` in C so we can choose to reflow individual body chunks while accumulating. I feel a better solution would be to delay the accumulation and, instead of providing separate html and text strings in mu-msg-sexp.c, just provide an array of bodies, each with their own content type, and handle the accumulation in emacs-lisp.

### DIFF
--- a/lib/mu-msg-sexp.c
+++ b/lib/mu-msg-sexp.c
@@ -423,10 +423,30 @@ append_sexp_thread_info (GString *gstr, const MuMsgIterThreadInfo *ti)
 }
 
 static void
+append_sexp_param (GString *gstr, GSList *param)
+{
+	for (;param; param = g_slist_next (param)) {
+		char *key, *value;
+
+		key = param->data;
+		key = key ? mu_str_escape_c_literal (key, FALSE) : "";
+
+		param = g_slist_next (param);
+		value = param->data;
+		value = value ? mu_str_escape_c_literal (value, FALSE) : "";
+
+		g_string_append_printf (gstr, "(\"%s\" . \"%s\")", key, value);
+		if (param->next)
+			g_string_append_c (gstr, ' ');
+	}
+}
+
+static void
 append_message_file_parts (GString *gstr, MuMsg *msg, MuMsgOptions opts)
 {
 	const char	*str;
 	GError		*err;
+	const GSList    *params;
 
 	err = NULL;
 
@@ -444,6 +464,13 @@ append_message_file_parts (GString *gstr, MuMsg *msg, MuMsgOptions opts)
 	str = mu_msg_get_header (msg, "User-Agent");
 	if (str || (str = mu_msg_get_header (msg, "X-Mailer")))
 		append_sexp_attr (gstr, "user-agent", str);
+
+        params = mu_msg_get_body_text_content_type_parameters (msg, opts);
+        if (params) {
+		g_string_append_printf (gstr, "\t:body-txt-params (");
+		append_sexp_param (gstr, params);
+		g_string_append_printf (gstr, ")\n");
+	}
 
 	append_sexp_body_attr (gstr, "body-txt",
 			  mu_msg_get_body_text(msg, opts));
--- a/lib/mu-msg.c
+++ b/lib/mu-msg.c
@@ -520,6 +520,76 @@ get_body (MuMsg *self, MuMsgOptions opts, gboolean want_html)
 		return g_string_free (bdata.gstr, FALSE);
 }
 
+
+struct _ContentTypeData {
+	const GMimeContentType *ctype;
+	gboolean want_html;
+};
+typedef struct _ContentTypeData ContentTypeData;
+
+
+static void
+find_content_type (MuMsg *msg, MuMsgPart *mpart, ContentTypeData *cdata)
+{
+	if (!GMIME_IS_PART(mpart->data))
+		return;
+
+	/* text-like attachments are included when in text-mode */
+
+	GMimePart *wanted = NULL;
+
+	if (!cdata->want_html &&
+	    (mpart->part_type & MU_MSG_PART_TYPE_TEXT_PLAIN))
+		wanted = mpart->data;
+	else if (!(mpart->part_type & MU_MSG_PART_TYPE_ATTACHMENT) &&
+		 cdata->want_html &&
+		 (mpart->part_type & MU_MSG_PART_TYPE_TEXT_HTML))
+		wanted = mpart->data;
+	if (wanted)
+		cdata->ctype = g_mime_object_get_content_type (wanted);
+}
+
+
+static const GSList*
+get_content_type_parameters (MuMsg *self, MuMsgOptions opts, gboolean want_html)
+{
+	ContentTypeData cdata;
+	cdata.want_html = want_html;
+	cdata.ctype = NULL;
+
+	mu_msg_part_foreach (self, opts,
+			     (MuMsgPartForeachFunc)find_content_type,
+			     &cdata);
+
+	if (cdata.ctype) {
+		const GSList *paramlist;
+		const GMimeParam *param;
+
+		paramlist = NULL;
+		param = g_mime_content_type_get_params (cdata.ctype);
+
+		for (; param; param = param->next) {
+			paramlist = g_slist_prepend (paramlist,
+						     g_strdup (param->name));
+
+			paramlist = g_slist_prepend (paramlist,
+						     g_strdup (param->value));
+		}
+
+		return free_later_lst(self, g_slist_reverse (paramlist));
+	}
+	return NULL;
+}
+
+
+const GSList*
+mu_msg_get_body_text_content_type_parameters (MuMsg *self, MuMsgOptions opts)
+{
+	g_return_val_if_fail (self, NULL);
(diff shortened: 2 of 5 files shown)


Chris00:
I did not really dig into your code but it looks to me like a display issue so I am surprised there is quite a bit of C code involved.  Did you have a look to the recommendations for [Gnus](https://www.emacswiki.org/emacs/GnusFormatFlowed)?  See also [Emacs MIME](https://www.gnu.org/software/emacs/manual/emacs-mime.html#Flowed-text-1).

stig:
Hi @Chris00 there's quite a bit of C code because mu doesn't currently (as far as I could see) expose any of the content-type params to mu4e; it just gives the text and html bodies without any parameters. As I understand it, without the `format=flowed` and `delsp=yes` params mu4e can't determine whether to reflow a message.

I don't quite understand the relevance of the Gnus resource you linked to. It seems to be talking about long lines, which is not what I want. I am aiming for body text to be reflowed to my desired fill column, not wrapping at the frame width.

Chris00:
On 2016-12-27, Stig Brautaset wrote:
>
> Hi @Chris00 there's quite a bit of C code because mu doesn't currently 
> (as far as I could see) expose any of the content-type params to mu4e; 
> [...]

OK.

> I don't quite understand the relevance of the Gnus resource you linked 
> to. It seems to be talking about long lines, which is not what I 
> want. I am aiming for body text to be reflowed to my desired fill 
> column, not wrapping at the frame width.

I read more your code and I see you use Gnus flow-fill which is good — 
so the link does not add anything useful.

lib/mu-msg-sexp.c
```diff
@@ -445,6 +446,25 @@ append_message_file_parts (GString *gstr, MuMsg *msg, MuMsgOptions opts)
 	if (str || (str = mu_msg_get_header (msg, "X-Mailer")))
 		append_sexp_attr (gstr, "user-agent", str);
 
+        params = mu_msg_get_body_text_content_type_parameters(msg, opts);
+        if (params) {
+		GString *paramstr = g_string_new("");
```
djcb: Please split declaration + initialization into separate lines. Also, use tabs for indentation, spaces for alignment.

lib/mu-msg-sexp.c
```diff
@@ -445,6 +446,25 @@ append_message_file_parts (GString *gstr, MuMsg *msg, MuMsgOptions opts)
...
+			param = params->data;
+			g_string_append_printf (paramstr, ":%s \"%s\"", 
+						(char*)param->data, 
+						(char*)g_slist_next(param)->data);
+
+			if (params->next)
+				g_string_append_c(paramstr, ' ');
+		}
+
+		g_string_append_printf(gstr, "\t:body-txt-params (%s)\n", paramstr->str);
```
djcb: Lines should not go over the 80-char limit

lib/mu-msg-sexp.c
```diff
@@ -445,6 +446,25 @@ append_message_file_parts (GString *gstr, MuMsg *msg, MuMsgOptions opts)
...
 
+        params = mu_msg_get_body_text_content_type_parameters(msg, opts);
+        if (params) {
+		GString *paramstr = g_string_new("");
+          
+		for (; params; params = g_slist_next(params)) {
+			GSList *param;
+			param = params->data;
+			g_string_append_printf (paramstr, ":%s \"%s\"", 
+						(char*)param->data,
```
djcb: Shouldn't param->data be escaped here?

lib/mu-msg.c
```diff
@@ -520,6 +520,76 @@ get_body (MuMsg *self, MuMsgOptions opts, gboolean want_html)
...
+		const GSList *paramlist;
+		const GMimeParam *params, *param;
+
+		paramlist = NULL;
+		params = g_mime_content_type_get_params(cdata.ctype);
+
+		for (param = params; param; param = params->next) {
+			const GSList *item;
+
+			item = g_slist_prepend(NULL, param->value);
```
djcb: g_slist_prepend <SPACE> (

lib/mu-msg.c
```diff
@@ -520,6 +520,76 @@ get_body (MuMsg *self, MuMsgOptions opts, gboolean want_html)
...
+	else if (!(mpart->part_type & MU_MSG_PART_TYPE_ATTACHMENT) &&
+		 cdata->want_html &&
+		 (mpart->part_type & MU_MSG_PART_TYPE_TEXT_HTML))
+		wanted = mpart->data;
+	if (wanted)
+		cdata->ctype = g_mime_object_get_content_type(wanted);
+}
+
+
+const GSList*
```
djcb: function should be static

lib/mu-msg.c
```diff
@@ -520,6 +520,76 @@ get_body (MuMsg *self, MuMsgOptions opts, gboolean want_html)
...
+		}
+
+		return g_slist_reverse(paramlist);
+	}
+	return NULL;
+}
+
+
+const GSList*
+mu_msg_get_body_text_content_type_parameters(MuMsg *self, MuMsgOptions opts)
```
djcb: space between function name and params

mu4e/mu4e-message.el
```diff
@@ -181,6 +182,17 @@ representation."
       ;; otherwise, use text
       (t nil))))
 
+(defun mu4e~want-format-flowed (msg)
```
djcb: Hmm, I think `mu4e~message-body-has-content-type-param` would be better, with parameters `msg' and some string; furthermore, seems that it's a property of a specific mime-part, not necessarily of the message as a whole, or?

stig: > furthermore, seems that it's a property of a specific mime-part, not necessarily of the message as a whole, or?

Yeah, as I mentioned in the PR description, I fear this approach is too simplistic because of the way mu accumulates plain-text bodies. It could produce very nasty results if some plain-text bodies are marked format=flowed and others are not, but then everything is format-flowed. Hence I submitted this as a starting point for further discussion rather than expecting it to be merged in its current form.

lib/mu-msg.c
```diff
@@ -520,6 +520,76 @@ get_body (MuMsg *self, MuMsgOptions opts, gboolean want_html)
...
+
+		for (; param; param = param->next) {
+			const GSList *item;
+
+			item = g_slist_prepend (NULL, param->value);
+			item = g_slist_prepend (item, param->name);
+			paramlist = g_slist_prepend (paramlist, item);
+		}
+
+		return g_slist_reverse (paramlist);
```
stig: I think this is a memory leak. I tried using `free_later_lst`, but this made the tests crash. I *think* this is because `mu_msg_destroy` assumes that all entries in `_free_later_lst` are lists of strings rather than lists of tuples (lists). Do you think it's best to flatten the list, or add a new `_free_later_tuples` member to `MuMsg` struct?

mu4e/mu4e-message.el
```diff
@@ -167,20 +168,26 @@ This is equivalent to:
 on PREFER-HTML and whether the message supports the given
 representation."
   (let* ((txt (mu4e-message-field msg :body-txt))
-	  (html (mu4e-message-field msg :body-html))
```
stig: Sorry about this. Any time I tried to edit any of the code near this function Emacs wanted to "fix" the indentation here... If the *new* indentation is right, I'm happy to submit it as a separate PR. If it is not, are you able to share what kind of settings you use to make Emacs indent it that way?

lib/mu-msg.c
stig: It was definitively a memory leak, but I kinda answered my own question with that comment. I flattened the list, and added free statements.

stig:
Thanks for all the comments @djcb! I believe I have addressed all of them. Where do I take this from here?

djcb:
Ah, sorry for the late response. I like the code and you even added tests, very good! 

I would like to merge it sometime soon, but one thing I'd like to do before that is add some generic transform function hook,  and your flow-handling would could use that. So, give me a few days to add that; after that, I think we can get your change in (without need to further change `mu4e-message-body-text`. Cheers!

djcb:
There's `mu4e-message-body-rewrite-functions` now (see `mu4e-message-outlook-cleanup` for an example); so if you can update your code for that, we can take it in, thanks!

stig:
@djcb Thanks, I've rebased onto master and combined all the commits into one. Let me know if there's anything else you want from this!

djcb:
Merged -- thanks!

PR #991, opened 2016-12-23
mu4e: add mu4e-update-timeout

j-hao:
I think this is useful if there was any issue and update process was stuck forever. I have personally encountered the same issue with isync and exchange 2016 just like https://sourceforge.net/p/isync/bugs/25/ and this allow me to workaround it.

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -1038,7 +1038,14 @@ run in the background; otherwise, pop up a window."
     ;; if we're running in the foreground, handle password requests
     (unless run-in-background
       (process-put proc 'x-interactive (not run-in-background))
-      (set-process-filter proc 'mu4e~get-mail-process-filter))))
+      (set-process-filter proc 'mu4e~get-mail-process-filter))
+    (if mu4e-update-timeout
+        (lexical-let ((proc proc))
+          (run-at-time mu4e-update-timeout nil
+                       (lambda ()
+                         (let ((status (process-status proc)))
+                           (if (eq status 'run)
+                               (kill-process proc)))))))))
 
 (defun mu4e-update-mail-and-index (run-in-background)
   "Get a new mail by running `mu4e-get-mail-command'. If
--- a/mu4e/mu4e-vars.el
+++ b/mu4e/mu4e-vars.el
@@ -117,6 +117,13 @@ mu4e."
   :group 'mu4e
   :safe 'integerp)
 
+(defcustom mu4e-update-timeout nil
+  "Number of seconds for update operation to timeout. If nil, no timeout."
+  :type '(choice (const :tag "No timeout" nil)
+                 (integer :tag "Seconds"))
+  :group 'mu4e
+  :safe 'integerp)
+
 (defvar mu4e-update-pre-hook nil
   "Hook run just *before* the mail-retrieval / database updating process starts.
  You can use this hook for example to `mu4e-get-mail-command' with


thierryvolpiatto:
Jun Hao <notifications@github.com> writes:

> I think this is useful if there was any issue and update process was
> stuck forever. I have personally encountered the same issue with isync
> and exchange 2016 just like https://sourceforge.net/p/isync/bugs/25/
> and this allow me to workaround it.

Perhaps it would be better to kill process instead of interrupt in
`mu4e-interrupt-update-mail` ?

Most of the time when hitting "S" interrupt-process is not enough to
stop the process.

Sometimes the update process is just very slow and not stuck forever, a
timeout would stop it instead of letting it finishing...

--
Thierry

j-hao:
This PR is actually will kill process (if process is in running state) instead of interrupt it. So I think it matches your use case as well.

Also thanks for the info about `S` key. Don't know that :)

thierryvolpiatto:
Jun Hao <notifications@github.com> writes:

> This PR is actually will kill process (if process is in running state)
> instead of interrupt it. So I think it matches your use case as well.

Yes, sending SIGKILL is better indeed, but using a timeout is not the
good approach IMO, it will stop the update on slow connections even when
no particular problem happens during this connection.

> Also thanks for the info about S key. Don't know that :)

Is using the "S" key with SIGKILL instead of SIGINT would fit your needs
?

I have a PR ready for this that allow configuring which signal to send
to update process to stop it among other things.

--
Thierry

j-hao:
Thierry Volpiatto <notifications@github.com> writes:

> Jun Hao <notifications@github.com> writes:
>
>> This PR is actually will kill process (if process is in running state)
>> instead of interrupt it. So I think it matches your use case as well.
>
> Yes, sending SIGKILL is better indeed, but using a timeout is not the
> good approach IMO, it will stop the update on slow connections even when
> no particular problem happens during this connection.
>

I don't think we can tell is either a slow connection or an actual
issue. Any idea/suggestion?

>> Also thanks for the info about S key. Don't know that :)
>
> Is using the "S" key with SIGKILL instead of SIGINT would fit your needs
> ?
>
> I have a PR ready for this that allow configuring which signal to send
> to update process to stop it among other things.

Actually, "S" with SIGINT solves my issue and I just don't want to
manually run it every couple minutes. The connection appears to be
stuck forever. PS, by default, timeout won't run. I think these two
configuration doesn't conflict and can coexist.

-- 
Thanks - Jun

thierryvolpiatto:
Jun Hao <notifications@github.com> writes:

> Thierry Volpiatto <notifications@github.com> writes:
>
>> Jun Hao <notifications@github.com> writes:
>>
>>> This PR is actually will kill process (if process is in running state)
>>> instead of interrupt it. So I think it matches your use case as well.
>>
>> Yes, sending SIGKILL is better indeed, but using a timeout is not the
>> good approach IMO, it will stop the update on slow connections even when
>> no particular problem happens during this connection.
>>
>
> I don't think we can tell is either a slow connection or an actual
> issue. Any idea/suggestion?

Logging the process output (done in the ready PR I have here).
https://github.com/thierryvolpiatto/mu/tree/update_process

>>> Also thanks for the info about S key. Don't know that :)
>>
>> Is using the "S" key with SIGKILL instead of SIGINT would fit your needs
>> ?
>>
>> I have a PR ready for this that allow configuring which signal to send
>> to update process to stop it among other things.
>
> Actually, "S" with SIGINT solves my issue and I just don't want to
> manually run it every couple minutes. The connection appears to be
> stuck forever. PS, by default, timeout won't run. I think these two
> configuration doesn't conflict and can coexist.

Yes sure, I have no problems with the timeout, it will work fine for me
99% of the time even when enabled, just wanted to point out that a valid
connection but slow can be aborted for no reason by the timeout.

BTW note that you can use with-timeout to implement timeout here:

    (with-timeout (mu4e-update-timeout
                   (let ((status (process-status proc)))
                     (when (eq status 'run)
                       (kill-process proc))))
      [...])

Something like this.

-- 
Thierry

djcb:
So, closing this. Thanks for the effort though!

Issue #997, opened 2016-12-30
Make electric-quote-local-mode work


Chris00:
At the moment, `electric-quote-local-mode` does not work (outside "comments" seen as lines starting with `>` or `|`).  This is due to `comment-use-syntax` left to `undecided`.  As the documentation for this variable says *Major modes should set this variable*, I believe `mu4e-compose` should set to to `nil` (so `electric-quote-local-mode` works anywhere).  If there is an agreement, I can submit a PR.

djcb:
It would be useful, but setting it to `nil` would not be enough I think; quote local mode should be neutralized for the message-headers, as well as in any of the special parts like attachment-blobs.

Can you make a PR for that?

Chris00:
Why disable it in the headers?  `"` is not affected, only `` ` `` and `'` and these may be useful for the `Subject:`.

Chris00:
Also, electric quote mode works by adding a function to `'post-self-insert-hook` so it only changes quotes that one types, not those present in attached files, pasted text,...

djcb:
I wouldn't want it in headers, not convinced non-ascii will work there in all cases (and think of "Resend"). It certainly won't work in the mime-meta blocks. If not too hard, we probably also want to exclude it from code blocks, ie. 
```
--8<---------------cut here---------------start------------->8---
   code
--8<---------------cut here---------------end--------------->8---
```
Can you add that? I think all it involves setting a few of the `comment-` regexps.

Apart from that, looks useful, thanks!

Chris00:
I checked the code of `electric-quote-post-self-insert-function` and when `comment-use-syntax` is `nil`, it basically only cares about `` ` `` or `'` being pressed.  It it is `t` (and because `comment-start` is defined), it tries to see whether the point is in a comment (quoted text for emails) or a string to enable the conversion.  I checked org-mode—which sets `comment-use-syntax` to `nil`—but it does not "protect" quotes for code blocks (one has to type `C-q '` to have `'` instead of `’`—or (temporarily) disable electric quote.

Note that `electric-quote-local-mode` does not affect `"`.

PR #1003, opened 2017-01-02
Make electric-quote-(local-)mode work when composing emails

Chris00:
electric-quote-(local-)mode used to have no effect because
`comment-use-syntax' was left to `undecided' so the electric mode was
not enabled for paragraphs.

Fixes https://github.com/djcb/mu/issues/997

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -407,6 +407,8 @@ buffers; lets remap its faces so it uses the ones for mu4e."
     ;; set this to allow mu4e to work when gnus-agent is unplugged in gnus
     (set (make-local-variable 'message-send-mail-real-function) nil)
     (make-local-variable 'message-default-charset)
+    ;; Set to nil to enable `electric-quote-local-mode' to work:
+    (set (make-variable-buffer-local 'comment-use-syntax) nil)
     ;; message-mode has font-locking, but uses its own faces. Let's
     ;; use the mu4e-specific ones instead
     (mu4e~compose-remap-faces)
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -437,16 +437,27 @@ buffers; lets remap its faces so it uses the ones for mu4e."
 	'(left-curly-arrow right-curly-arrow))
       (visual-line-mode t))
 
-    (when (lookup-key message-mode-map [menu-bar text])
-      (define-key-after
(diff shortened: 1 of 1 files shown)


Chris00:
Since it is not possible to turn it off for certain portions of the text, I added a menu entry to toggle it—so one remembers how to do it when needed.

djcb:
Thanks. Are you planning on updating this so headers / mml parts (and, i guess, code-blocks) are not affected by the electric-quote mode? Also note, electric-quote-mode only appeared with (iirc) emacs 25.1, so some fboundb check is needed.

Chris00:
I do not think that it needs additional checks because `:enable (functionp 'electric-quote-local-mode)` will only enable it if `electric-quote-local-mode` is defined.  Or maybe you would like that the menu entry is not visible at all then `electric-quote-local-mode` is not defined?

Chris00:
About headers, this is the same mode so I do not see how to turn electric quote selectively for them.

For mml parts, I am not sure what you want.  Since electric quote only changes quotes that one __types__, it will not affect mml parts.  May you be more precise?

djcb:
Well, it's a corner-case, but I sometimes edit the mime-blobs (say for the mime type or the disposition), and wouldn't want to see fancy headers there. Same for the headers.

Now, the question is if that can be prevented... if `comment-start` would allow for a regexp, we could simply match the unwanted areas, and only enable it for paragraphs. But I see now that it's not easily possible; also the other items (strings and paragraps) are not easy to manipulate either  :-/ 

So, sadly not much we can do about that. I'll take your change then, if you add a comment for  the setting of `comment-use-syntax` (i.e, why are we setting it), and not show the menu item if emacs is too old.

Chris00:
If I got you right, for mime-blobs, you should not have too much trouble as electric quote does not change the behavior of `"`.

I also opened a [bug against Emacs message mode](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=25333).  Let's see if they think electric quote needs to be improved.  

I guess another way is to have several modes in the buffer and activate electric quote only for the mode active when typing text.

djcb:
Ah, so, I meant that we can't really do anything more fancy than in your patch; anyway, I can merge it if you  can add a comment for the setting of `comment-use-syntax` (i.e, why are we setting it), and not show the menu item if emacs is too old. Thanks.

Chris00:
Done.

djcb:
Merged -- thanks!

PR #1004, opened 2017-01-03
Replace "focus" with "context"

jschwab:
There were a handful places, primarily in the main menu, where the
word "focus" seemed to be used as a synonym for "context".  This
changes all of those instances to "context".


djcb:
This has been merged, thanks!

PR #1008, opened 2017-01-06
mu4e: Add possibility to forward messages as attachments

KrzysiekJ:
This commit adds a global variable `mu4e-compose-forward-as-attachment`. To enable choosing forwarding method on a per-message basis would probably require either:

* changing the mu server backend so that it distinguishes between inline forwarding and forwarding as attachment;
* changing the mu server backend so that it doesn’t return attachments at all and making both inline and as attachment forwarding via MIME (and also making mu4e actually display MIME-enclosed inline emails).

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -190,20 +190,25 @@ place to do that."
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 
+(defun mu4e-compose-attach-message (msg)
+  "Insert message MSG as an attachment."
+  (let ((path (plist-get msg :path)))
+    (unless (file-exists-p path)
+      (mu4e-warn "Message file not found"))
+    (mml-attach-file
+      path
+      "message/rfc822"
+      (or (plist-get msg :subject) "No subject")
+      "attachment")))
+
 (defun mu4e-compose-attach-captured-message ()
   "Insert the last captured message file as an attachment.
 Messages are captured with `mu4e-action-capture-message'."
   (interactive)
   (unless mu4e-captured-message
     (mu4e-warn "No message has been captured"))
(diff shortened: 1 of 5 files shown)


mu4e/mu4e-compose.el
```diff
@@ -190,20 +190,25 @@ place to do that."
...
-    (unless (file-exists-p path)
-      (mu4e-warn "Captured message file not found"))
-    (mml-attach-file
-      path
-      "message/rfc822"
-      (or (plist-get mu4e-captured-message :subject) "No subject")
-      "attachment")))
+  (mu4e-compose-attach-message mu4e-captured-message))
+
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
djcb: Ah, the above is a nice re-factoring

mu4e/mu4e-compose.el
```diff
@@ -514,7 +519,7 @@ automatically encrypt that reply."
 	  (encrypt (mml-secure-message-encrypt))
 	  (sign-and-encrypt (mml-secure-message-sign-encrypt)))))
 
-(defun* mu4e~compose-handler (compose-type &optional original-msg includes)
+(defun* mu4e~compose-handler (compose-type &optional original-msg includes as-attachment)
   "Create a new draft message, or open an existing one.
```
djcb: In this one (and the other), don't pass 'as-attachment' as a parameter; rather, use the value of mu4e-compose-forward-as-attachment directly in the function, so only `mu4e~compose-handler` and `mu4e~draft-forward-construct` need changes.

djcb:
Ah, thanks, looks interesting. See my comments in the code.

KrzysiekJ:
I’ve updated the commit to use global variables (previous version was deliberately not doing it to make the code more pure) and also removed adding two newlines to message body when forwarding as attachment.

djcb:
Merged, thanks!

PR #1031, opened 2017-02-17
Add support for specifying --muhome via MU_HOME in the environment

avar:
When you have multiple mu home directories, e.g. for the use case
detailed in my "Changing mu4e-{maildir,mu-home} from a context hook"
post to the mailing list it's quite inconvenient to have to hammer out
"mu --muhome=.. find .." every time you want to run some ad-hoc
command.

This allows me to set up a screen session where I do searches in mu
directory A in some screen panes, and searches in directory B in
others.

I initially called this MU_MUHOME but then I noticed that the perl
plugin has MUP_MU_HOME for analogous functionality, so I'm just
following its example.

The code I'm adding in mu-util.c is just a copy/paste & adjustment of
the same sort of already tested functionality in
mu_util_guess_maildir() just a few lines earlier.

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -16,6 +16,9 @@
    - prefer gpg2 for crypto
    - fix a crash when running on OpenBSD
    - fix --clear-links (broken filenames)
+   - You can now set the MU_HOME environment variable as an
+     alternative way of setting the mu homedir via the --muhome
+     commadn-line parameter.
     
 *** mu4e
 
--- a/lib/mu-util.c
+++ b/lib/mu-util.c
@@ -199,9 +199,16 @@ gchar*
 mu_util_guess_mu_homedir (void)
 {
 	const char* home;
+	const gchar *hdir1;
 
-	/* g_get_home_dir use /etc/passwd, not $HOME; this is better,
-	 * as HOME may be wrong when using 'sudo' etc.*/
+	/* first, try MU_HOME */
+	hdir1 = g_getenv ("MU_HOME");
+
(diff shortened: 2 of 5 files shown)


djcb:
Merged, thanks!

daphniz:
Has this been deleted?  I use

		alias mcfind='mu cfind --muhome="$MUHOME"'
		alias mextract='mu extract --muhome="$MUHOME"'
		alias mfind='mu find --muhome="$MUHOME"'
		alias mhelp='mu help --muhome="$MUHOME"'
		alias mindex='mu index --muhome="$MUHOME"'
		alias mmkdir='mu mkdir --muhome="$MUHOME"'
		alias mremove='mu remove --muhome="$MUHOME"'
		alias mscript='mu script --muhome="$MUHOME"'
		alias mserver='mu server --muhome="$MUHOME"'
		alias mview='mu view --muhome="$MUHOME"'
		
but this should be unnecessary.

djcb:
Yeah, mu now (>= 1.3.x) uses the XDG Base Directory Specification, but you can override it by manipulating `XDG_CACHE_HOME`,

djcb:
Ie. `alias mindex=XDG_CACHE_HOME=mypath mu index`

daphniz:
but isn't this going to affect other programmes?

daphniz:
ok, it won't in this case, but my point was that i don't want to have to create aliases.  it should be possible, i think, to set an env variable, and then use `mu index` as usually.

daphniz:
ie., there should be a `$MUHOME`, and this should have
precedence over `$XDG_CACHE_HOME`.

PR #1033, opened 2017-02-17
perl: Add a .gitignore file with the MYMETA build assets

avar:
These get created by ExtUtils::MakeMaker, so they won't be found by
grepping our own code, but they're created by the build process.


djcb:
Merged, thanks!

PR #1034, opened 2017-02-17
Add example of matching maildir and its sub-directories in contexts

jschwab:
Inspired by this mu-discuss thread:
https://groups.google.com/d/msg/mu-discuss/rAT2YJPEwsA/eYmoGJbUCgAJ


djcb:
Merged -- thanks!

Issue #915, opened 2016-08-30
Honour `message-dont-reply-to-names'


wavexx:
`message-mode` complains if some addresses match `message-bogus-addresses`. Normally it would also skip addresses matching `message-dont-reply-to-names` when composing a reply, which is great to skip most of the noreply@ addresses when composing a wide reply.

It would be nice to replicate this behavior.

wavexx:
I think we can close this now.

PR #1036, opened 2017-02-19
Allow to ignore addresses when doing wide replies

wavexx:
Introduce a new variable, mu4e-compose-reply-ignore-address, which matches addresses to be skipped when doing wide replies.

This is identical in behavior to messages-dont-reply-to-names from message.el (which we default on). See RFE #915.

In fact, since it supports the same values, I would be tempted to ``defalias`` instead of ``defcustom``, or use ``message-dont-reply-to-names`` directly. But I'm flexible here. An alternative would be to default to mu4e-compose-complete-ignore-address-regexp.

This finally fixes the ability to perform wide replies with noreply addresses. For instance, when replying to github issues with a "\\bnoreply\\n" regexp, mu4e will not even prompt for a wide reply anymore, since all Cc: addresses get removed (as they should).

Feedback appreciated.

### DIFF
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -191,6 +191,30 @@ the list form the original."
       reply-to)))
 
 
+(defun mu4e~strip-ignored-addresses (addrs)
+  "Return all the addresses in ADDRS not matching
+`mu4e-compose-reply-ignore-address'."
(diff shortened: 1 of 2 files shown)


djcb:
Looks useful, thanks! Merged.

PR #1037, opened 2017-02-20
Do not duplicate headers views in iconified/hidden frames

wavexx:
When the frame running mu4e is in another virtual desktop or iconified, but
still contains a visible headers buffer, do not attempt to create a new view.


djcb:
Merged, thanks. Still a bit unsure about the best way to handle multiple view/windows/frames though

wavexx:
On Tue, Feb 21 2017, Dirk-Jan C. Binnema wrote:
> Merged, thanks. Still a bit unsure about the best way to handle multiple
> view/windows/frames though

It's tricky. I also noticed there is some other strange interaction that
I couldn't trigger easily: sometimes the header buffer seem to be
populated from results of a previous query.

PR #1044, opened 2017-03-06
mu4e: update the org-contacts package link in the doc

garaud:
Hi,
The previous link didn't exist. 
So I updated it!

Cheers,
Damien


djcb:
Merged -- thanks!

PR #1066, opened 2017-04-12
s/defsubst/defun/ for all but the most trivial cases

avar:
Change various 'defsubst' to 'defun' for all cases except things where
the entire body of the function is something trivial like `(plist-get
msgpart field)`.

I want this because I'd like to override mu4e\~headers-contact-str in
my .emacs to e.g. format E-Mail addresses without names like
foo.bar@example.com as "Foo Bar" in the headers view, but
mu4e\~headers-contact-str itself is defsubst, so overriding it doesn't
help, and then it's used in /another/ defsubst'd function,
mu4e\~headers-from-or-to, which in turn is used by the defun'd
mu4e\~headers-field-apply-basic-properties.

So both need to be s/defsubst/defun/'d so you can override them.

More generally, when I asked about this on #emacs on Freenode one
advice was "I have my doubts inlining pays off there" [...] "the
biggest wins are for OO-like access which you'd like to hide behind a
function" [...] "as calling the function takes more time than
accessing a slot, you use defsubst to avoid that".

So I'm leaving the OO-like access as defsubst'd, but I very much doubt
that we're getting anything performance-wise out of having 10-20 line
functions be defsubst'd.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -393,7 +393,7 @@ If SKIP-HOOK is not nil, `mu4e-message-changed-hook' will be invoked."
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(defsubst mu4e~headers-contact-str (contacts)
+(defun mu4e~headers-contact-str (contacts)
   "Turn the list of contacts CONTACTS (with elements (NAME . EMAIL)
 into a string."
   (mapconcat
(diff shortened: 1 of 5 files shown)


djcb:
Anything against `mu4e-header-info-custom`? That's the recommended way to handle such things.

avar:
I didn't know about it, and looking at it it seems to be something I can use to replace my evil hack. So I don't need this integrated for that reason.

Still, seems sensible to make these overridable for other reasons, and as noted substituting them inline probably isn't doing much anyway.

djcb:
They are all in the 'hot path', and should make running them a bit faster; that is why defsubst exists. I don't know much faster (and of course that may vary with emacs version and perhaps other factors), but see no reason for *not* using it.

Anyway, if experienced users don't know about `mu4e-header-info-custom`, we should probably advertise it a bit better!

PR #1071, opened 2017-04-25
mu4e: advance after mark can now be disabled

akoehn:
The variable mu4e-headers-advance-after-mark can be set to nil if you
want to disable moving to the next mail after marking.


djcb:
Looks useful, merged, thanks!

PR #1082, opened 2017-05-18
Added new mu4e-split-view mode: single-window

vsedach:
Single-window mode is meant to minimize mu4e window operations (opening,
killing, resizing, etc) and buffer changes, while still retaining the
view and headers buffers. In addition, it replaces mu4e main view with a
minibuffer prompt containing the same information.

Eventually I would like to do a single buffer mode that can toggle between view and header mode. Single-buffer mode will make it convenient to have multiple mu4e search sessions going on at the same time. Doing a single buffer mode will entail changing mu4e to keep track of messages in a data structure instead of implicitly in the headers buffer, which is a major refactor. This will have the additional benefit of making it possible to write integrations for tools like helm and ivy for browsing the message list and selecting messages; essentially letting people use whatever tool they want instead of the headers buffer. My planned approach is to first provide the message list via imenu, as the package imenu-anywhere integrates helm, ivy, and a few others, which would give mu4e-ivy etc headers integrations "for free." It should also then be possible to integrate mu4e search with helm/ivy flex searching.

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -105,7 +105,7 @@
    Klaus Holst, Lukas Fürmetz, Magnus Therning, Maximilian Matthe,
    Nicolas Richard, Piotr Trojanek, Prashant Sachdeva, Remco van 't
    Veer, Stephen Eglen, Stig Brautaset, Thierry Volpiatto, Thomas
-   Moulia, Titus von der Malsburg, Yuri D'Elia
+   Moulia, Titus von der Malsburg, Yuri D'Elia, Vladimir Sedach
 
 * Old news
   :PROPERTIES:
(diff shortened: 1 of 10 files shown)


vsedach:
Found an issue, please hold off from merging.

vsedach:
Alright, looks like 835d619 fixes the outstanding issues that I can see. Let me know what you think.

djcb:
Thanks! Looks useful and seems to work with a little testing. -- merged the code (manually -- not this PR). Maybe you want to announce / introduce this on the mailing list?

vsedach:
Thank you! Good suggestion. I will make a post on the mailing list next
week when I have returned from traveling.

Vladimir

PR #1088, opened 2017-06-05
Refactor read_key to use g_strrstr() and remove unneeded continues fr…

shaleh:
…om read_val

No functionality change, just code simplification.

g_strrstr() simplifies the logic of splitting the input on the colon.

The ``continue`` does nothing since the if/else logic contains all of the actions.


shaleh:
Grumble. I dropped something when I rebased......

shaleh:
No sense leaving a busted PR.

PR #1089, opened 2017-06-05
Simplify logic of mu_str_remove_ctrl_in_place. Add tests.

shaleh:


djcb:
Merged, with a few stylistic fixes. Thanks!

PR #1094, opened 2017-06-10
mu4e: remove `realdelete' option of OfflineIMAP

wataash:
Quotation from the commit log:

> offlineimap/Changelog.md says Gmail `realdelete' option is considered
> harmful and was removed on OfflineIMAP v6.5.2.1 (2012-04-04).

so we should remove it.

See: https://github.com/OfflineIMAP/offlineimap/blob/master/Changelog.md#offlineimap-v6521-2012-04-04


djcb:
Merged, thanks!

PR #1099, opened 2017-06-16
mu4e: don't loose search history when editing search

stsquad:
Even though the user may be editing this expression there is no reason
to not have the mu4e~headers-search-hist present for the prompt. Emacs
will only replace it with system wide history which would likely
contain irrelevant history for the action.


djcb:
Makes sense -- thanks. Merged.

stsquad:
Is there a time when expr is handed in to the function?

PR #1100, opened 2017-06-17
mu: remove useless assignment in mu-extract

ptroja:
The `filepath` variable is assigned with `NULL`, but not read until reassigned few lines below.

Detected with [cppcheck](http://cppcheck.sourceforge.net/).


djcb:
Don't think removing the assignment improves things, but thank you  for the manual fix!

PR #1105, opened 2017-07-15
Only require html2text if emacs version supports it.

cybercode:
html2text has been obsoleted in emacs 26.  Only require the file if `emacs-version' < "26.0"

### DIFF
--- a/mu4e/mu4e-message.el
+++ b/mu4e/mu4e-message.el
@@ -31,7 +31,7 @@
 (require 'mu4e-utils)
 
 (require 'cl)
-(require 'html2text)
+(and (version< emacs-version "26.0") (require 'html2text))
 (require 'flow-fill)
 
 


djcb:
Thanks! However, we can't remove that yet though, since it is still the default value for `mu4e-html2text-command`, and don't want to change that just yet.  Let's check again when emacs 26 has been released.

cybercode:
If you look at the patch, you will see the current behavior is unchanged. the require is only skipped for emacs 26 and above, where it is never used and causes annoying deprecation warnings when the file is loaded. The initialization form `mu4e-html2text-command` looks for `shr-insert-document` before attempting to use `html2text` anyway, so the change works in all version of emacs. 

I actually used the most conservative approach, specifically checking for emacs 26. I fact a better approach (which i think would no longer load html2text for emacs25 either) would be:
``` elisp
(or (fboundp 'shr-insert-document) (require 'html2text))
```

which is in alignment with the defcustom:

``` elisp
(defcustom mu4e-html2text-command
  (if (fboundp 'shr-insert-document) 'mu4e-shr2text 'html2text)
```

jabranham:
I upgraded to the Emacs 26 pretest today. Loading mu4e 0.9.18 results in the message `Package html2text is obsolete!`. A quick check around and I happened on this PR. Any chance of changing this in the soon-to-be-released mu 1.0.0? @djcb

valankar:
Should this be re-opened now?

PR #1109, opened 2017-07-19
Add info on update status in modeline

galaunay:
## Pull purpose
Just add an indicator "(updating)" in the modeline when an update is running. 
It is disabled by default and can be enabled through the custom variable `mu4e-display-update-status-in-modeline`.
mu4e

## Motivation
I use auto update in the background when displaying the mails of my inbox (to be sure to see the last messages).
It can be annoying to be unable to know when it's done.


djcb:
Thanks! Let's give this a try...

PR #1111, opened 2017-07-30
Support for gmime 3.0

wavexx:
Since gmime 3.0 is available, here's an update for mu, with some notes relevant to mu:

- Fetching from is now uniform with other address fields, making the code cleaner.
- Message references and content parameter lists are now arrays, so no iterators are needed. For references, the code is cleaner. For content types, is on par.
- internet_address_get_name can actually return "" as opposed to NULL, so an extra check is necessary.
- GMIME_ENABLE_RFC2047_WORKAROUNDS is now the default.
- Some functions got renamed or accept a nullable flags parameter. Nothing major.
- Crypto is the biggest change and is the part where I have some doubts on:
  - decrypt and verify no longer require an explicit context, which is actually nice.
  - The agent is now required with the gpg2 machinery, as such the associated options got removed.
  - I don't see a way to specify the gpg path anymore. Some advice needed.
  - Likewise for automatic signature downloading.
  - Error flags are now returned directly in the GMimeSignatureStatus enum, requiring some extra work to get the signature validity and status. Please have a second look on how we remap the signature status.

100% test pass, including crypto (although I verified this by using my own messages).
There are still references to MU_GPG_PATH and MU_MSG_OPTION_AUTO_RETRIEVE to fix.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -146,11 +146,11 @@ AC_SUBST(GLIB_CFLAGS)
 AC_SUBST(GLIB_LIBS)
 glib_version="`$PKG_CONFIG --modversion glib-2.0`"
 
-# gmime, some late-2012 version
-PKG_CHECK_MODULES(GMIME,gmime-2.6 >= 2.6.7)
+# gmime, version 3.0 or higher
+PKG_CHECK_MODULES(GMIME,gmime-3.0)
 AC_SUBST(GMIME_CFLAGS)
 AC_SUBST(GMIME_LIBS)
-gmime_version="`$PKG_CONFIG --modversion gmime-2.6`"
+gmime_version="`$PKG_CONFIG --modversion gmime-3.0`"
 
 
 # xapian checking - we need 1.2.x at least
--- a/contrib/gmime-test.c
+++ b/contrib/gmime-test.c
@@ -31,13 +31,13 @@
 #include <locale.h>
 
 static gchar*
-get_recip (GMimeMessage *msg, GMimeRecipientType rtype)
+get_recip (GMimeMessage *msg, GMimeAddressType atype)
 {
 	char *recep;
 	InternetAddressList *receps;
 
-	receps = g_mime_message_get_recipients (msg, rtype);
-	recep = (char*)internet_address_list_to_string (receps, FALSE);
+	receps = g_mime_message_get_addresses (msg, atype);
+	recep = (char*)internet_address_list_to_string (receps, NULL, FALSE);
 
 	if (!recep || !*recep) {
 		g_free (recep);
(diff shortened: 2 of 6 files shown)


wavexx:
I don't see gmime 3.0 in any ubuntu version yet, and considering how slow travis images generally update, we would need to pull it in via an external source in order to test it there. Shame :/

dralves:
Just dropped in to say thanks. Setting brew to use your gmime-3.0 branch was the only way that I could get mu to build on macOS high sierra. the included gmime 2.6 would always fail to build due to a missing gpg header and I couldn't solve it by installing mono or changing the build flags like some other posts suggest.

wavexx:
On Wed, Dec 06 2017, David Ribeiro Alves wrote:
> Just dropped in to say thanks. Setting brew to use your gmime-3.0
> branch was the only way that I could get mu to build on macOS high
> sierra. the included gmime 2.6 would always fail to build due to a
> missing gpg header and I couldn't solve it by installing mono or
> changing the build flags like some other posts suggest.

It's currently a bit behind the master though.

dralves:
I noticed, was only looking for at least 0.9.16 though, so was recent enough for me. Thanks for pointing that out though.

norbusan:
Hi everyone,
any chance that this is getting updated or even better included into the main branch?

Thanks a lot

Norbert
(Debian maintainer of mu)

Chris00:
@wavexx Ubuntu people should be able to use Debian packages to get gmime-3.0 — maybe they just need a word about how to do.

wavexx:
On Tue, May 29 2018, Christophe Troestler wrote:
> @wavexx Ubuntu people should be able to use Debian packages to get
> gmime-3.0 — maybe they just need a word about how to do.

The latest ubuntu LTS 18.04 has both.

Keeping support for both would probably be quite a bit uglier, and gmime
3 does improve in some corners. It depends whether djcb wants to support
gmime <3 really.

rjhorniii:
I just had to rebuild mu because the latest LEAP 15 from OpenSuSE has some version mismatch issues with the mu package.  That distro has both versions, so you just need clarity in the build instructions and configure error messages.

wavexx:
On Sun, Jun 03 2018, Dirk-Jan C. Binnema wrote:
> ERROR:test-mu-cmd.c:814:test_mu_verify_good: assertion failed (retval == 0): (256 == 0)

It's been a while since I tested this, but it was passing to the end at
the time. I'm currently using gmime 2 as I needed a couple of various
fixes in the master branch, so I didn't use the branch for long.

Did you push any additional fix yet?

rjhorniii:
I'm not sure how my comment on OpenSuse LEAP 15.0 got attached to this.  It's a separate issue.  I reported it to OpenSuse so they could forward it to whoever packages mu for that distribution.  They must have attached it to this by mistake when passing it upstream.

The issue with OpenSuse is different.  The packages that were built for that release show dependencies on versions that are replaced, etc.  The packages won't install properly.  Xapian versions don't find dependencies, etc.

My fix was to take the v1.0  source from github and follow the INSTALL procedure.  That compiled, passed make check, and installed fine.  My assumption is that there is just a problem specific to the packaging instructions for OpenSuse LEAP 15.0.  I don't know enough about that to do any checking or suggest any fixes.

Issue #865, opened 2016-06-08
when answering a gpg encrypted message, the answer should by default be encrypted, too


ArneBab:
## Expected or desired behavior

- I answer an encrypted message.
- It starts with the mime header: <#secure method=pgpmime mode=signencrypt>

## Actual behavior

I have to hit `C-c RET e` everytime I answer an encrypted message, which makes it far too easy to accidentally answer unencrypted (thereby spilling the quoted parts of the message I received) by simply forgetting to explicitly enable encryption.

## Steps to reproduce

hit R on an encrypted email, then hit a to select “reply to all”.

## Versions of mu, mu4e/emacs, operating system etc.

- mu 0.9.16
- emacs 24.5.1
- gnupg 2.0.28

semente:
Also, when replying signed-only messages, would be great if add mime header to sign. An option to always sign encrypted message is welcome too.

There is three kind of PGP messages: signed, encrypted and signed & encrypted:

```
<#secure method=pgpmime mode=sign>
<#secure method=pgpmime mode=encrypt>
<#secure method=pgpmime mode=signencrypt>
```

My proposal is add three options similarly what we have in Gnus:

``` 
Often, you would like to sign replies to people who send you signed messages. Even more often, you might want to encrypt messages which are in reply to encrypted messages. Gnus offers gnus-message-replysign to enable the former, and gnus-message-replyencrypt for the latter. In addition, setting gnus-message-replysignencrypted (on by default) will sign automatically encrypted messages.
```

ArneBab:
that would be great, yes.

rekado:
I'm using this hook to achieve this:

```elisp
(add-hook 'mu4e-compose-mode-hook
  (defun my/maybe-reply-encrypted ()
    (let ((msg mu4e-compose-parent-message))
      (when (and msg (member 'encrypted (mu4e-message-field msg :flags)))
        (mml-secure-message-sign-encrypt)))))
```

ArneBab:
That looks good. Is there a chance to get this into mainline mu4e (so I don’t have to configure the hook for every installation but have it do the right thing automatically instead)?

djcb:
I've added a `mu4e-compose-crypto-reply-policy` configuration variable for this.

ArneBab:
Thank you!

MaximeMaW:
[...]

PR #1118, opened 2017-08-27
Allow to set the reply policy in a smart manner

MaximeMaW:
Depending on whether the original email was encrypted, one can set
different reply policies.

This fixes an issue raised in issue #865, when `mu4e-compose-crypto-reply-policy` conflicted with sign-by-default. Now, two variables replace the old `mu4e-compose-crypto-reply-policy`:

- mu4e-compose-crypto-reply-encrypted-policy
- mu4e-compose-crypto-reply-plain-policy

I'd be very happy to make any adjustment.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -162,7 +162,8 @@ We have the following choices:
 	   :safe 'symbolp
 	   :group 'mu4e-compose))
 
-(defcustom mu4e-compose-crypto-reply-plain-encrypted-policy 'sign  "Policy for signing/encrypting replies to encrypted messages.
+(defcustom mu4e-compose-crypto-reply-plain-policy 'sign
+  "Policy for signing/encrypting replies to messages received unencrypted.
 We have the following choices:
 
 - `sign': sign the reply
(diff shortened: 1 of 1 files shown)


mu4e/mu4e-compose.el
```diff
@@ -146,7 +146,7 @@ Also see `mu4e-context-policy'."
 	   :group 'mu4e-compose))
 
 
-(defcustom mu4e-compose-crypto-reply-policy 'sign-and-encrypt
+(defcustom mu4e-compose-crypto-reply-encrypted-policy 'sign-and-encrypt
   "Policy for signing/encrypting replies to encrypted messages.
 We have the following choices:
```
djcb: I think we'd need a `make-obsolete-variable`, to warn users of the old variable.

mu4e/mu4e-compose.el
```diff
@@ -162,6 +162,22 @@ We have the following choices:
 	   :safe 'symbolp
 	   :group 'mu4e-compose))
 
+(defcustom mu4e-compose-crypto-reply-plain-encrypted-policy 'sign  "Policy for signing/encrypting replies to encrypted messages.
+We have the following choices:
```
djcb: this line is a bit too long - please limit it to 80 chars

mu4e/mu4e-compose.el
```diff
@@ -525,13 +541,23 @@ buffers; lets remap its faces so it uses the ones for mu4e."
 
 (defun mu4e~compose-crypto-reply (parent compose-type)
   "When composing a reply to an encrypted message, we can
-automatically encrypt that reply."
-  (when (and  (eq compose-type 'reply)
+automatically encrypt that reply. When the message is unencrypted, 
+we can decide what we want to do."
+  (message "%S %S" parent compose-type)
+  (if (and  (eq compose-type 'reply)
```
djcb: Those debug `message` things should probably go.

djcb:
Anyway, apart from those details, looks good!

mu4e/mu4e-compose.el
MaximeMaW: I'm not sure about the standard way of doing that. I proposed something in commit  e761f74.

MaximeMaW:
@djcb, any comment on the updated commits (e761f74)?

Thank you very much!

djcb:
Merged, thanks!

PR #1134, opened 2017-10-09
mu4e: simplify what "e" does in mu4e-view mode

jabranham:
The default was to use e to save one attachment or C-u e to save multiple. This
simplifies it so that e simply offers to save one or many attachments.


jabranham:
I've been using this for 3 months now without any bad side effects FWIW.

djcb:
Merged -- thanks

PR #1141, opened 2017-10-29
Ensure mu4e-view-mode-hook run after text insertion

thierryvolpiatto:
* mu4e/mu4e-view.el (mu4e-view): Do it.

In `mu4e-view` `mu4e-view-mode` run before any text is inserted, as a result all functions from 
`mu4e-view-mode-hook` modifying text in message have no effect.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -306,13 +306,13 @@ the the message view affects HDRSBUF, as does marking etc.
 As a side-effect, a message that is being viewed loses its 'unread'
 marking if it still had that."
   (let* ((embedded ;; is it as an embedded msg (ie. message/rfc822 att)?
-          (when (gethash (mu4e-message-field msg :path)
-                         mu4e~path-parent-docid-map) t))
-         (buf
-          (if embedded
+	   (when (gethash (mu4e-message-field msg :path)
+		   mu4e~path-parent-docid-map) t))
+	  (buf
+	    (if embedded
 	      (mu4e~view-embedded-winbuf)
-            (get-buffer-create mu4e~view-buffer-name)))
-         (insert-it (or embedded (not (mu4e~view-mark-as-read-maybe msg)))))
+	      (get-buffer-create mu4e~view-buffer-name)))
+          (insert-it (or embedded (not (mu4e~view-mark-as-read-maybe msg)))))
     (with-current-buffer buf
       (switch-to-buffer buf)
       ;; When MSG is unread, mu4e~view-mark-as-read-maybe will trigger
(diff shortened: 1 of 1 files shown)


djcb:
thanks -- but can you change it to not affect the whitespace?

thierryvolpiatto:

> thanks -- but can you change it to not affect the whitespace?

Done.

-- 
Thierry

djcb:
Merged -- thanks!

wavexx:
Mmmh, this causes:

```
Debugger entered--Lisp error: (wrong-type-argument hash-table-p nil)
  hash-table-count(nil)
  mu4e~view-get-attach-num(....)
````

when opening attachments.

**Comment by thierryvolpiatto** [comment]
Indeed this is because `mu4e-view-message-text` calls `mu4e~view-construct-attachments-header` which set `mu4e~view-attach-map` for current buffer, and then `mu4e-view-mode` which run after this clean local variables, so `mu4e~view-attach-map` is reset to nil.
So we must call the mode before calling `mu4e-view-message-text`.
@djcb I suggest you reverse this PR for now and we find something else.
A simple change would be calling the hook with `run-mode-hooks` at end of function, but hooks would be called twice, a workaround for this would be to let-bind `mu4e-view-mode-hook` to nil when calling the mode, like this:

```elisp
(defun mu4e-view (msg)
  "Display the message MSG in a new buffer, and keep in sync with HDRSBUF.
'In sync' here means that moving to the next/previous message in
the the message view affects HDRSBUF, as does marking etc.

As a side-effect, a message that is being viewed loses its 'unread'
marking if it still had that."
  (let* ((embedded ;; is it as an embedded msg (ie. message/rfc822 att)?
	   (when (gethash (mu4e-message-field msg :path)
		   mu4e~path-parent-docid-map) t))
	  (buf
	    (if embedded
	      (mu4e~view-embedded-winbuf)
	      (get-buffer-create mu4e~view-buffer-name)))
         mode-enabled)
    (with-current-buffer buf
      (unless (setq mode-enabled (eq major-mode 'mu4e-view-mode))
        (let (mu4e-view-mode-hook) (mu4e-view-mode)))
      (setq mu4e~view-msg msg)
      ;; When MSG is unread, mu4e~view-mark-as-read-maybe will trigger
      ;; another call to mu4e-view (via mu4e~headers-update-handler as
      ;; the reply handler to mu4e~proc-move)
      (let ((inhibit-read-only t))
        (when (or embedded (not (mu4e~view-mark-as-read-maybe msg)))
	  (erase-buffer)
	  (mu4e~delete-all-overlays)
	  (insert (mu4e-view-message-text msg))
	  (goto-char (point-min))
	  (mu4e~fontify-cited)
	  (mu4e~fontify-signature)
	  (mu4e~view-show-images-maybe msg)
          (mu4e~view-make-urls-clickable)
	  (when embedded (local-set-key "q" 'kill-buffer-and-window))
          (unless mode-enabled (run-mode-hooks 'mu4e-view-mode-hook)))))
    (switch-to-buffer buf)))
```
Another solution would be to add a new hook...

PR #1146, opened 2017-10-30
mu4e: fix bug preventing mu4e-headers-change-sorting to sort by list

mekeor:
The little bug was caused by a little confusion of the key for list-IDs.
Namely, we use `:list` and not `:list-id` for list-IDs.


djcb:
Merged -- thanks!

PR #1148, opened 2017-11-04
Fix again #1141 (mu4e-view-mode-hook should run after text insertion)

thierryvolpiatto:
Run mu4e-view-mode without its hook (#1141)
before text insertion and run finally the hook at end.
This ensure local variables are not killed after being set.
* mu4e/mu4e-view.el (mu4e-view): Do it.


thierryvolpiatto:
I am using this since 17 days now as an advice without any problems.

thierryvolpiatto:
ping

djcb:
Merged -- thanks!

PR #1157, opened 2017-11-26
Use defcustom for the customizable variables in mu4e-lists.el

andersjohansson:
Kind of self-explanatory I hope!


djcb:
Merged -- thanks!

PR #1161, opened 2017-12-03
shorten modeline str if longer than max-width

gambhiro:
This is a fix to limit the length of the query string displayed in the `spaceline` modeline.

The `spaceline` package displays the query in a segment of the modeline. When this is a long (perhaps bookmarked) query, the modeline either (a) becomes crowded or (b) another problem manifests where the whole spaceline modeline goes blank.

The variable `mu4e-modeline-max-width` is introduced to limit the length and `mu4e~quote-for-modeline` is extended a bit to shorten the query string if needed.


JoeOsborn:
I've run into the issue that this PR fixes!  It would be a big help for Spaceline users.

djcb:
Merged -- thanks!

PR #1162, opened 2017-12-05
mu4e: don’t use msgid as format template

eigengrau:
The Message-Id header may contain percent characters, which we don’t want to
pass on as format specifiers. Instead of escaping these characters too, we can
instead get rid of the call to format when a string has already been supplied.


eigengrau:
Additionally, this fixes the reference to `msgid`, which should not have been bound here as a variable. For some reason, this didn’t trip an error for me (maybe the name crept in as a global variable?).

djcb:
A very, very similar fix has gone in already. Anyway, thank you!

PR #1163, opened 2017-12-05
mu4e/mu4e-vars.el: replace references to mu4e-message-get-field

eqyiel:
This is just a tiny fix for doc strings that confused me when I was looking for the function `mu4e-message-field`.


djcb:
Merged -- thanks!

PR #1169, opened 2017-12-19
Bump version string to 1.0-alpha3

jschwab:
The value in configure.ac is out of sync with the git tags.


djcb:
Ah yes, forgot to push that. Thanks!

PR #1176, opened 2018-01-12
No longer use obsolete alias speedbar-with-attached-buffer

tarsius:
Instead use the aliased variable `dframe-with-attached-buffer`.
The alias was established in Emacs 24.4.

### DIFF
--- a/mu4e/mu4e-speedbar.el
+++ b/mu4e/mu4e-speedbar.el
@@ -82,7 +82,7 @@
 
 (defun mu4e~speedbar-maildir (&optional text token ident)
   "Jump to maildir TOKEN. TEXT and INDENT are not used."
-  (speedbar-with-attached-buffer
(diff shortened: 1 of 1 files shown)


tarsius:
Didn't know you are still supporting 24.1.
Would you accept a patch that wrapped these forms with `with-no-warnings` instead?

djcb:
Let's look at this again after the mu 1.0 release. Perhaps then I should bump the emacs version requirement.

tarsius:
Okay. Thanks for the heads up!

tarsius:
> Let's look at this again after the mu 1.0 release.

Could you please have another look now?

> Perhaps then I should bump the emacs version requirement.

What's the plan here?

djcb:
Merged, thanks!

tarsius:
Thanks to you too!

PR #1190, opened 2018-02-09
mu4e: ensure sent dir is created with proper maildir structure

jwhitbeck:
offlineimap will only sync directories that have 'cur', 'tmp', and
'new' sub-directories. Currently, mu4e can end up in a state where the
sent folder only has the 'cur' directory and is therefore not synced.

Upon saving a sent email, emacs' write-file will prompt to create
parent directories if the sent maildir does not already exist. This
results in the following directory structure:

`<mu4e-maildir>/<sent>/cur/<email>`

Note that the 'tmp' and 'new' directories are missing. This commit
ensures that they are always created.


djcb:
Looks good, merged. Thanks!

PR #1198, opened 2018-02-19
Only require html2text if it will be used

c-nixon:
html2text is deprecated in the emacs pretest, by unconditionally
importing it mu4e causes a mildly irritating yes/no prompt to appear
during startup.

This change ensures that html2text is not loaded unless the emacs
running is a version that does not have 'shr-insert-document


djcb:
Looks good, merged. Thanks!

PR #1208, opened 2018-03-18
Fix display of bugs link in mu4e-about.org

eklitzke:
It appears that these need to be on the same line for Emacs to render the link correctly.


djcb:
Merged, thanks.

PR #1211, opened 2018-03-22
Use automake support for texi to generate version info in the manual

egli:
Automake contains some support for automagically adding version information to a texi manual (see
https://www.gnu.org/software/automake/manual/automake.html#Texinfo).

> If the .texi file @includes version.texi, then that file will be automatically generated

This patch gets rid of the texi.texi.in and instead uses the standard method to inject version information in the manual.

This fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=870634


djcb:
Merged -- thanks!

PR #1212, opened 2018-03-24
mu4e.texi: Fix and update external links

tarleb:
External links are updated to point to the most informative pages. Pure
http links are upgraded to https were possible.

The only broken link is the one for *org-contacts*. I wasn't able to find a current page for that project.


djcb:
Merged. Thanks!

PR #1228, opened 2018-04-23
Mu4e improved thread prefix

mhcerri:
Hi!

This PR adds support for thread trees similar to mutt. The new style should look like that:

```
Original mail subject
└>
 └>
  ├>
  │├>
  │└>
  │ └>
  └>
```

In opposition to the current style:
```
+ Original mail subject
 + 
  + 
   + 
    \ 
    + 
     \ 
   | 
```
I added the config `mu4e-headers-include-related` to make it possible to switch between the new and old styles. I also added a couple of prefix configuration, in a similar way the current code does, to allow further customization.

Please feel free to proposed any changes.


wavexx:
This is awesome! The original mu thread rendering is indeed quite lacking.

djcb:
Oh, looks good, definitely an improvement. In fact, I'd be happy if you remove "classic" style. Oh, and can you update the "Header view" overview in the manual? Thanks.

mhcerri:
Thanks for the feedback, @djcb. I have updated the following:

1. Removed the "classic" style and renamed the prefix variables to match the pattern used for the old ones (I didn't reuse the same names though).
2. Replaced the lexical binding used for the state with a local buffer variable. I haven't found any concurrency issues with the previous approach, but that should be safer.
3. Updated the header view example in the manual with the new prefix style.
4. Changed the default prefix for orphan messages to empty. That should be explicit already by the indentation level.

stardiviner:
Finally improved this thread prefix. This is really great!!

djcb:
Merged, thanks!

mhcerri:
Good point. I agree.

I simply added the option of adding a mark in front of the message when its parent is empty, but I left the default value empty since I considered the identation level would make that explicit.

However, considering your example, that might be very confusing indeed. The alternative you proposed is actually how mutt handles orphan messages and it's a very good solution.

From the configuration perspective, it will require two prefixea though, since we will have to specify one additional prefix for orphan messages that don't have siblings (something like "-->").

wavexx:
On Thu, Apr 26 2018, Marcelo Cerri wrote:
> However, considering your example, that might be very confusing
> indeed. The alternative you proposed is actually how mutt handles
> orphan messages and it's a very good solution.

If you toggle mu4e-headers-include-related to t, this happens much more
frequently.

> From the configuration perspective, it will require two prefixea
> though, since we will have to specify one additional prefix for orphan
> messages that don't have siblings (something like "-->").

This would be nice for completeness, but AFAIK mutt doesn't use anything
special in this case.

mhcerri:
Hi, @wavexx. Can you test the PR #1236 to check if that makes sense to you?

PR #1231, opened 2018-04-24
fix: image not show the first time we view a mail

tangxinfa:
Prepare settings:

       setq mu4e-view-show-images t
       setq mu4e-view-auto-mark-as-read t

We receive a mail with image attachments, when we view it, mu4e will send a `view` command to mu, mu will response a `view` msg with image attachments(:temp not nil), then mu4e send a `move` command to mu for mark as read, mu responsed with a `update` msg without attachments(:temp is nil), mu4e display it with `mu4e-view`, then image lost.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -365,7 +365,7 @@ headers."
 	  ;; update it; that way, the flags can be updated, as well as the path
 	  ;; (which is useful for viewing the raw message)
 	  (when (mu4e~headers-view-this-message-p docid)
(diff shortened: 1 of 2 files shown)


djcb:
Oh, great catch, that's been a longstanding problem, and you have found the root cause!  I've fixed it in a slightly different way, but many thanks for hunting this down!

tangxinfa:
thanks, it works in the latest version!

joostkremers:
Great, thanks! It works!

PR #1236, opened 2018-05-03
Mu4e orphan thread prefix

mhcerri:
Use special prefixes for orphan threads as suggested by @wavexx. It should look like:

```
Unrelated message
─>Single orphan message
┬>Orphan messsage
├>Second message on orphan thread
└>Last message on orphan thread
```

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -248,8 +248,11 @@ This prefix should have the same length as `mu4e-headers-thread-blank-prefix'.")
 
 This prefix should have the same length as `mu4e-headers-thread-connection-prefix'.")
 
-(defvar mu4e-headers-thread-orphan-prefix '("" . "")
-  "Prefix for orphan messages.")
+(defvar mu4e-headers-thread-orphan-prefix '("┬>" . "┳▶ ")
+  "Prefix for orphan messages with siblings.")
+
+(defvar mu4e-headers-thread-single-orphan-prefix '("─>" . "━▶ ")
+  "Prefix for orphan messages with no siblings.")
 
 (defvar mu4e-headers-thread-duplicate-prefix '("=" . "≡ ")
   "Prefix for duplicate messages.")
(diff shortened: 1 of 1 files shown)


wavexx:
Definitely much better! This always makes the separate threads obvious.
I have an hard time triggering a single orphan message though.

Some things I noticed:

2018-05-03t120222

In the above example, I guess the thread rendering is done before duplicates are removed by mu4e-headers-toggle-skip-duplicates, right? If I include the duplicate messages, the rendering looks right:

2018-05-03t120511

Also, there is probably a limit of a single mu4e-headers-thread-connection-prefix. On a thread with multiple replies, only the right-most connection seems to be rendered:

2018-05-03t120240

I mean, these are just minor quibbles I'm pointing out in case if weren't aware. The thread rendering is still a massive improvement over the original mu4e!

mhcerri:
Hi, @wavexx.

Thanks for the feedback. I wasn't aware of those issues.

Regarding the first one, I do agree it's probably caused by the moment when the duplicates are removed by `mu`. I will take a closer look at it later.

For the later issue, I believe that's a bug on the code I added to construct the visual tree. I was assuming that a message with a missing parent couldn't appear in the middle of a thread. I included one additional commit to the PR, can you check how your thread looks like with that change?

Thank you.

wavexx:
Before I give it a shot, do you have by chance a small maildir with a
sigle orphan thread?

I tried to make one, but couldn't.

mhcerri:
I couldn't make one either. At least not an orphan message that would show up with the single-orphan prefix. Maybe `mu` doesn't level indent this kind of message.

When I manually delete a parent message that only has a single child, that single child appears as a regular message like that:

```
Re: Original subject
```
Since it doesn't have a prefix, that means its level is zero.

And only when I respond to the parent message I get the messages indented at level 1, resulting in:

```
┬>Re: Original subject
└>
```
I also searched my mailing list maildirs (where I have most of my broken threads) and that seems to confirm that.

Maybe we can drop the single-orphan prefix.

wavexx:
2018-05-03t153316

Looks much better, but still not completely connected. To be honest, this is a thread from a ticketing system, so the references might be broken, resulting in that mess.

Checking another mailing list, it seems that the rendering of multiple connecting lines looks fine there:

2018-05-03t153522
2018-05-03t153607

wavexx:
Uhm, looking again at that first thread (the not-completely-connected), it looks like that there are two kinds of joins missing:

- a single "|-" connected to a subsequent "┬>", or a single "-". These look like are possible due to the shuffling caused by the date sorting.

djcb:
Merged -- thanks!

wavexx:
On Fri, May 11 2018, Dirk-Jan C. Binnema wrote:
> Merged -- thanks!

I think the single-orphan prefix should probably removed, as it's not
used.

I was checking with Gnus the other day, where threads like this are
called "false-root".

sdycjsj:
> 2018-05-03t120222
> 
> In the above example, I guess the thread rendering is done before duplicates are removed by mu4e-headers-toggle-skip-duplicates, right? If I include the duplicate messages, the rendering looks right:
> 
> 2018-05-03t120511
> 

Hi, @wavexx 

Just want to make sure. Is the first issue, i.e., threading before removing duplicated, fixed?
I am using a version built from a git clone of the latest repo of mu, with a gmail account.
When looking at the "All Mail" folder, with mu4e-headers-skip-duplicates set to t, there are disconnected entries and open ends in some of the threads.
If I press V to toggle this option to nil, all threads are complete at the price of showing duplicated messages.
I guess this is similar to the the first issue you mentioned, but since it was several months ago, I am not sure whether I am experiencing the same issue.

Thanks

wavexx:
On Fri, Dec 21 2018, sdycjsj wrote:
> Just want to make sure. Is the first issue, i.e., threading before
> removing duplicated, fixed? I am using a version built from a git
> clone of the latest repo of mu, with a gmail account. When looking at
> the "All Mail" folder, with mu4e-headers-skip-duplicates set to t,
> there are disconnected entries and open ends in some of the threads.
> If I press V to toggle this option to nil, all threads are complete at
> the price of showing duplicated messages. I guess this is similar to
> the the first issue you mentioned, but since it was several months
> ago, I am not sure whether I am experiencing the same issue.

It's indeed the same issue, and, no, the issue has never been fixed so
far.

sdycjsj:
> It's indeed the same issue, and, no, the issue has never been fixed so far.

Thanks you for the clarification.
I will just turn off the thread view for gmail.

PR #1240, opened 2018-05-15
mu4e-conversation: Init

Ambrevar:
This is an early prototype.

Call `mu4e-conversation` on a message from the headers buffer (the more complicated the thread, the nicer the preview!).

Features:
- Display all messages from a thread in a single buffer.
- Each message is colored according to the sender.
- The user's own messages are in a separate color.
- If the thread has only one message, fallback to the regular view.

Planned features:
- Add support for images and attachments.
- Make a dedicated mode with bindings to browse the messages, fold them, etc.
- Mark messages as read as they appear.

Bugs:
- Current implementation works by setting some handlers to temporary functions.  It's rather brittle since an error in the process will leave the handlers set to their temporary value, which breaks mu4e.
- When opening a conversation, the headers buffer is emptied.  I don't know why. 

Comments are welcome and help would be very much needed for those two bugs.

This is still an early draft and the code might not be as clear as it should be.  Don't hesitate to ask for clarifications.

### DIFF
--- a/mu4e/mu4e-conversation.el
+++ b/mu4e/mu4e-conversation.el
@@ -136,6 +136,20 @@ The message can be retrieved from `mu4e~conversation-thread'.")
   "Face for conversation message sent by someone else."
   :group 'mu4e-faces)
 
+(defun mu4e-conversation-trim-bottom-quote (message)
+  "Trim the replied-to emails quoted at the end of message."
+  (require 'rx)
+  (replace-regexp-in-string (rx (+ line-start
+                                 (+ ">")
+                                 (* not-newline)
+                                 (? ?\n))
+                              (* (or space blank ?\n))
+                              (or
+                               string-end
+                               (and line-start  "--" (* space) ?\n (* anything))))
+                            "[...]"
+                            message))
+
 (defun mu4e-conversation-print-message (index)
   "Insert formatted message found at INDEX in `mu4e~conversation-thread'."
   ;; See the docstring of `mu4e-message-field-raw'.
@@ -168,7 +182,8 @@ The message can be retrieved from `mu4e~conversation-thread'.")
                                         (mu4e-message-field msg :flags)))
                         'face
                         'mu4e-conversation-header)
-            (propertize (mu4e-message-body-text msg) 'face
+            ;; TODO: Add button to display trimmed quote.
+            (propertize (mu4e-conversation-trim-bottom-quote (mu4e-message-body-text msg)) 'face
                         (if from-me-p
                             'mu4e-conversation-sender-me
                           (gethash (cdr from) sender-faces)))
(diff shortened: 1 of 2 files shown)


Ambrevar:
screen-2018-05-15-18 52 59

emacsomancer:
This is really nice, @Ambrevar . 

In addition to your current planned features, one really nice feature would be the ability to hide/collapse blocks of >-quoted text; especially at end of a message (and/or before a signature). [maybe also the ability to hide signatures, at least one's own].

Ambrevar:
@emacsomancer: I forgot to mention it but I have already started working on such a feature!
It will work just like Gmail's `...`.

Ambrevar:
@djcb: What's your opinion on this?
I'll finish it very soon.

djcb:
Nice work!

I think this should be an alternative to the current way, not a replacement (at least for the foreseeable future). So, for now e.g. in the headers-view, `C-RET`  would activate the conversation-view while `RET` would just show the message as we do now.

`mu4e-conversation-print-message` is a simpler version of what we do in mu4e-view, so people might miss the all those features. We'd probably want to have only one implementation. Something to look into is `mu4e-view-use-gnus`, which I recently added; when enabled, it uses `gnus` for the message display, and makes mu4e-view much simpler.
  
Apart from that, perhaps you can ship conversation mode as an external thing for now? That gives us some time to test and vet various scenarios, and then decide if we can integrate it with main mu4e. Make sure to announce it on the mu mailing-list, I bet many people want to play with this!

emacsomancer:
@Ambrevar Great work!  One bugreport: The gmail-style quote trimming doesn't seem to apply to sender messages but only to others' messages.

Ambrevar:
Thanks for the report!
Well, actually the quote trimming is unfinished and probably completely
broken :)

I'll finish the implementation soon.

Ambrevar:
Alright, I've published a much more polished version there: https://notabug.org/Ambrevar/mu4e-conversation.

New features:
- Tree view in Org-mode!
- Initial bindings (next message, view toggle).
- Fix header display.
- Handler overrides are now done with advices, which is a bit cleaner but still not perfect.

Planned features:
- Message folding in linear view.
- Quote trimming.
- URL display + bindings. (I'll look into `mu4e-view-use-gnus`.)
- More bindings, including the very needed "reply".

Dreamed features:
- Colour quotes using the colour of the original author.

Feedback is very welcome!
(Please report issues on the new repository.)

Ambrevar:
Quote trimming is now implemented.  @djcb: you might like the implementation.
Bindings are much more complete now.  URL display mostly work.

I'm now hosting the project on https://gitlab.com/ambrevar/mu4e-conversation.

Also see the annoucement on Reddit: https://www.reddit.com/r/emacs/comments/8qryri/mu4econversation_single_buffer_fullthread_display/.

Enjoy!

PR #1241, opened 2018-05-18
Fix call to make-obsolete-variable

jabranham:
Missing the quote makes `nil` obsolete instead of `mu4e-compose-crypto-reply-policy`.


djcb:
Oh, good catch, thanks!

PR #1242, opened 2018-05-21
Patch application updates

stsquad:
A couple of tweaks to make patch application better.

### DIFF
--- a/mu4e/mu4e-actions.el
+++ b/mu4e/mu4e-actions.el
@@ -217,36 +217,44 @@ store your org-contacts."
 
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defvar mu4e~patch-directory-history nil
+  "History of directories we have applied patches to.")
+
+;; This essentially works around the fact that read-directory-name
+;; can't have custom history.
+(defun mu4e~read-patch-directory (&optional prompt)
+  "Read a `PROMPT'ed directory name via `completing-read' with history."
+  (unless prompt
+    (setq prompt "Target directory:"))
+  (file-truename
+   (completing-read prompt 'read-file-name-internal #'file-directory-p
+                    nil nil 'mu4e~patch-directory-history)))
+
 (defun mu4e-action-git-apply-patch (msg)
-  "Apply the git [patch] message."
-  (let ((path (ido-read-directory-name "Target directory: "
-		(car ido-work-directory-list)
-		"~/" t)))
-    (setf ido-work-directory-list
-      (cons path (delete path ido-work-directory-list)))
-    (shell-command
-      (format "cd %s; git apply %s"
-	path
-	(mu4e-message-field msg :path)))))
+  "Apply `MSG' as a git patch."
+  (let ((path (mu4e~read-patch-directory "Target directory: ")))
+    (let ((default-directory path))
+      (shell-command
+       (format "git apply %s"
+               (shell-quote-argument (mu4e-message-field msg :path)))))))
 
 (defun mu4e-action-git-apply-mbox (msg &optional signoff)
-  "Apply and commit the git [patch] MSG with optional SIGNOFF.
+  "Apply `MSG' a git patch with optional `SIGNOFF'.
 
 If the `default-directory' matches the most recent history entry don't
 bother asking for the git tree again (useful for bulk actions)."
 
-  (let ((cwd (car ido-work-directory-list)))
+  (let ((cwd (substring-no-properties
+              (or (car mu4e~patch-directory-history)
+                  "not-a-dir"))))
     (unless (and (stringp cwd) (string= default-directory cwd))
-      (setq cwd (ido-read-directory-name "Target directory: "
-		  cwd
-		  "~/" t))
-      (setf ido-work-directory-list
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Alex Bennée <notifications@github.com> writes:

> A couple of tweaks to make patch application better.

Seems your completing-read is not sorting directories, perhaps you can
use something like this:

    (completing-read "Target directory: " (lambda (&rest args)
                                            (sort (apply 'read-file-name-internal args) 'string-lessp))
                     #'file-directory-p
                     nil nil 'mu4e--patch-directory-history)

-- 
Thierry

stsquad:
@thierryvolpiatto: well the idea is to let completing read do it's own thing - in my case it uses ivy. However with your suggestion I loose the flexibility of ivy selection.

thierryvolpiatto:
Note also that you can only access directories under default-directory with such a completing-read.

stsquad:
@thierryvolpiatto I just tried with emacs -q and I could certainly read a directory above where I was (although it came out as ../../../foo). Maybe it would be enough to add a file-truename in-front?

thierryvolpiatto:
Alex Bennée <notifications@github.com> writes:

> @thierryvolpiatto I just tried with emacs -q and I could certainly
> read a directory above where I was

It doesn't with helm unless specifying a dynamic handler for your fn, but I just
fixed it for this special case and it works without this requirement if
you specify in your function you are completing filenames by let-binding
minibuffer-completing-file-name.

> (although it came out as ../../../foo).

Not here with helm, it is working fine as long you enter in minibuffer
the full path e.g. ~/

> Maybe it would be enough to add a file-truename in-front?

Don't know.


PS: I am interested in this because I would like to fix some corner
cases like this in helm-mode, but I have to say that using
completing-read+read-file-name-internal [1] is not the way to go, really,
you have better time to use read-directory-name which provide a much
enhanced completion even if you lack a history argument.

[1] completing-read+read-file-name-internal is what ffap use for
find-file-at-point and it is a really bad implementation.

--
Thierry

stsquad:
@thierryvolpiatto I would have used read-directory-name but that loses the ability have task specific history. But yes the idea of this patch was to allow ido, ivy and helm all hook into the completing read function rather than forcing idio on everyone.

thierryvolpiatto:
Alex Bennée <notifications@github.com> writes:

> @thierryvolpiatto I would have used read-directory-name but that loses
> the ability have task specific history.

Yes I understood this.

> But yes the idea of this patch was to allow ido, ivy and helm all hook
> into the completing read function rather than forcing idio on
> everyone.

Yes, that's nice, thanks, for helm I have commited a change that allow
to use your completing-read if you let-bind
minibuffer-completing-file-name around it.

For ido, I am not sure it is working with read-file-name-internal as
COLLECTION argument (or any function):

(ido-completing-read "test: " 'read-file-name-internal #'file-directory-p)
=>
Debugger entered--Lisp error: (wrong-type-argument sequencep read-file-name-internal)
  reverse(read-file-name-internal)
  ido-set-matches()
  ido-read-internal(list "test: " nil nil nil nil)
  ido-completing-read("test: " read-file-name-internal file-directory-p)


Thanks.

-- 
Thierry

mu4e/mu4e-actions.el
```diff
@@ -217,35 +217,42 @@ store your org-contacts."
...
 

 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+(defvar mu4e--patch-directory-history nil
+  "History of directories we have applied patches to.")
+
+;; This essentially works around the fact that read-directory-name
+;; can't have custom history.
+(defun mu4e--read-patch-directory (&optional prompt)
```
djcb: please use `~` instead of `--` for internal functions.

mu4e/mu4e-actions.el
```diff
@@ -217,35 +217,42 @@ store your org-contacts."
...
-      (cons path (delete path ido-work-directory-list)))
+  "Apply `MSG' as a git patch."
+  (let ((path (mu4e--read-patch-directory "Target directory: ")))
     (shell-command
-      (format "cd %s; git apply %s"
-	path
-	(mu4e-message-field msg :path)))))
+     (format "cd %s; git apply %s"
+             path
+             (shell-quote-argument (mu4e-message-field msg :path))))))
```
djcb: when we're changing things here anyway, `path` should be escaped here as well, and we should let-bind `default-directory` rather than `cd` in the shell-command.

mu4e/mu4e-actions.el
```diff
@@ -217,35 +217,42 @@ store your org-contacts."
...
-		  "~/" t))
-      (setf ido-work-directory-list
-	(cons cwd (delete cwd ido-work-directory-list))))
+      (setq cwd (mu4e--read-patch-directory "Target directory: ")))
     (shell-command
-      (format "cd %s; git am %s"
-	(shell-quote-argument cwd)
-	(shell-quote-argument (mu4e-message-field msg :path))))))
+      (format "cd %s; git am %s %s"
+              (shell-quote-argument cwd)
```
djcb: Same here; let-bind `default-directory`

stsquad:
@djcb thanks for the review, I'll fold those in. I still need to find a reasonable way to run through the various completion options to check they all behave sanely.

djcb:
Merged -- thanks!

PR #1243, opened 2018-05-23
view: render URLs as "<url> [%d]" not "<url>[%d]"

avar:
When we render URLs like https://gnu.org/[1] right-clicking on them in
e.g. GNOME Terminal will also copy the "[1]" as well as the URL,
inserting whitespace between the two avoids this.

I know about "g" (mu4e-view-go-to-url), but sometimes I want to open a
URL in a different browser, or copy it into a non-Emacs program. This
makes that easier. I think this improves the UI at a very trivial cost
to users that don't care about this use-case.

I could make this configurable, but unless someone vehemently objects
to this I don't see the point of not just making it the default.


djcb:
I'd like to keep the [1] right next to the url... perhaps there's some zero-width Unicode char we can use so terminals will do the right thing?

avar:
@djcb : Late reply, sorry. Yeah that works, GitHub's diff viewer doesn't do a good job of showing it, but a `git diff` will show:

```
-               (propertize (format "[%d]" num)
+               (propertize (format "<U+200B>[%d]" num)
```

As noted in the updated commit message this works.

djcb:
Oh, great, thanks! But could you make the <U+200B> explicit, so it's visible in the code (ie., `\uxxxx`) so it's clear when reading the code? Thanks.

avar:
Makes sense, updated to use that.

Issue #1244, opened 2018-05-25
Extra empty line at the beginning of messages


Profpatsch:
The `mail-header-separator` is stripped quite literally, causing a message like:

```
To: foo@foo.com
--text follows this line--

Hi @all
```

to result in this sent mail:

```
To: foo@foo.com


Hi @all
```
(Two newlines)

One could always start one’s messages directly after the line:

```
--text follows this line--
Hi @all
```

but even the `reply` shortcut inserts a line by default:

```
--text follows this line--

Mr X wrote:
>
```

so by default every reply sent from `mu4e` starts with two newlines.

There seems to be some non-trivial stripping logic involved (I’ve seen functions like `mu4e~draft-insert-mail-header-separator`, `mu4e~compose-insert-mail-header-separator` and `mu4e~compose-remove-mail-header-separator`) so I don’t think I can reliably create a patch that solves the problem exhaustively.

djcb:
I think this is correct -- the end of the headers is marked by a blank line as per the RFCs.

In the reply case, we do add an empty line and place the cursor there, so people can start typing their reply.

PR #1246, opened 2018-05-26
Remove mail-header-separator newline on draft save

Profpatsch:
This is my take on #1244.

Somewhere in the middle I noticed that the problem most likely lies within `message-mode` instead, which introduces the separator and removes it as well on `message-send`.

This does solve the problem of #1244, but removes one line of whitespace on each draft save. :roll_eyes: 
So it’s just a temporary fix at best and should probably go into `message-mode`.


djcb:
As discussed in #1244, I don't think this is needed. Still, thank you for the effort!

PR #1251, opened 2018-05-28
mu4e: show sender mail addresses by default

Profpatsch:
Hiding email addresses poses a big security risk for the receiver, since
anybody can set the display name however they wish and thus impersonate another
person trivially.

Before this change, `Alice <admin@mallory.com>` would be shown as `Alice` to
Bob, who was expecting Mail from `Alice <alice@gmail.com>`.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -66,7 +66,7 @@ For the complete list of available headers, see `mu4e-header-info'."
   :group 'mu4e-view)
 
 
-(defcustom mu4e-view-show-addresses nil
+(defcustom mu4e-view-show-addresses t
   "Whether to initially show full e-mail addresses for contacts in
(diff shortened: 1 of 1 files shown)


djcb:
You can also change the sender-address easily, so in general you can't trust that part (without signing); so showing the address doesn't make much difference. Still, if you want, there's the setting.

So, appreciate the effort, but not taking the PR. Thanks.

Profpatsch:
> You can also change the sender-address easily, so in general you can't trust that part (without signing)

Ah, that’s exactly what DKIM does today. So at least you can be mostly assured that the email wouldn’t have reached you if the sender didn’t control the domain part (and for services like `gmail` you can be reasonably sure that the user part has semantic meaning as well). Exactly that domain vanishes if you trust display names, which can be freely set in any scheme.

PR #1256, opened 2018-06-06
Use better warning/hardening compiler flags, and fix some one byte strncpy() overflows

eklitzke:
This adds better support in the makefile for `-Wall`. There was some existing `WARN_CFLAGS` logic in the project but it didn't seem like it was working correctly, e.g. there was no `AC_SUBST()` for it. This hooks up that logic. I've also added some additional logic for common hardening options supported by GCC and Clang. I'm using some new AX_ macros that are copied from upstream as is usual. The flags will only be enabled on compilers that actually support these flags.

This also fixes some minor issues with how mu is using `strncpy()`, which was caught by GCC:

```
  CC       mu-contacts.lo
In function ‘encode_email_address’,
    inlined from ‘mu_contacts_add’ at mu-contacts.c:293:10:
mu-contacts.c:235:13: warning: ‘strncpy’ specified bound 255 equals destination size [-Wstringop-truncation]
  for (cur = strncpy(enc, addr, sizeof(enc)); *cur != '\0'; ++cur)
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The issue is that in calls like this, if `strlen(addr) == sizeof(enc)` then after this call `enc` will not be null-terminated, which is normally not expected.

There's also a warning I fixed related to a pedantic function signature mismatch.

With this these changes I can build the project cleanly (without warnings) on GCC 8.1.1

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -48,6 +48,14 @@ AX_CXX_COMPILE_STDCXX_14
 m4_ifdef([AX_COMPILER_FLAGS],[AX_COMPILER_FLAGS(,,[yes])])
 AX_VALGRIND_CHECK
 
+AX_CHECK_COMPILE_FLAG([-Wall], [AX_APPEND_COMPILE_FLAGS([-Wall], [WARN_CFLAGS])])
+AC_SUBST(WARN_CFLAGS)
+
+AX_CHECK_COMPILE_FLAG([-Wformat-security], [AX_APPEND_COMPILE_FLAGS([-Wformat-security], [HARDEN_CFLAGS])])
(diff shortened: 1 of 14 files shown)


yafiyogi:
These changes don't fix the problem. if `strlen(addr) >= sizeof(enc) - 1`, enc will not be null terminated. 
From the `strncpy()` man page:
       

> The strncpy() function is similar, except that at most n bytes  of  src
       are  copied.  Warning: If there is no null byte among the first n bytes
       of src, the string placed in dest will not be null terminated.

You need to add:
    `enc[sizeof(enc) - 1] = '\0';`
to ensure that enc is correctly null terminated.

There is a discussion of your blog post on Hacker News. See https://news.ycombinator.com/item?id=17248031

djcb:
I have fixed in a slightly different way now -- anyway, thanks for bringing this to my attention!

PR #1265, opened 2018-06-15
mu4e-view: enable hooks to run when the message is unread

seanfarley:
As far as I can tell, the mu4e-view hooks are not running when first opening an
message because of this conditional. I've tested this with a few custom hooks
and it seems to work but perhaps I'm missing something so feedback is
appreciated.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -329,15 +329,18 @@ article-mode."
 	  (buf
 	    (if embedded
 	      (mu4e~view-embedded-winbuf)
-	      (get-buffer-create mu4e~view-buffer-name)))
-         mode-enabled)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Your fix is ok, it works, the problem is that now the view buffer is printed twice and the hooks run twice.
I sent a new PR that fix this issue as well explaining exactly what is happening.
This will as well improve slow display of larges message that were displayed twice (+ the hooks running twice as well). Probably #1267 and this PR should be closed in favor of #1268 (perhaps I can add the full explanation which is in commit message as comment in the code @djcb ?)

seanfarley:
Oh, nice! Thanks for the fix! :-D

PR #1267, opened 2018-06-17
ensure mu4e-view-mode hooks run

jvshahid:
This is duplicate of #1265. I made this PR because I wasn't aware at the time of the PR #1265 and I'm not sure if the fix in that PR actually addresses the underlying problem. The problem seems to the nesting of `(run-mode-hooks 'mu4e-view-mode-hook)` inside ` (when (or embedded (not (mu4e~view-mark-as-read-maybe msg)))` and doesn't have anything to do with the condition `mode-enabled`

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -348,8 +348,8 @@ article-mode."
 	  (mu4e~fontify-signature)
 	  (mu4e~view-make-urls-clickable)
 	  (mu4e~view-show-images-maybe msg)
-	  (when embedded (local-set-key "q" 'kill-buffer-and-window))
-          (unless mode-enabled (run-mode-hooks 'mu4e-view-mode-hook)))))
+          (when embedded (local-set-key "q" 'kill-buffer-and-window)))
+        (unless mode-enabled (run-mode-hooks 'mu4e-view-mode-hook))))
     (switch-to-buffer buf)))
 
 (defun mu4e~view-gnus (msg)


seanfarley:
Hehe, no worries :-) I, too, am not sure of the correct solution. Just to explain the situation to others, the code path for opening a new (unread) message is:

1. call mu4e~view-internal
1. when it gets to `(setq mode-enabled (eq major-mode 'mu4e-view-mode)` at this point, `major-mode` is `fundamental-mode` so this evaluates to false
1. since the previous step is false, set a local variable `mu4e-view-mode-hook` to nil in a scope while calling `(mu4e-view-mode)` (so as to not run any hooks)
1. run a few more lines to get to `when (or embedded (not (mu4e~view-mark-as-read-maybe msg)))`
1. the `mu4e~view-mark-as-read-maybe` method will in turn call `mu4e~view-internal` recursively
   1. re-enter `mu4e~view-internal`
   1. when it gets to `(setq mode-enabled (eq major-mode 'mu4e-view-mode)` at this point, `major-mode` is now `mu4e-view-mode` so this evaluates to true (so `mode-enabled` is non-nil)
   1. run a few more lines to `when (or embedded (not (mu4e~view-mark-as-read-maybe msg)))` which will now pass as true
   1. enter the when statement and hit the line `(unless mode-enabled (run-mode-hooks 'mu4e-view-mode-hook))` which will not run since in this recursive evaluation we've already set that
   1. pop out of the recursive stack
1. `(mu4e~view-mark-as-read-maybe msg)` returned true (since we marked it as read in the recursive call)
1. so `(not (mu4e~view-mark-as-read-maybe msg))` will evaluate as false and we not enter the when statement
1. hence the hook method is not called :-(

There are two solutions to this: we could the hooks in the `(mu4e~view-mark-as-read-maybe msg)` call (which is what my PR does) or we could bring that hook call outside the `when` block (which is what this PR does). I suspect this PR is the correct one since it keeps the `(unless ...)` call and mine does not (meaning if that code path is hit another way it could trigger the hooks when they were not supposed to be triggered).

jvshahid:
Thanks @seanfarley for the explanation. It wasn't obvious that your pr is running the hooks in the recursive call. I think a comment could help clarify the intention. I also think that `mode-enabled` isn't needed anymore.

seanfarley:
Thierry Volpiatto <notifications@github.com> writes:

> Sean Farley <notifications@github.com> writes:
>
>> There are two solutions to this: we could the hooks in the
>> (mu4e~view-mark-as-read-maybe msg) call (which is what my PR does) or
>> we could bring that hook call outside the when block (which is what
>> this PR does). I suspect this PR is the correct one since it keeps the
>> (unless ...) call and mine does not (meaning if that code path is hit
>> another way it could trigger the hooks when they were not supposed to
>> be triggered).
>
> Didn't follow the whole thread, but here the hooks are running fine with
> actual code.

As I outlined above, I don't see how that's possible. To be clear, this
bug is only happening when opening an unread email and marking it read.
Opening an already email runs the hooks.

In fact, the behavior is exactly similar to:
https://groups.google.com/forum/#!topic/mu-discuss/gG7q3vHzwUY

> As a reminder:
>
> The intention of original code was to run the hook only when text was
> inserted, not before, this allowed for example 
>     (add-hook 'mu4e-view-mode-hook 'smiley-buffer)
> to treat smileys once the text was already inserted, otherwise
> 'smiley-buffer was run on nothing (empty buffer).
>
> See PR #1141 and #1148 for more infos.

I've gone over those pull requests (and the un-merge, re-merge) of them.
I don't see how hooks can ever run when `(mu4e~view-mark-as-read-maybe
msg)` will return nil the first time this email is opened (thereby not
entering the `(when (or embedded (not (mu4e~view-mark-as-read-maybe
msg)))` block; and in the recursive call, `(setq mode-enabled (eq
major-mode 'mu4e-view-mode)` is set to t, making the `(unless
mode-enabled (run-mode-hooks 'mu4e-view-mode-hook))` not run.

It'd be helpful if you could read what I wrote above and point out where
I went wrong so that we can get to the bottom of this bug.

seanfarley:
I just ran your [mu4e-config.el](https://github.com/thierryvolpiatto/emacs-tv-config/blob/master/mu4e-config.el) in a separate `emacs -Q` just to try and see how this works for you but I cannot reproduce your statement that this works. Evaluating an `edebug-defun` for any of the hook functions shows that nothing is run when opening an unread email for the first time (they run upon opening the message a second time).

thierryvolpiatto:

> I just ran your mu4e-config.el in a separate emacs -Q just to try and
> see how this works for you but I cannot reproduce your statement that
> this works. Evaluating an edebug-defun for any of the hook functions
> shows that nothing is run when opening an unread email for the first
> time (they run upon opening the message a second time).

You are right, as I don't follow the thread from start I didn't notice
the problem was when opening an unread email, and yes in this case the
hooks don't run.

#1267 fix (this PR) is not working for me wheras #1265 is working,
perhaps you can remove as well the mode-enabled binding which is no
more needed.
IIRC I was checking for mode-enabled to avoid running twice the hooks,
it seems this is no more needed.

Thanks.

--
Thierry

jvshahid:
Sorry for the poor explanation of the underlying problem that the PR is trying to address. I should have done better explaining the issue.

I'm not sure why this PR is not working for you, but I'm happy with either fixes. I think running the hooks as part of the recursive call is fine since that's where the msg text is being inserted anyway. @seanfarley can you please add a brief comment explaining the order of the calls.  May be something similar to your previous comment but shorter.

thierryvolpiatto:
John Shahid <notifications@github.com> writes:

> I'm not sure why this PR is not working for you, but I'm happy with
> either fixes. I think running the hooks as part of the recursive call
> is fine since that's where the msg text is being inserted
> anyway.

This is fine.
The complexity IIRC was that `mu4e~view-mark-as-read-maybe` was
involving another call of `mu4e-view-mode` and BTW of its hooks that
were indeed running because `mu4e-view-mode-hook` was this time not
let-bounded, so the original intention when I wrote this patch was to
not run twice the hooks.
However it seems this is not more true (correct me if i am wrong) so
there is no need to check if mode is enabled or not, we can run the
hooks inconditionally.

Hope that helps.

Thanks.

--
Thierry

seanfarley:
Ah, great, this is good news! I'll update my PR and add some of the highlights from this discussion to the commit message and relevant highlights to the code comments.

seanfarley:
Ok, I've updated #1265; please let me know if I missed anything!

thierryvolpiatto:

> Ok, I've updated #1265; please let me know if I missed anything!

Yes, see comment in #1265 and PR #1268.
Thanks!

-- 
Thierry

djcb:
Think this should work now. Thanks for the effort.

Issue #1217, opened 2018-04-01
Reduce hang in visiting next email


bradyt:
## Expected or desired behavior

Hold down the `n` key and flip quickly through emails.

This youtube video suggests that mutt can handle this [Using Mutt deleting spam in realtime - YouTube](https://www.youtube.com/watch?v=Vq6U38WHpsw). I've reproduced in mutt, with html rendering via w3m, and autoview for all html emails, and there is still no real noticeable hang.

## Actual behavior

For just a few key presses, there is a slightly noticeable amount of "Waiting for message..." before messages appear rendered.

When holding down the `n` key, there is also intermittent hanging before cursor in headers pane moves.

## Steps to reproduce

Occurs even for

```
(setq mu4e-view-prefer-html nil
      mu4e-html2text-command (lambda (msg) ""))
```

## Versions of mu, mu4e/emacs, operating system etc.

```
mu (mail indexer/searcher) version 1.0
GNU Emacs 25.3.1 (x86_64-apple-darwin17.4.0, NS appkit-1561.20 Version 10.13.3 (Build 17D102)) of 2018-03-30
```

djcb:
I don't notice any "hang"; how many millisecs are we talking?

bradyt:
I now think this is two issues, not sure if we should split this. Again, I have key repeat rate up, so not sure if you are taking that step.

# Issue 1

TL;DR: Part of my issue is found to be `evil-mode`'s `normal` state in `fundamental-mode`, for `Waiting for message...`.

When holding down the `n` key, part of the issue is associated with a message, from `*Messages*` buffer, `user-error: "mark": pattern not found`, then `M-x toggle-debug-on-error` doesn't quite catch that one until a suggestion on #emacs to try `(setq debug-ignored-errors (delq 'user-error debug-ignored-errors))`. Then I note stack trace has `evil-search-next`, ponder how that's being called, and notice the `Waiting for message...` buffer is in `fundamental-mode` with `evil-mode`'s `normal` state.

Maybe the `Waiting for message...` buffer should already be in `mu4e:view` mode? This is what you see when starting `gnus`, but then, that start up page is not something we are flipping through. If I look at how `gnus` traverses articles, it doesn't move cursor in headers buffer until view has updated. Maybe that sort of hang is not desired. I'm not sure what `mutt` does, I see flashing in their version of echo area.

# Issue 2

[...]

Issue #1192, opened 2018-02-13
mu4e-view-mode-hooks not run on unread messages


malb:
## Expected or desired behavior

Hook is run on unread messages same as on read messages.

## Actual behavior

Hook is never run on unread messages.

## Steps to reproduce

Add a `mu4e-view-mode-hook` and observe it is not being run.

## Versions of mu, mu4e/emacs, operating system etc.

mu 1.0, Emacs 25.1

## Any other detail

The following adaptation fixes the issue for me (see line with malb)

```lisp
(defun mu4e-view (msg)
  "Display the message MSG in a new buffer, and keep in sync with HDRSBUF.
'In sync' here means that moving to the next/previous message in
the the message view affects HDRSBUF, as does marking etc.

As a side-effect, a message that is being viewed loses its 'unread'
marking if it still had that."
  (let* ((embedded ;; is it as an embedded msg (ie. message/rfc822 att)?
          (when (gethash (mu4e-message-field msg :path)
                         mu4e~path-parent-docid-map) t))
         (buf
          (if embedded
	      (mu4e~view-embedded-winbuf)
            (get-buffer-create mu4e~view-buffer-name)))
         mode-enabled)
    (with-current-buffer buf
      (unless (setq mode-enabled (eq major-mode 'mu4e-view-mode))
        (let (mu4e-view-mode-hook) (mu4e-view-mode)))
      (setq mu4e~view-msg msg)
      ;; When MSG is unread, mu4e~view-mark-as-read-maybe will trigger
      ;; another call to mu4e-view (via mu4e~headers-update-handler as
      ;; the reply handler to mu4e~proc-move)
      (let ((inhibit-read-only t))
        (when (or embedded (not (mu4e~view-mark-as-read-maybe msg)))
	  (erase-buffer)
	  (mu4e~delete-all-overlays)
	  (insert (mu4e-view-message-text msg))
	  (goto-char (point-min))
	  (mu4e~fontify-cited)
	  (mu4e~fontify-signature)
	  (mu4e~view-make-urls-clickable)
	  (mu4e~view-show-images-maybe msg)
	  (when embedded (local-set-key "q" 'kill-buffer-and-window))
          ;; malb: (unless mode-enabled (run-mode-hooks 'mu4e-view-mode-hook))
          (run-mode-hooks 'mu4e-view-mode-hook)
          )))
    (switch-to-buffer buf)))
```

djcb:
Fixed.

PR #1268, opened 2018-06-19
Fix mu4e-view-mode-hook not called on unread messages

thierryvolpiatto:
Fix #1265 and #1267.

Here's the problem:

When `mu4e~view-internal` is called it makes a call to
`mu4e~view-mark-as-read-maybe`.
When the message is unread `mu4e~view-mark-as-read-maybe` calls
`mu4e~proc-move` which flag the message as read and calls again
`mu4e~view-internal` which in turn calls again
`mu4e~view-mark-as-read-maybe` which this time returns nil and
`mu4e~view-internal` continues printing the view buffer and running
hooks.
Then `mu4e~view-mark-as-read-maybe` first call continue to run and
calls now `mu4e~proc-view` which as `mu4e~proc-move` calls again
`mu4e~view-internal` and prints a second time the buffer and run a
second time the hooks.

The first thing this patch does is always running hooks after printing
message, then to avoid doing this again it advice temporarily
`mu4e-view-func` aka `mu4e~view-internal` to prevent printing and
running hooks again.
Finally this advice function remove itself the advice to allow
printing the next message.

* mu4e/mu4e-view.el (mu4e~view-internal): Remove mode-enabled checking
which is useless.
(mu4e~view-advice-view-func): Advice mu4e-view-func to ensure it
doesn't run a second time.
(mu4e~view-mark-as-read-maybe): Use it.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -329,10 +329,9 @@ article-mode."
 	  (buf
 	    (if embedded
 	      (mu4e~view-embedded-winbuf)
-	      (get-buffer-create mu4e~view-buffer-name)))
-         mode-enabled)
+	      (get-buffer-create mu4e~view-buffer-name))))
     (with-current-buffer buf
-      (unless (setq mode-enabled (eq major-mode 'mu4e-view-mode))
+      (unless (eq major-mode 'mu4e-view-mode)
         (let (mu4e-view-mode-hook) (mu4e-view-mode)))
       (setq mu4e~view-msg msg)
       ;; When MSG is unread, mu4e~view-mark-as-read-maybe will trigger
@@ -349,7 +348,7 @@ article-mode."
 	  (mu4e~view-make-urls-clickable)
 	  (mu4e~view-show-images-maybe msg)
 	  (when embedded (local-set-key "q" 'kill-buffer-and-window))
-          (unless mode-enabled (run-mode-hooks 'mu4e-view-mode-hook)))))
+          (run-mode-hooks 'mu4e-view-mode-hook))))
     (switch-to-buffer buf)))
 
 (defun mu4e~view-gnus (msg)
@@ -903,8 +902,13 @@ changes, it triggers a refresh."
 	;; mark /all/ messages with this message-id as read, so all copies of
 	;; this message will be marked as read.
 	(mu4e~proc-move msgid nil "+S-u-N")
+        (add-function :override mu4e-view-func 'mu4e~view-advice-view-func)
         (mu4e~proc-view docid mu4e-view-show-images (mu4e~decrypt-p msg))
 	t))))
+
+(defun mu4e~view-advice-view-func (&rest args)
+  (apply 'ignore args)
+  (remove-function mu4e-view-func 'mu4e~view-advice-view-func))
 
 (defun mu4e~view-browse-url-func (url)
   "Return a function that executes `browse-url' with URL.
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -335,8 +335,13 @@ article-mode."
         (let (mu4e-view-mode-hook) (mu4e-view-mode)))
       (setq mu4e~view-msg msg)
       ;; When MSG is unread, mu4e~view-mark-as-read-maybe will trigger
-      ;; another call to mu4e-view (via mu4e~headers-update-handler as
-      ;; the reply handler to mu4e~proc-move)
+      ;; 2 other calls to mu4e~view-internal when running
+      ;; mu4e~proc-move and mu4e~proc-view. To avoid printing and
+      ;; running hooks a second time we advice mu4e-view-func aka
+      ;; mu4e~headers-view-handler aka mu4e-view aka
+      ;; mu4e~view-internal so that it does nothing. See
+      ;; mu4e~view-mark-as-read-maybe and mu4e~view-advice-view-func.
+      ;; Ref Issues #1265 #1267 #1268.
       (let ((inhibit-read-only t))
         (when (or embedded (not (mu4e~view-mark-as-read-maybe msg)))
 	  (erase-buffer)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
This PR fix as well #1192 and possibly #1217

thierryvolpiatto:
@djcb ping

djcb:
Hmmm, using `advise` here seems a bit unnecessary since we control the code ourselves, or?

 I'd prefer a solution where  `mu4e~view-mark-as-read-maybe` returns `t` if actually marks the message, and we can do something like: 
```
(unless (mu4e~view-mark-as-read-maybe) 
       ... view the message, run hooks etc.)
```
then, we also run the hooks on the message the user actually sees (rather than still-unread one)

thierryvolpiatto:

> Hmmm, using advise here seems a bit unnecessary since we control the
> code ourselves, or?

Did you read the commit message and comments?

      ;; When MSG is unread, mu4e~view-mark-as-read-maybe will trigger
      ;; 2 other calls to mu4e~view-internal when running
      ;; mu4e~proc-move and mu4e~proc-view. To avoid printing and
      ;; running hooks a second time we advice mu4e-view-func aka
      ;; mu4e~headers-view-handler aka mu4e-view aka
      ;; mu4e~view-internal so that it does nothing. See
      ;; mu4e~view-mark-as-read-maybe and mu4e~view-advice-view-func.

So if you don't advise `mu4e-view-func`, `mu4e~proc-view` will print a
second time the buffer and also run hook a second time.

> I'd prefer a solution where mu4e~view-mark-as-read-maybe returns t if
> actually marks the message, and we can do something like:

Of course it would be better but it is not possible unless you change
the signature of some functions to accept an extra optional arg that say
"buffer is printed and hook already ran so do nothing now".

In this patch we are using `add-function` which is exactly the tool
needed for such cases, it affects nothing else as the advice function
remove itself when it run and finally avoid doing complex changes in
sensible function that may break something else.

Thanks.

--
Thierry

thierryvolpiatto:
```
mu4e~view-internal
    => mu4e~view-mark-as-read-maybe
       => mu4e~proc-move
          => mu4e~view-internal
             => mu4e~view-mark-as-read-maybe -> nil (print buffer and run hooks)
                => mu4e~proc-view
                   => mu4e~view-internal
                      => mu4e~view-mark-as-read-maybe -> nil (print buffer and run hooks)[2th]
       => t
```

seanfarley:
Gentle ping

djcb:
When setting `mu4e-view-mode-hook` to something like
```
(add-hook 'mu4e-view-mode-hook
  (lambda()
    (message "running mu4e-view-mode-hook")))
```
I see that message only exactly once for either read or unread messages.. so under what circumstances does this fail?

And indeed, I would prefer changing functions to do the right things, I'd consider advice only for code one cannot otherwise change.

seanfarley:
I don't see how that's possible and even tested it just now. For me, your hook doesn't run on unread messages.

I am deterred from helping out any more since you seem so flippant in your response. There is much discussion in your own project on this behavior that happened *two months ago*:

* https://github.com/djcb/mu/pull/1265
* https://github.com/djcb/mu/pull/1267

I'll continue to monkey patch and fix this myself since you don't seem to have the time.

thierryvolpiatto:

> When setting mu4e-view-mode-hook to something like
>
> (add-hook 'mu4e-view-mode-hook
>   (lambda()
>     (message "running mu4e-view-mode-hook")))
>
> I see that message only exactly once for either read or unread
> messages..

Hmm...

> so under what circumstances does this fail?

Unread messages, read the commit message of my patch, it is all explained.

> And indeed, I would prefer changing functions to do the right things,

If you want to do this, consider rewriting a bunch of things in
mu4e-proc...

> I'd consider advice only for code one cannot otherwise change.

This is not advice, well not really, it is `add-function` which have
been conceived to interact from inside source code, in contrast with
`advice-add` which is to interact with source code from outside
i.e. user side.

-- 
Thierry

thierryvolpiatto:

> I don't see how that's possible and even tested it just now. For me,
> your hook doesn't run on unread messages.

Agree.

> I'll continue to monkey patch and fix this myself since you don't seem to have the time.

Same here.

-- 
Thierry

djcb:
"flippant"? Let's assume good faith, shall we?

djcb:
So, I think this should be fixed now (even if in a slightly different way). Thanks for taking the time.

thierryvolpiatto:
I have updated mu and @djcb FYI your fix is still not working.
Only https://github.com/thierryvolpiatto/mu/tree/mu4e-view-mode-hook_fix is working for me for now.
IIRC What you did was good but insufficient, you need to merge your changes with my branch (fully not tested).

djcb:
Hmmm... it works for me. Ie.,
```
(setq mu4e-view-use-gnus nil)
(add-hook 'mu4e-view-mode-hook
  (lambda() (message "running mu4e-view-mode-hook")))
```

And I can see the `message` for each message I open (unread or not). Do you see something else?

thierryvolpiatto:

> Hmmm... it works for me. Ie.,
>
> (setq mu4e-view-use-gnus nil)
> (add-hook 'mu4e-view-mode-hook
>   (lambda() (message "running mu4e-view-mode-hook")))
>
> And I can see the message for each message I open (unread or not). Do you see something else?

Yes smileys are not activated with read or unread messages.
(add-hook 'mu4e-view-mode-hook 'smiley-buffer)

-- 
Thierry

djcb:
So, the hook does run, but it runs a bit too early for smiley-mode to anything useful. I've pushed a change so the hooks run later, and that seems to work. For you as well, @thierryvolpiatto ?

thierryvolpiatto:

> So, the hook does run, but it runs a bit too early for smiley-mode to
> anything useful. I've pushed a change so the hooks run later, and that
> seems to work.  For you as well, @thierryvolpiatto ?

Seems to work fine now, thanks ;-).

--
Thierry

tasmo:
> Hmmm... it works for me. Ie.,
> 
> ```
> (setq mu4e-view-use-gnus nil)
> (add-hook 'mu4e-view-mode-hook
>   (lambda() (message "running mu4e-view-mode-hook")))
> ```
> 
> And I can see the `message` for each message I open (unread or not). Do you see something else?

Unfortunately it does not work here. The message appears in read emails but not when I open unread.

PR #1278, opened 2018-06-30
mu4e: extend "reply to all/sender only" with "reply to mailing list"

mbunkus:
At the moment it isn't possible to reply solely to a mailing list with mu4e. I can only reply to all and remove the non-mailing list addresses manually.

This PR adds a third mode: "reply to list". It will use the `List-Post` header if present; otherwise it'll use the first address in the `To` header (this is a bit of a heuristic as well as a way to make "reply to list" always do something).

### DIFF
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -325,20 +325,30 @@ never hits the disk. Also see `mu4e~draft-insert-mail-header-separator."
       (let ((inhibit-read-only t))
 	(replace-match "")))))
 
-(defun mu4e~draft-reply-all-p (origmsg)
-  "Ask user whether she wants to reply to *all* recipients.
+(defun mu4e~draft-user-wants-reply-all (origmsg)
+  "Ask user whether she wants to reply to *all* recipients, just the sender or
+the mailing list address (the latter only if a `List-Post' header is found).
 If there is just one recipient of ORIGMSG do nothing."
-  (let* ((recipnum
-	   (+ (length (mu4e~draft-create-to-lst origmsg))
+  (let* ((cc-reps (mu4e~draft-create-cc-lst origmsg t))
+	 (recipnum
+	  (+ (length (mu4e~draft-create-to-lst origmsg))
 	     (length (mu4e~draft-create-cc-lst origmsg t))))
-	  (response
-	    (if (< recipnum 2)
-	      'all ;; with less than 2 recipients, we can reply to 'all'
+	 (response
+	  (let (list-address (origmsg-as-string ""))
+	    (with-temp-buffer
+	      (insert-file-contents (plist-get origmsg :path))
+	      (setq origmsg-as-string (buffer-string)))
+	    (if (string-match "^List-Post: *\\(?:<mailto:\\)?\\([^>[:space:]]+\\)" origmsg-as-string)
+	        (setq list-address (match-string 1 origmsg-as-string)))
+	    (if (and (= recipnum 1) (not list-address))
+	        'all ;; with one recipient, we can reply to 'all'....
 	      (mu4e-read-option
-		"Reply to "
-		`( (,(format "all %d recipients" recipnum) . all)
-		   ("sender only" . sender-only))))))
-    (eq response 'all)))
+	       "Reply to "
+	       `( (,(format "all %d recipients" recipnum) . all)
+		  ,(if (or list-address cc-reps)
+		       `("list only" . ,(or list-address (cdar cc-reps))))
+		  ("sender only" . sender-only)))))))
+    response))
 
 (defun mu4e~draft-message-filename-construct (&optional flagstr)
   "Construct a randomized name for a message file with flags FLAGSTR.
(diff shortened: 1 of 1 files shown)


djcb:
Hmm, I simply use `R` (and 'sender only') which seems to work fine for all the mailing-lists I'm using. Doesn't that work for you?

mbunkus:
No, it doesn't, at least not for lists who don't set `Reply-To` to the list's address. Let'stake the IETF's CELLAR list (standard mailman). `From` is the original sender's address (not the mailing list address), `To` is the mailing list address, `Reply-To` isn't set, `List-Post` is set. In that case:

* "sender only" uses what's in `From`, meaning the person who sent the mail
* "all" uses the original `From` (the person) as the new `To` and the original `To` (the list address) becomes the new `Cc` requiring me to remove the new `To` and move the new `Cc` to the new `To`.

I don't have a list around that sets `Reply-To` to the list's address in order to test how that works. I really want an option that implements "definitely send to the list, no matter who else is copied and no matter how the list is configured" as this minimizes the mental work I have to do. I don't have to think about "does this list set `Reply-To` and would I therefore get away with 'reply to sender'", and I don't have to use "reply to all" with the additional manual steps of removing the additional person addresses and moving the list address from `Cc` to `To`.

I've been using these functions for well over a year now. I just thought others might be happy to have such functionality as it was requested [over on Reddit](https://www.reddit.com/r/emacs/comments/8uxq59/mu4e_reply_to_list/).

djcb:
I think the it would be better to set `reply-to` to the mailing-list address then, for such messages. The patch seems to do some manual re-parsing in mu4e-draft; that's not really something I want to do.

mbunkus:
> I think the it would be better to set reply-to to the mailing-list address then, for such messages.

Uhm, the `Reply-To` header is not under my control as a recipient of the mailing list message. This function is about enabling the user to do whatever she thinks fits best no matter how other people have configured their software (mailing list software, email client).

> The patch seems to do some manual re-parsing in mu4e-draft; that's not really something I want to do.

Fair enough. The only reason I did that is that there didn't seem to be any other way to get the `List-Post` header as it isn't part of the `origmsg` structure. If you have a better idea how to pass the header around, I can look into implementing it.

Look, I won't be offended if you don't want to incorporate such a function. It's just something that a lot of other power-user mail clients (mutt, Thunderbird, kmail) have had for ages and that some people really expect to be there.

djcb:
What I meant is that perhaps mu should set the `reply-to` flag to the mailing-list address for such messages; then mu4e would DTRT in a more cases.

djcb:
Do you have an example message that's not handled correctly at the moment?

mbunkus:
> Do you have an example message that's not handled correctly at the moment?

Sure, [here you go](https://www.bunkus.org/misc/mu4e-reply-behavior-1.eml).

With that message reply to "all" puts Steve Lhomms address into `To` and the list's address into `Cc. With reply to "sender only" it puts Steve's address into `To` and nothing into `Cc`. What I often want is to only reply to the list (lists's address in `To`, nothing in `Cc`), and current mu4e simply doesn't offer a way to do that safe for replying to "all" and manually fixing the headers — which is annoying and error-prone.

djcb:
I've changed it so that mu sets the `reply-to` for such messages; they now behave just like any other mailing-list replies.

dominikh:
The issue with mailing lists that set the Reply-To header, as well as mu4e doing the same, is that it may become impossible to respond to the sender of the email. Reply-To exists to allow the sender to specify an address that is different from the one used in From. If we overwrite the header, we lose information.

See https://www.unicom.com/pw/reply-to-harmful.html for more opinions on mailing lists and Reply-To.

bitclick:
I think i am suffering from this too (mu 1.2.0): I cannot reply to the sender only anymore. the fake reply-to header sneaks in and now i accidentally replied to the whole list several times!  please change this or at least tell me in the ui choice labels..

i would like a 3 choices solution very much :-)

brandoninvergo:
> I've changed it so that mu sets the `reply-to` for such messages; they now behave just like any other mailing-list replies.

This change breaks a standard usage of mailing lists. For example, you can have mailing lists within an organization where only a few people are subscribed (say, a committee). Other people inside or outside the organization email the list to get in touch with the committee. When members of the committee respond, they want to respond to the original inquiry while still CC'ing the list to keep a record of the discussion.

As an example, this breaks standard usage of `report-emacs-bug` from the perspective of the Emacs developers (I am not an Emacs dev). It is assumed that the Emacs user reporting the bug is not subscribed to bug-gnu-emacs@gnu.org, yet the user would expect to receive replies from the subscribers of the list receiving the report (namely, from the devs who would handle the bug report). If an Emacs dev is using mu4e to reply to the bug report, say to ask for more info, they would have to manually copy the original reporter's email address and paste it into their reply's `To:` field while moving the list to the `Cc:` field. This is massively inconvenient.

djcb:
I've update the implementation to do some special-handling for mailing-lists; this is somewhat inspired by the PR here (thanks for that!) but quite different as well. Anyway, it should hopefully now work to anyone's satisfaction; feel free to leave any comments in #1459.

PR #1284, opened 2018-07-09
Perform threading calculation on related set instead of entire result.

nicolasavru:
The current threading algorithm is applied to the entire result of a query, even
if maxnum is specified, and then the result of the threading algorithm is
truncated to maxnum. The improves threading results by returning the entire
thread even when only a single message makes it into the top maxnum results.

This commit applies the threading algorithm to the related message set of the
maxnum-truncated query result instead of to the entire query result. For a given
set of messages, the set of messages which will share threads with any of the
original messages is exactly the related message sets. Put another way, either
any messages returned by the original query but removed by the maxnum truncation
will also be returned by the related message query, or they would not have been
needed anyway because they would not be members of any visible thread.

To maintain backward compatibility and allow threading to be used without
including related messages, the related message set is found for the threading
calculation, but any messages which would not have matched the original query
are then pruned, resulting in a superset of the truncated query, but a subset of
the untruncated query.

This does not improve (or degrade) the run time of a threading calculation when
maxnum is not set, but significant improves it when maxnum is set by making it
scale (roughly) linearly in terms of maxnum. On a maildir with ~200k messages
and maxnum set to 500 (the default), the run time of a threading calculation is
lowered from ~1m to ~0.1s.

### DIFF
--- a/lib/mu-query.cc
+++ b/lib/mu-query.cc
@@ -369,15 +369,23 @@ get_related_query (MuMsgIter *iter, GHashTable **orig_set)
 
 
 static void
-include_related (MuQuery *self, MuMsgIter **iter, int maxnum,
-		 MuMsgFieldId sortfieldid, MuQueryFlags flags)
+get_related_messages (MuQuery *self, MuMsgIter **iter, int maxnum,
+                      MuMsgFieldId sortfieldid, MuQueryFlags flags,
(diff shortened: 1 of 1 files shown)


djcb:
Great work; merging. There's may be some further feedback (after use), stay tuned!

PR #1286, opened 2018-07-12
mu4e: fix action to show thread for single-window mode

ericdanan:
When in single-window mode and invoking `mu4e-action-show-thread` from
the view buffer, stay in the headers buffer rather than going back to
the view buffer.


djcb:
Merged -- thanks.

PR #1290, opened 2018-07-13
org-mu4e: move mu4e requirement inside org-mu4e-open

ericdanan:
This allows to defer loading mu4e. That is, if we simply put in our
init file:

```
(with-eval-after-load 'org
  (require 'org-mu4e))
```
then mu4e is only loaded if either we follow a mu4e link from an org
buffer or we call the autoloaded function `mu4e` to read our emails.

Storing org-mode links from mu4e buffers is unaffected by this change
since `org-mu4e-store-link` only applies from mu4e buffers.

Regarding the second, experimental part of org-mu4e.el about editing
mails with org-mode, I am not completely sure but I think it should
also be unaffected since it also seems to only apply from mu4e buffers.

### DIFF
--- a/mu4e/org-mu4e.el
+++ b/mu4e/org-mu4e.el
@@ -38,7 +38,6 @@
 (require 'org-exp nil 'noerror)
 
 (eval-when-compile (require 'cl))
-(require 'mu4e)
 
 (defgroup org-mu4e nil
   "Settings for the org-mode related functionality in mu4e."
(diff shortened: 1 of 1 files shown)


ericdanan:
Also, I noted org-mu4e depends on library org-exp, but I can't find this library in org (nor org-plus-contrib). Internet search only suggests an old library that was part of emacs as of 2012. Should the dependency be removed?

djcb:
Thanks - and I have removed org-exp.

PR #1291, opened 2018-07-13
Add custom variable mu4e-hide-main-view-buffer

ericdanan:
If non-nil, mu4e's main view buffer will be replaced with a minibuffer
prompt.

The function `mu4e-main-view` is modified to rely on this variable
rather than `mu4e-split-view` (currently, the minibuffer is used if
and only if `mu4e-split-view` is set to `'single-window`. Thus one can
now use single-window mode together with the main view buffer (useful
e.g. with the mu4e-maildirs-extension package) or another view mode
together with the minibuffer menu.

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -198,7 +198,7 @@ clicked."
 
 (defun mu4e~main-view ()
   "Create the mu4e main-view, and switch to it."
-  (if (eq mu4e-split-view 'single-window)
+  (if mu4e-hide-main-view-buffer
       (if (buffer-live-p (mu4e-get-headers-buffer))
 	  (switch-to-buffer (mu4e-get-headers-buffer))
 	(mu4e~main-menu))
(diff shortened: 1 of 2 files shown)


mu4e/mu4e-main.el
```diff
@@ -198,7 +198,7 @@ clicked."
 
 (defun mu4e~main-view ()
   "Create the mu4e main-view, and switch to it."
-  (if (eq mu4e-split-view 'single-window)
+  (if mu4e-hide-main-view-buffer
       (if (buffer-live-p (mu4e-get-headers-buffer))
```
djcb: Seems like this breaks the existing behavior, no?

ericdanan: Indeed, with `mu4e-split-view` set to `single-window`, the default behavior is now to display the main view buffer. To recover the former default behavior of displaying the main menu instead, `mu4e-hide-main-view-buffer` must be set to `t`. For other values of `mu4e-split-view`, the default behavoir is unchanged.

I'm not sure how to make the main view / main menu choice independent of the split view mode without breaking any default behavior. One possibility would be to make the default value of `mu4e-hide-main-view-buffer` dependent on the value of `mu4e-split-view`, but that wouldn't work for users who set `mu4e-split-view` after loading mu4e. Or perhaps we could only offer the choice for single-window mode (e.g. adding a variable `mu4e-single-window-hide-main-view` defaulting to `t`) and assume users of other split-view modes want the main view buffer.

djcb:
Seems this has gotten stale (sorry!)

PR #1303, opened 2018-08-18
view-gnus: Add mu4e headers before attachments

Chris00:
Fixes https://github.com/djcb/mu/issues/1269

PR #1305, opened 2018-08-18
view-gnus: Fix body decoding problems

Chris00:
Example:
```
To: Sender <sender@gmail.com>
From: =?utf-8?Q?C=C3=A9line?= <XXX@gmail.com>
Subject: Body UTF-8 + flowed
Date: Tue, 12 Jun 2018 09:05:26 +0200
Content-Type: text/plain; charset="utf-8"; format=flowed
Content-Transfer-Encoding: 8bit
Content-Language: fr-classic

répondre 
```


Chris00:
> I've found that it can be easy to fix some messages, yet breaking some others, so perhaps you can attach a (raw) message where this fixes things?

If you have some messages that are not fixed by this, could you please post them here?  I tried on messages from various sources (base64 encoded, flowed, HTML¹,...) and so far it seems to work well.

¹ I have `(setq mu4e-view-prefer-html t)`.

Chris00:
Here is an example of message that does not work with the above fix:
```
Content-Transfer-Encoding: 8bit
From: test <test@gmail.com>
Date: Mon, 20 Aug 2018 10:05:08 +0200
Content-Type: text/html; charset="utf-8"
Subject: test

<html><body>
Afficher  Documents partagés
</body>
</html>
```
I am actually surprised that `charset="utf-8"` is not reflected anywhere in the `msg` structure.  This is a case where `(mm-disable-multibyte)` should be called because of this.  What is a proper test to know whether a charset was specified for the message?

Chris00:
It is necessary for messages such as
```
Subject: test
To: XX <XX@domain>
Date: Fri, 17 Aug 2018 18:18:07 +0200
Content-Type: text/plain; charset="utf-8"; format=flowed
Content-Transfer-Encoding: 8bit
Content-Language: fr-BE

Je préfère
```
Moreover messages such as
```
Content-Transfer-Encoding: 8bit
From: test <test@gmail.com>
Date: Mon, 20 Aug 2018 10:05:08 +0200
Content-Type: text/html; charset="utf-8"
Subject: test

<html><body>
Afficher  Documents partagés
</body>
</html>
```
are not decoded properly without `(mm-disable-multibyte)`: the body decodes as _Afficher Documents partagis_ while it should read _Afficher Documents partagés_.  The above condition `(mu4e-message-field msg :body-txt-params)` is not enough for that — in my repo I added a *or* condition with `(mu4e-message-field msg :body-html)` but it is still a hack.  Hopefully, you understand better than me what is going on...

djcb:
Thanks. 

What I like to avoid though, is any use of `msg` for the gnus-version. Gnus is parsing that message, so it should be able to handle any decoding without anything that mu provides. We just have to find out how to do that...

Chris00:
It seems that `(mm-disable-multibyte)` can always be used — I thought I had a message that was not working but I cannot find it — *except* in the draft folder (because messages do not have a charset I guess).  If
```elisp
      (unless (string= (mu4e-message-field msg :maildir) mu4e-drafts-folder)
        (mm-disable-multibyte))
```
is not OK with you, one may first try to insert the message with `(mm-disable-multibyte)` and, if it turns out that it does not have a charset, reinsert it without disabling multibyte.

Chris00:
The latter solution may also be useful for broken messages that do not specify charsets...

Chris00:
The only thing that sometimes occurs whit this PR is that, after replying to a message, the `*Article*` buffer is redisplayed with incorrect decoding of accented letters.  Quitting the buffer and displaying it again performs the decoding properly.  For the particular message I just encounter:
```
Content-Type: text/html; charset="iso-8859-1"
Content-Transfer-Encoding: quoted-printable
```

Chris00:
@djcb Any additional comments on this PR?

djcb:
Thanks finally, had some time! It seems only the seconds commit makes a difference now, and it indeed fixes some of the problem cases. Great!

So I have installed that fix. The Maildir: addition... hmm, I'd like something a bit more generic (so people can add other headers). Have to think a bit about it.

So am going to close this one for now... but your fix is in, thanks again!

Chris00:
> The Maildir: addition... hmm, I'd like something a bit more generic (so people can add other headers). Have to think a bit about it.

OK.  If I can help, let me know.  (I prefer not to work on [another PR](https://github.com/djcb/mu/pull/1303) without a sense of direction.)

PR #1307, opened 2018-08-19
mu4e: fix mu4e-select-other-view for single-window mode

ericdanan:
In single-window mode, switch between the headers and view buffers.


vsedach:
My fault, I did not test mu4e-select-other-view when making single-window mode (it never occurred to me, as I consider redundant, and avoid, buffer-switching and window management provided by application packages, which was the motivation for single-window mode).

djcb:
Thanks for the effort.. this doesn't apply anymore, so am going to close this. Sorry!

PR #1310, opened 2018-08-23
org-mu4e: run hook before opening link

ericdanan:
The hook can be used in particular to open links in another window or
frame, e.g.:
```el
(add-hook 'org-mu4e-open-pre-hook (lambda () (other-window 1)))
```


djcb:
Org-mu4e has been obsolete for a long time, so no more updates, sorry...

PR #1311, opened 2018-08-26
mu4e: Don't descend in maildirs when recursively getting maildirs

juergenhoetzel:
maildirs are already the leaf nodes in the folder structure.
Also reduce number of intermediate (garbage collected) values.

I wondered why `[j]ump to some maildir` was a bit laggy on my system (I have a fairly large maildir structure).  This commit improves performance from:

```
ELISP> (benchmark 100 '(mu4e~get-maildirs-1 "~/.mail/gmail/" "/"))
"Elapsed time: 52.475803s (12.332123s in 100 GCs)"
```
to

```
ELISP> (benchmark 100 '(mu4e~get-maildirs-1 "~/.mail/gmail/" "/"))
"Elapsed time: 0.237148s (0.107507s in 1 GCs)"
```

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -323,15 +323,14 @@ Function will return the cdr of the list element."
 	     (directory-files-and-attributes
 	       (concat path mdir) nil
 	       "^[^.]\\|\\.[^.][^.]" t))))
-    (dolist (dentry dentries)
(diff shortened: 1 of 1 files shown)


djcb:
The way it's done is explicitly to allow for deeper folders, which many people do have (e.g. "Maildir/account1/folder1" etc.) The speed-up you get is quite impressive though, bet that function can be optimized...

PR #1320, opened 2018-09-18
mu4e: Replace cl with cl-lib

jabranham:
Emacs 24.3 obsoletes the old `cl` library in favor of the namespace-clean `cl-lib` library (and adds some byte compile warnings about using cl at runtime). This PR removes most of the `cl` requires in favor of `cl-lib` since mu4e now requires Emacs 24.4+. 

`cl` is still required at compile time for `lexical-let` in a few files.


djcb:
Hmm, I'm okay with using the new library, but I don't really like `cl-` prefixes for all calls... can you make a version without those?

jabranham:
Moving everything into the `cl-` pseudo namespace was one reason for the cl-lib library's creation. That's why the byte compiler warns about requiring `cl` at runtime; it's worried about namespace conflicts.

jabranham:
@djcb have you had any time to think about this more?

toogley:
@djcb you can hide the namespace with https://github.com/Malabarba/names or https://github.com/Malabarba/Nameless

djcb:
Yeah, let's take it. Thanks!

Issue #947, opened 2016-11-04
mu4e org-mode links need to be updated for org version 9.0


jkitchin:
Links in org 9.0 are defined differently than they were before. Here is how the mu4e link needs to be defined in org-9.0. The old links still work when followed, but without this, you cannot store a link with C-c C-l.

```
(org-link-set-parameters
 "mu4e"
 :follow 'org-mu4e-open
 :store 'org-mu4e-store-link)
```

neurolit:
I think it was solved by b885dd308eaf5255514ff6f3c09d5ef894696398. However homebrew package isn't up-to-date.

Try to evaluate this definition (added in b885dd308eaf5255514ff6f3c09d5ef894696398) in a buffer, and try again with C-c l. It should work.

```emacs-lisp
(if (fboundp 'org-link-set-parameters)
    (org-link-set-parameters "mu4e"
			     :follow #'org-mu4e-open
			     :store #'org-mu4e-store-link)
  (org-add-link-type "mu4e" 'org-mu4e-open)
  (add-hook 'org-store-link-functions 'org-mu4e-store-link))
```

neurolit:
It was merged into master at the end of September (9fd6d3696323b9e9d4df40cee0a51b4eef82d1dc).

jkitchin:
cool thanks. I clearly wasn't up to date on mu4e! I like that solution too for the version dependence. I will use something like it in org-ref. Thanks for the quick reply.

PR #1322, opened 2018-09-21
org-mu4e: fix org-mu4e-store-and-capture for Org 9.0

yiufung:
#920 fix links in Org 9 (Related discussion: #947). At the same time, `org-mu4e-store-and-capture` is still using `org-mu4e-store-link` only, so it doesn't work in Org 9. Fix by interactively calling `org-store-link` instead.

### DIFF
--- a/mu4e/org-mu4e.el
+++ b/mu4e/org-mu4e.el
@@ -138,7 +138,7 @@ the query (for paths starting with 'query:')."
 `org-mu4e-link-query-in-headers-mode', and capture it with
 org-mode)."
   (interactive)
-  (org-mu4e-store-link)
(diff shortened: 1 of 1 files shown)


djcb:
Does this still work with org 8.x?

yiufung:
I tried to reproduce it with Org 8, seems fine to me. Couldn't find Org 8 source code on official site, so I used [John Wiegley's old fork](https://github.com/jwiegley/org-mode), which is of version 8.2.3c. The init file is [here](https://gist.github.com/yiufung/c96a1a1ee147c6d019ce7751a0acac2e). 

mu4e-org-comparison

djcb:
Okay, merged. Thanks!

PR #1328, opened 2018-10-13
mu4e: restrict deleting trailing whitespace to region

pitti:
If no region is selected, mu4e-fill-paragraph will delete trailing whitespace on
all lines in the buffer.  This might not be desired, e.g. when adding a
signature to the mail which is delimited by the line '-- '.


djcb:
Okay, that's true... let's merge this. Merged --thanks!

PR #1334, opened 2018-10-31
doc: mu-query: add unseen flag and sort flags

ian-kelling:


djcb:
Thanks -- but note that `unread` (not 'unseen') is a synonym for `flag:new OR NOT flag:seen`.

djcb:
Updated the manpage -- thanks!

PR #1336, opened 2018-11-05
Fix missing continuation in tests/Makefile.am

wavexx:
Fix tests/Makefile.am not expanding properly.
Probably caused by the recent asan changes.


djcb:
Thanks -- fixed.

PR #1338, opened 2018-11-11
mu4e: allow :query to accept a function returning a query string

ibizaman:
This allows one to dynamically generate the query string, based for example on the value of a variable.

I added an example in the manual, but actually my real motivation is the following feature.

I'm subscribed to a bunch of mailing lists and they're quite noisy. So I don't want to see them in my standard unread bookmark. Because I don't want to hardcode this list of mailing lists to exclude, I store them in a variable.

I have a header view action to add the current message's mailing list to that variable.
And I have another bookmark that shows the unread messages for only one list, by asking me what list I want to look at.

```elisp
(defcustom ibizaman/mu4e-unread-excluded-lists nil
  "Mailing lists to be excluded from default unread view."
  :group 'mu4e
  :type '(repeat string))

(defun ibizaman/mu4e-add-message-list-to-excluded-lists (msg)
  (let ((list (mu4e-message-field msg :mailing-list)))
    (add-to-list 'ibizaman/mu4e-unread-excluded-lists list)
    (message "Added %s to excluded list" list)))

(add-to-list 'mu4e-headers-actions
             '("Exclude list" . ibizaman/mu4e-add-message-list-to-excluded-lists) t)

(defun ibizaman/mu4e-generate-unread-filter ()
  (concat "flag:unread AND NOT flag:trashed"
          (mapconcat (lambda (v) (concat " AND NOT list:" v))
                     ibizaman/mu4e-unread-excluded-lists "")))

(defun ibizaman/mu4e-get-unread-list-filter-query (wanted-list)
  (interactive (list (completing-read "List: " ibizaman/mu4e-unread-excluded-lists)))
  (concat "flag:unread AND NOT flag:trashed AND list:" wanted-list))

(setq mu4e-bookmarks
      `(,(make-mu4e-bookmark
          :name  "Unread messages not list"
          :query (lambda () (ibizaman/mu4e-generate-unread-filter))
          :key ?u)
        ,(make-mu4e-bookmark
          :name  "Unread list messages"
          :query (lambda () (call-interactively 'ibizaman/mu4e-get-unread-list-filter-query))
          :key ?l))
```

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -454,7 +454,9 @@ KAR, or raise an error if none is found."
 		 (mu4e-bookmarks))
 	    (mu4e-warn "Unknown shortcut '%c'" kar)))
 	 (expr (mu4e-bookmark-query chosen-bm))
-	 (query (eval expr)))
+         (expr (if (not (functionp expr)) expr
+                 (funcall expr)))
+         (query (eval expr)))
     (if (stringp query)
       query
       (mu4e-warn "Expression must evaluate to query string ('%S')" expr))))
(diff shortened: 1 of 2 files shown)


djcb:
Looks good - merged. Thanks!

PR #1340, opened 2018-11-14
Non-breaking spaces should be replaced with a space and not removed

### LINKED ISSUE #1339: mu4e-message-outlook-cleanup eats unicode nonbreaking space (U+00A0)

When emails contain unicode non-breaking space these get displayed at least as spaces, if not actual non-breaking spaces.

## Actual behavior

Non-breaking spaces are removed from body so if words are separated by them then there are no spaces anymore.

If if an email has `foo=A0bar` (where `=A0` is a non-breaking space) - this gets displayed as `foobar` rather than `foo bar`

## Steps to reproduce

Open an email with non-breaking spaces between words and notice the words flow into one-another.

mu 1.0, emacs 26.1, Ubuntu 18.04

## Any other detail
This is due to `mu4e-message-outlook-cleanup` which purposely replaces non-breaking spaces with the empty string - so either this should replace non-breaking spaces with a single space, OR only replace multiple consecutive non-breaking spaces with the empty string. Or perhaps replace the first non-breaking space with a real space, and then remove any following non-breaking spaces?

alexmurray:
Fixes #1339

### DIFF
--- a/mu4e/mu4e-message.el
+++ b/mu4e/mu4e-message.el
@@ -247,6 +247,7 @@ unless PREFER-HTML is non-nil."
       (replace-match
 	(cond
 	  ((string= (match-string 0) "’") "'")
+	  ((string= (match-string 0) " ") " ")
 	  (t ""))))
     (buffer-string)))
 


djcb:
Merged -- thanks!

PR #1341, opened 2018-11-18
mu4e/mu4e-view.el (mu4e-view-open-attachment): Ensure attachments are set

abo-abo:
Before this change, I would press "o" and `mu4e~view-attach-map' would
be nil. Now it works as expected.


djcb:
Merged, thanks!

PR #1360, opened 2019-01-11
mu4e: Fix jumping to specific header in unselect window

rhdxmr:
While I am reading a message in a split view, the cursor of mu4e-headers window moves to top after auto updation and rerun-search.
I found mu4e call mu4e-headers-goto-message-id but it failed to jump to the specific header if the buffer is currently inside unselect window.

Also mu4e~headers-highlight function has similar problem, but it works with with-current-buffer macro.


rhdxmr:
@djcb 
Is this pr unacceptable? If it is, could you tell me why? Then I will correct my pr.

djcb:
Merged -- thanks. Due to my schedule it can take some time...

Issue #1357, opened 2019-01-06
"Do what I mean" mu4e contexts


Ambrevar:
Recently I helped out a friend to set up mu4e with multiple accounts.  This was
not my first time, but it was still rather difficult to go through the whole
process and to explain everything, even for a single account.

What felt most frustrating is that much of the initial configuration feels
unnecessarily picky and mu4e would surely benefit to have smarter default values
to ease migration for newcomers.

Since this is Emacs, I implemented a structure on top of "mu4e-contexts" so that
it fills the "configuration holes" automatically, leaving the user with only the
bare minimum to fill.

I've also added support for mailboxes which don't use the trash flag (like Gmail).

Here is the implementation:

https://gitlab.com/ambrevar/dotfiles/blob/master/.emacs.d/lisp/patch-mu4e-account.el

And here is an example user config (which removed about 100 lines from my
initial config):

```elisp
(setq user-full-name "Pierre Neidhardt")

;; https://docs.gandi.net/en/gandimail/standard_email_settings/index.html
(let ((gandi-smtp-vars '((smtpmail-smtp-server . "mail.gandi.net")
                         (smtpmail-stream-type . starttls)
                         (smtpmail-smtp-service . 587))))
  (make-ambrevar/mu4e-account
   :name "mail"
   :user-mail-address "mail@ambrevar.xyz"
   :sent-folder "Sent"
   :vars gandi-smtp-vars)
  (make-ambrevar/mu4e-account
   :name "work"
   :user-mail-address "pierre@work.place"
   :sent-folder "Sent"
   :condition (lambda (msg)
                (mu4e-message-contact-field-matches
                 msg '(:from :to) "boss@work.place"))
   :vars gandi-smtp-vars))
```

It's quite clear from the above implementation the the `ambrevar/mu4e-account`
structure does not add any new feature, it's just trying to be smarter and could easily merged into `mu4e-context`
structure.  Would you consider it?  I can send a patch.  Let me know!

## Versions of mu, mu4e/emacs, operating system etc.

mu: 1.0c
Emacs: 26.1
OS: GuixSD

Ambrevar:
Sorry if I was unclear: we obviously don't want a separate data structure.  I
made it separate so that my local "hack" was clearly separate from mu4e's
codebase.

What I meant is that we could merge my "mu4e-account" into "mu4e-context".

Last but not least: do you like my "fix" for Gmail's trash?

Let me know and I'll submit a patch.

djcb:
[...]

PR #1362, opened 2019-01-14
Account setup helper + Gmail trash fix

Ambrevar:
This is a first shot at implementing #1357.

It needs some testing and the documentation surely needs some editing :)

### DIFF
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -22,7 +22,7 @@
 
 ;;; Commentary:
 
-;; A mu4e 'context' is a a set of variable-settings and functions, which can be
+;; A mu4e 'context' is a set of variable-settings and functions, which can be
 ;; used e.g. to switch between accounts.
 
 (eval-when-compile (byte-compile-disable-warning 'cl-functions))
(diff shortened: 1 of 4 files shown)


Ambrevar:
Did you have time to look at this?

edumerco:
Hello everyone.

> [...] I like the idea of having a simple way to setup an account 
> [...]

Me too (it wasn't easy to start, but I was learning emacs at the 
same time).

Just to check... it is too crazy to make a guide that asks 
questions (a wizard?) and results in/outputs a basic configuration 
in a file?
If yes, I can't code but can do the desition tree and the texts. 

-- 
eduardo mercovich

 Donde se cruzan tus talentos 
 con las necesidades del mundo, 
 ahí está tu vocación. 
 (Anónimo)

mu4e/mu4e-context.el
```diff
@@ -72,13 +72,135 @@ none."
...
+  ;; Folders:
+  maildir
+  (drafts-folder "drafts")
+  (sent-folder "sent")
+  (trash-folder "trash")
+  (refile-folder "archive")
+  ;; Trash fix.
+  no-trash-flag
+  ;; Rule for matching the context.
+  predicate)
```
djcb: `mu4e-context` is very simple and should stay unchanged, I think. So no extra variables etc. we have `vars` for that already; same for `predicate` and `match-func` already.

All the extra magic can go to `make-mu4e-context-account`.

mu4e/mu4e-headers.el
```diff
@@ -744,6 +744,28 @@ after the end of the search results."
...
+    (if (not (seq-filter (lambda (re)
+                           (string-match re msg-dir))
+                         mu4e-move-to-trash-patterns))
+        (mu4e-headers-mark-for-trash)
+      (mu4e-mark-set 'move (if (functionp mu4e-trash-folder)
+                               (funcall mu4e-trash-folder (mu4e-message-at-point))
+                             mu4e-trash-folder))
+      (mu4e-headers-next))))
 
 ;;; headers-mode and mode-map ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
djcb: Oh, that looks useful

mu4e/mu4e.texi
```diff
@@ -2626,6 +2651,50 @@ no context matches (or if you always want to be asked).
...
+@itemize
+@item Only the @code{name} slot is mandatory.
+@item The maildir default to the context name.
+@item Folders only need to be given a name, not a relative path.
+They will be automatically stored under the maildir.
+@item When the @code{match-func} is not provided, the context is matched
+against @code{predicate} if provided or the maildir of the current
+message otherwise.
+@end itemize
+
```
djcb: Looks good; perhaps an gmail example as well?

And this could use a line: "If the context created by `make-mu4e-context-account` is not enough, you can use the generated context using (e.g. using `M-x describe-variable mu4e-contexts`), and tweak the context as needed."

Ambrevar:
@edumerco : That would be a very good idea.  My patch is a first step in that direction.

edumerco:
Hi Pierre.

I'm a UX guy, just tell me when we do start working on this and we'll
arrange it somehow. :)

Best...

-- 
eduardo mercovich

Donde se cruzan tus talentos
con las necesidades del mundo
ahí está tu vocación.

mu4e/mu4e.texi
Ambrevar: I'd avoid a Gmail example because I don't want to suggest a service that goes very much against the free software philosophy (to which mu and mu4e adhere).

Gmail only needs `no-trash-flag` set to t, so it's trivial anyways.

Ambrevar:
Hi @edumerco,
Thanks for the offer but after this patch I won't have time to work on mu4e anytime soon.
Your best bet is to see with @djcb.  I suggest you open an issue to make the discussion publicly visible.

mu4e/mu4e-context.el
Ambrevar: Updated: see #1390.

Ambrevar:
@djcb: Sorry for the delay.  I've taken your changes into account.  We still need some testing :p

edumerco:
Hi Pierre.

I can only do it with others, so I'll wait until at least one other wants
to do this too... :)

Best regards...

-- 
eduardo mercovich

Donde se cruzan tus talentos
con las necesidades del mundo
ahí está tu vocación.

djcb:
Note that this has been superseded by #1390 which has been merged.

PR #1368, opened 2019-01-22
mu-config: add NO_COLOR environment variable

oleskiewicz:
Other than `MU_NOCOLOR`, there is a conventional environment variable `NO_COLOR`, which is shared across many other projects – see https://no-color.org.

This is a very simple change which would make `mu` support that.


djcb:
Where is NO_COLOR defined?

oleskiewicz:
I forgot to surround it with the inverted commas, my apologies.

PR #1374, opened 2019-02-02
Increase `max-specpdl-size' when decoding messages

Chris00:
Some messages (e.g. Outlook automatic replies) need a greater value
than the default one of `max-specpdl-size' to be displayed properly.

An example of such message is [attached](https://github.com/djcb/mu/files/2824519/message.txt)


Chris00:
Any opinion on this PR?

djcb:
I'm not too fond of influencing emacs internals like this... but I guess here it's justified (and only let-bound). 

So, merged. Thanks!

PR #1375, opened 2019-02-02
mu4e: properly display messages in the draft folder (with gnus)

Chris00:
This is necessary to display messages with UFT-8 chars when visiting the draft folder.  Here is an [example of such a message](https://github.com/djcb/mu/files/2824530/draft.txt).

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -360,10 +360,14 @@ article-mode."
     (switch-to-buffer (get-buffer-create mu4e~view-buffer-name))
     (erase-buffer)
     (unless marked-read
-      ;; when we're being marked as read, no need to start rendering the messages; just the minimal
-      ;; so (update... ) can find us.
-      (mm-disable-multibyte)
+      ;; when we're being marked as read, no need to start rendering
+      ;; the messages; just the minimal so (update... ) can find us.
       (insert-file-contents-literally path)
+      (unless (message-fetch-field "Content-Type" t)
+        ;; For example, for messages in `mu4e-drafts-folder'
+        (let ((coding (or (default-value 'buffer-file-coding-system)
+                          'prefer-utf-8)))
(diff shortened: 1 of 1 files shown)


Chris00:
I added a commit that removes
```elisp
      (article-de-base64-unreadable)
      (article-de-quoted-unreadable)
```
because they actually *prevent* properly decoding base64 messages.  Here is  [an example](https://github.com/djcb/mu/files/2824600/decoding.txt)

mu4e/mu4e-view.el
```diff
@@ -362,14 +362,17 @@ article-mode."
...
-      (mm-disable-multibyte)
       (insert-file-contents-literally path)
+      (message-narrow-to-headers-or-head)
+      (if (message-fetch-field "Content-Type" t)
+          (widen)
+        ;; For example, for messages in `mu4e-drafts-folder'
+        (erase-buffer)
+        (mm-enable-multibyte)
+        (insert-file-contents path))
       (setq
```
djcb: Can you explain what this block does and esp., why?

djcb:
The change does seem to work with all "problem case" messages, great!

mu4e/mu4e-view.el
Chris00: It inserts the message with multibytes turned on.  This enables the interpretation of UTF-8 byte sequences for draft messages — which do not have a charset.

djcb: Thanks, I see... can you add a comment tin the code that explains that?

Do we really need to re-load the file from disk (`insert-file-contents`)? Would be nice if we could avoid that.

Chris00: Thanks for prompting me to do a better job! ☺  I reworked the commit not to read the file twice.

djcb:
Merged -- thanks for your efforts!

Issue #1385, opened 2019-02-20
mu4e-action-view-in-browser doesn't show any headers


davidcsterratt:
## Expected or desired behavior

I'd like the From, To, Subject and Date headers to appear in the HTML page opened with `mu4e-action-view-in-browser`

## Actual behavior

Only the body of the message appears.

Chris00:
@davidcsterratt I guess this [can be closed now](https://github.com/djcb/mu/pull/1386).

davidcsterratt:
Yes - there's a comment on the pull request, but I think this issue can be closed.

PR #1386, opened 2019-02-20
Issue #1385: get mu4e-action-view-in-browser to show headers

davidcsterratt:
Here is a very simple patch that achieves what I want, though I am sure there are nicer ways of doing it.


djcb:
That's useful... let's take this in. Thanks!

joostkremers:
This is causing me problems. The patch uses a function `mu4e-html-construct-contacts-header`, which doesn't appear to exist. It seems the intended function is `mu4e~view-construct-contacts-header`.

PR #1390, opened 2019-03-04
Easy accounts with make-mu4e-context-account

Ambrevar:
Follow up on #1362 .

### DIFF
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -22,7 +22,7 @@
 
 ;;; Commentary:
 
-;; A mu4e 'context' is a a set of variable-settings and functions, which can be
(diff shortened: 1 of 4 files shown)


djcb:
Nice work -- I'm ready to merge this, do you have any final changes?

djcb:
(in particular, perhaps you could send a message to the mailing-list after the merge, so people will find out and you can take all credit :-) )

Ambrevar:
Thanks!

No final change, but beware, this has only be tested in my simple use
case.  I suspect you'll have to update/fix some parts to accommodate
future user needs :)

djcb:
Thanks -- I'll merge it, let's give it a test-run!

PR #1395, opened 2019-03-11
Look at all frames for header buffer when auto updating headers

kevinjfoley:
Currently if you have new messages but are in another frame `*mu4e-headers*` isn't auto updated.


djcb:
Merged -- thanks!

PR #1405, opened 2019-04-18
Quote '@' symbols in texinfo's @code section

wavexx:
Noticed while rebuilding today's master.


djcb:
I had an update but forgot to push it... so we'll take this one. Thanks!

PR #1406, opened 2019-04-20
Better conversion to HTML

Chris00:
Before this commit the HTML was looking as
```
From: From: email
To: To: email
```


djcb:
Oh thanks!

PR #1407, opened 2019-04-21
Escape HTML entities

Chris00:
In addition to the [previous fix](https://github.com/djcb/mu/pull/1406), one also has to escape HTML entities.


djcb:
Looks good, merged. Thanks!

PR #1413, opened 2019-04-29
Fix mu4e~headers-remove-handler

Chris00:
- The documentation speaks of `SKIP-HOOK` which is not an argument;
- I do not understand why `get-buffer-window-list` is limited to the current frame (if we were viewing the message in another frame, should not we close it too?).


djcb:
Yeah, that makes sense -- thanks. Merged.

PR #1414, opened 2019-04-30
mu4e: fix a remaining local variable 'short'

garaud:
Hi,

There was a traceback which looked like "void function error short" when the function `mu4e~action-header-to-html` was called.

Great software, keep going :slightly_smiling_face: !!

Thanks!


Chris00:
Thanks for spotting this!

djcb:
Merged, thanks!

PR #1422, opened 2019-05-11
Fix the behavior of guile/examples/msg-graphs and a typo in guile/mu/plots.scm

### LINKED ISSUE #1421: Typo in plot.scm
The file `plot.scm` in `guile/mu` exports the variable `mu:port`.  However, in the file, this variable is not defined and a variable `mu-port` is defined instead.

### LINKED ISSUE #1423: guile/examples/msg-graphs fails when called with --text option

The documented behavior of `guile/examples/msg-graphs` is that when called without the `--text` option, it should produce a graphical plot, while when called with `--text`, it should output a text-mode plot

## Actual behavior

Currently, however, it produces a text-mode graph when called *without* the `--text` option, and fails with an error when called with the `--text` option as described below.

## Steps to reproduce
```
./msg-graphs --what=per-day --text
Backtrace:
           4 (apply-smob/1 #<catch-closure 557b444c7720>)
In ice-9/boot-9.scm:
    705:2  3 (call-with-prompt _ _ #<procedure default-prompt-handle…>)
In ice-9/eval.scm:
    619:8  2 (_ #(#(#<directory (guile-user) 557b44560140>)))
In mu/plot.scm:
    67:13  1 (mu:plot-histogram _ "Messages per weekday matching " # …)
In unknown file:
           0 (string-append "reset\n" "set term " #t "\n" "set titl…" …)

ERROR: In procedure string-append:
In procedure string-append: Wrong type (expecting string): #t
```

`Emacs 26.2`, `guile-2.2`, `mu` from `master` along with the typo fixed in 7498760.

## Any other detail

As made clear by the error statement, the issue is that in `mu:plot` (in file `guile/mu/mu.scm`), the expression `(or output "dumb")` evaluates to the string `"dumb"` when `output` is `#f` and to the symbol (*not* string) `#t` when `output` is `#t`.  The code in `guile/examples/msg-graphs` calls `mu:plot` with `output` set to `#t` when `--text` is passed (thus the above error message) and with `output` set to `#f` when `--text` is not passed (so that the above form evaluates to `"dumb"` and a text-mode output is produced, contrary to the documentation).

piyushs:
7498760 fixes the minor typo pointed out in #1421 while c4e037c fixes #1423


djcb:
Merged, thanks!

piyushs:
Thanks.  This pull request also fixes #1421, so you might consider closing that too.

PR #1437, opened 2019-06-19
Replace AM_PROG_LIBTOOL with LT_INIT

sham1:
Since mu seems to already gate Autoconf to have version >= 2.68, which was released back in 2010, it seems rather silly to not use more modern m4 macros, like for instance `LT_INIT` which requires libtool version >= 1.9b, which itself was released in 2004 and where the first stable version, libtool 2.2 was released back in 2008.

Any distro that has a sufficiently recent Autoconf has a recent enough libtool.


djcb:
Merged, thanks!

PR #1442, opened 2019-07-01
mu4e: Allow "washing" of URLs in message view

rotty:
See the commit message for details; this PR does not yet include
documentation changes, as I first want to check whether this is (in
principle) considered a reasonable addition to mu4e.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -978,6 +978,36 @@ If the url is mailto link, start writing an email to that address."
   "Regexp that matches the beginning of http:/https:/mailto:
 URLs; match-string 1 will contain the matched URL, if any.")
 
+(defvar mu4e-view-url-washers nil
+  "A list of functions which can replace displayed URL text and/or targets.
+
+When displaying a URL, the list is walked, stopping at the first
+function which returns non-`nil' on the given URL, and using its
+return value for the URL text displayed and potentially replacing
+the link target as well.
+
+Each function in the list must accept a single argument, which is
+the string representing the URL. The return value must be one of:
+
+- `nil', to continue searching.
+
+- A string, which will be used as a replacement display text. The
+  link target will stay the same.
+
+- A two-element list, where the first element is the replacement
+  display text, and the second element the replacement link
+  target.")
+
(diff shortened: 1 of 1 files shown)


Chris00:
I glanced over your code but it does not seem to also target the Gnus-based viewer (which *may* become the default in some time...).

rotty:
Yeah, I've realized that after coding the patch, and was not motivated enough to address the Gnus view, since I had a solution that addressed my immediate issue; I'll take a look at what implementing this based on the Gnus article view would entail.

djcb:
Effort certainly appreciated, but yeah, we're trying to avoid any updates to the "old" mu4e-view.

Could this be used with the gnus-view? Or perhaps there's something like that there already?
There's https://www.gnu.org/software/emacs/manual/html_node/gnus/Article-Washing.html at least.

Chris00:
On 27 February 2020 at 00:23 CET, Dirk-Jan C. Binnema <notifications@github.com> wrote:
>
> Effort certainly appreciated, but yeah, we're trying to avoid any updates to the "old" mu4e-view.
>
> Could this be used with the gnus-view? Or perhaps there's something like that there already?  There's https://www.gnu.org/software/emacs/manual/html_node/gnus/Article-Washing.html at least.

At least we wan try to piggyback on that — at the moment, these commands return `gnus-summary-select-article: This is a pseudo-article`.

djcb:
Yeah, all those commands expect the "Summary buffer" (similar to our headers-view) to do something. Not sure if the gnus people would be interested in changing that, and/or how far we could get with e.g. advising functions. I'd like to avoid including modified versions as part of mu4e.

Chris00:
On 27 February 2020 at 08:45 CET, Dirk-Jan C. Binnema <notifications@github.com> wrote:
>
> Yeah, all those commands expect the "Summary buffer" (similar to our headers-view) to do something. Not sure if the gnus people would be interested in changing that, and/or how far we could get with e.g. advising functions. I'd like to avoid including modified versions as part of mu4e.

Advising it would likely do the trick — that was the point of introducing the local variable `mu4e~view-rendering`.

djcb:
Okay, guess this is obsolete then.

PR #1450, opened 2019-07-21
Bring g_vasprintf into scope on OpenBSD

c0dev0id:
This fixes compilation on OpenBSD.


djcb:
Hmm... on Linux at least, `Use of _XOPEN_SOURCE_EXTENDED  in  new source code should be avoided.`

Does `#define _XOPEN_SOURCE 500` work?

djcb:
Closed, no activity.

Issue #1455, opened 2019-07-30
make-mu4e-context-account: keyword argument error


npjg:
## Expected or desired behavior

Successfully creating a context with `make-mu4e-context-account`.

## Actual behavior

Using the example context-account configuration provided in the documentation, context creation fails with the following:
```
(error "Keyword argument personal not one of (:enter-func :leave-func :match-func :vars :user-mail-address :smtpmail-smtp-user :maildir :drafts-folder :sent-folder :trash-folder :refile-folder :no-trash-flag :predicate)")
```
Note that this error recurs for whatever keyword argument is placed first in the first `make-mu4e-ccontext-account` statement of the example configuration reproduced below.

## Steps to reproduce

Open Emacs with a default configuration (`emacs -q`), and then execute the following:
```
(add-to-list 'load-path "/usr/share/emacs/site-lisp/mu4e")
(require 'mu4e)

(let ((gandi-smtp-vars '((smtpmail-smtp-server . "mail.gandi.net")
                         (smtpmail-stream-type . starttls)
                         (smtpmail-smtp-service . 587))))
  (make-mu4e-context-account
   :name "personal"
   :user-mail-address "john@doe.xyz"
   :sent-folder "Sent"
   :vars gandi-smtp-vars)
  (make-mu4e-context-account
   :name "work"
   :user-mail-address "john@work.org"
   :sent-folder "Sent"
   :predicate (lambda (msg)
                (mu4e-message-contact-field-matches
                 msg ’(:from :to) "boss@work.org"))
   :vars gandi-smtp-vars))
```

## Versions

`mu` 1.3.3, `mu4e` 1.3.3 C, Emacs 26.2, Fedora 30

djcb:
@Ambrevar: ^^ Any idea?

Ambrevar:
At first glance, looks like `name` is not in the list of supported keyword arguments.
I don't have much time at the moment, let me know if you need more help to investigate.

np0g:
That's what I thought at first, but as noted above, when I make another keyword argument, say `user-mail-address` the first, the error occurs with that argument, although `user-mail-address` is indeed in the list of supported arguments.
```
make-mu4e-context-account: Keyword argument john@doe.xyz not one of (:enter-func :leave-func :match-func :vars :user-mail-address :smtpmail-smtp-user :maildir :drafts-folder :sent-folder :trash-folder :refile-folder :no-trash-flag :predicate)
```
I'll keep investigating.

[...]

PR #1456, opened 2019-07-30
mu4e: fix `make-mu4e-context-account` to see `name` keyword arg

npjg:
Simple fix from #1455, to enable documented use of the `name` keyword argument in `make-mu4e-context-account`.


djcb:
Merged -- thanks!

PR #1465, opened 2019-08-09
Balance parens in mu4e-compose-context-policy

rnkn:
Fixes unbalanced parens in mu4e-compose-context-policy, which prevents Custom mode UI working.

Using standard indentation helps prevent missing these little issues.


rnkn:
Actually I just saw that almost every custom option is unbalanced...

djcb:
Ah, good catch... apparently not something a lot of people use

djcb:
Merged, thanks!

rnkn:
> Ah, good catch... apparently not something a lot of people use

Thanks :)

And, yeah! I've noticed an almost hostility towards using custom, but 
it's so incredibly useful.. basically type-checking for Emacs lisp. 
Pretty often people's bugs can be resolved simply by setting things with 
custom instead of hand-coding (of course maybe there's an ego hit 
involved there...)

PR #1468, opened 2019-08-16
Initialize my addresses

wavexx:
Updated today to latest sources and found the "personal" flag didn't work as expected.

In set_personal_addresses(), aside from updating the db, we also need to refresh the addresses in the Store itself, as this is subsequently used during indexing (this would fail to pickup addresses put in --my-addresses during index --reindex otherwise).

I also fixed the "FIXME" in each_contact_check_if_personal using personal_addresses() directly. I assume a GSList was only necessary due to the old code, so let's avoid copies.

In the for(auto) statements I noticed I got a lot of gcc warnings regarding comparisons with size_t, as 0 is deduced as int. I fixed one such occurrence, but I think it's better to use size_t explicitly in these cases.

### DIFF
--- a/lib/mu-store.cc
+++ b/lib/mu-store.cc
@@ -709,7 +709,7 @@ mu_store_personal_addresses (const MuStore *store)
 
         const auto size = self(store)->personal_addresses().size();
         auto addrs = g_new0 (char*, 1 + size);
-        for (auto i = 0; i != size; ++i)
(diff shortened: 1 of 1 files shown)


djcb:
Oh, yeah, I had all forgotten about the commented-out parts and what we need to fix them. Good catch, and thanks for the patch!

PR #1470, opened 2019-08-16
update with opensuse distro instructions

MalloZup:
# Description:

this PR add a short description for users running openSUSE Linux distro. 

I could also initially not find `mu` in the pkg, so at least providing doc should be ok,


djcb:
Merged -- thanks!

PR #1471, opened 2019-08-19
[WIP] Start fixing cl-lib and lexical-binding issues

thierryvolpiatto:
I started recently having weird errors about `incf`, after looking at the code, I see you are using both `cl` and `cl-lib` and `lexical-let` (part of `cl`) in some places and `lexical-binding` in other places.
I saw also an abusive usage of `add-to-list` (incompatible with lexical-binding in some places).
So I started a new branch to fix this, please have a look, it is not fully finished and tested, but I have no errors so far.
Thanks.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -1,4 +1,4 @@
-;; mu4e-compose.el -- part of mu4e, the mu mail user agent for emacs
+;; mu4e-compose.el -- part of mu4e, the mu mail user agent for emacs -*- lexical-binding: t -*-
 ;;
 ;; Copyright (C) 2011-2019 Dirk-Jan C. Binnema
 
@@ -66,8 +66,7 @@
 ;;
 
 ;;; Code:
-(eval-when-compile
-  (require 'cl))
+(require 'cl-lib)
 (require 'message)
 (require 'mail-parse)
 (require 'smtpmail)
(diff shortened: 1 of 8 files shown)


djcb:
Overall, looks useful, thanks. What's the issue with the `add-to-list` vs `cl-pushnew`?
And since what emacs version are these new `cl-` functions available?

thierryvolpiatto:

> Overall, looks useful, thanks. What's the issue with the add-to-list
> vs cl-pushnew?

`add-to-list` is not working on lexical vars, generally it should not be
used in elisp programs, only in user configuration files.

I see in the places where I have replaced add-to-list by cl-pushnew you
was using the 'append argument of add-to-list which add the new element
at end of list instead of front, is it really important to do so?

> 
> And since what emacs version are these new cl- functions available?

The cl- functions I used here are not new, they exist since cl-lib have
been introduced, sure in emacs-24.1, maybe earlier don't remember.

Thanks.


-- 
Thierry

Get my Gnupg key:

thierryvolpiatto:
BTW we should add "-*- lexical-binding: t -*-" to all files for consistency, like this you don't have to think at lexical or dynamic when you write code.

thierryvolpiatto:
Using now this branch since ten days without errors, do you want I rebase it to fix commit messages?

djcb:
Oh, yes please. Then I can merge it. Thanks.

thierryvolpiatto:

> Oh, yes please. Then I can merge it. Thanks.

I am now travelling, after mid september I will do it.

Thanks.

> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.*


-- 
Thierry

Get my Gnupg key:

Issue #184, opened 2013-04-06
mu4e: error in process filter: Error 2: not a readable dir: '%s'


jleechpe:
I have mu+offlineimap running on Windows 7.

When running `mu index` from the bash shell it successfully updates.

However when trying to use `U` in mu4e I get the following error after a few seconds:

```
[mu4e] Retrieving mail...
error in process filter: mu4e-error-handler: Error 2: not a readable dir: '%s'
error in process filter: Error 2: not a readable dir: '%s'
```

Using the commands to view mail does work.

PR #1473, opened 2019-08-22
Fix get_checked_path error message

seblemaguer:
Related to #184, I realize the path was not part of the error but only '%s'. This is the minor fix which actually show the given path.


djcb:
This is not quite how `sprintf` works... I've fixed it a different way.

PR #1484, opened 2019-09-14
Completely use cl-lib and lexical-binding (rebase #1471)

thierryvolpiatto:
Previously both cl-lib.el and cl.el were used, now use only cl-lib.el.
Use lexical-binding where needed instead of requiring cl just for
`lexical-let`.
Replace some add-to-list with cl-pushnew as add-to-list is not
recommended in lisp program and anyway doesn't work properly with
lexical binding.


djcb:
Merged -- thanks!

thierryvolpiatto:

> Merged -- thanks!

Great, thanks, now for completeness, what can be done is bind locally to
all *.el files `lexical-binding`, and optionally replace all `add-to-list`
calls with `cl-pushnew` or whatever.
I can make a new PR for this if you want.

-- 
Thierry

Get my Gnupg key:

PR #1512, opened 2019-11-17
lib: Use more portable locale en_US.UTF-8

derekschrock:
On some systems (FreeBSD) en_US.utf8 doesn't exist.  Use a portable
locale en_US.UTF-8.


djcb:
Merged -- thanks!

PR #1520, opened 2019-12-05
Push mark on beginning/end of buffer

antoine-levitt:
This has been driving me mad for a while: mu4e's M-< and M-> commands were not consistent with the standard ones.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -900,11 +900,17 @@ buffer buried."
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
-(defun mu4e-compose-goto-top ()
+(defun mu4e-compose-goto-top (&optional arg)
   "Go to the beginning of the message or buffer.
 Go to the beginning of the message or, if already there, go to the
-beginning of the buffer."
-  (interactive)
+beginning of the buffer.
+
+Push mark at previous position, unless either a \\[universal-argument] prefix
+is supplied, or Transient Mark mode is enabled and the mark is active."
+  (interactive "P")
+  (or arg
+      (region-active-p)
+      (push-mark))
   (let ((old-position (point)))
     (message-goto-body)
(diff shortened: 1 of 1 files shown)


Chris00:
You mean in a draft?

antoine-levitt:
Yes.

Chris00:
You can press <kbd>M-<</kbd> twice to go to the beginning of the buffer.  I believe this behavior is the right one but if you do not like it, you can add the following to your config file:
```elisp
(define-key mu4e-compose-mode-map
  (vector 'remap 'beginning-of-buffer) nil)
```

antoine-levitt:
This is not about M-< going to the beginning of the draft (which I like), it's about it not pushing the mark. Look at the diff of this PR.

mu4e/mu4e-compose.el
```diff
@@ -913,13 +919,19 @@ beginning of the buffer."
...
+
+Push mark at previous position, unless either a \\[universal-argument] prefix
+is supplied, or Transient Mark mode is enabled and the mark is active."
+  (interactive "P")
+  (or arg
+      (region-active-p)
+      (push-mark))
   (let ((old-position (point))
-   (message-position (save-excursion (message-goto-body) (point))))
+	(message-position (save-excursion (message-goto-body) (point))))
```
Chris00: No reason to change the indentation here.

Chris00:
> it's about it not pushing the mark. 

OK—read only the description...

mu4e/mu4e-compose.el
antoine-levitt: Emacs did it by itself and I let it because it's right here. I can revert if you want.

Chris00: @djcb has the final word on that.  I mentioned it justbecause it is unrelated to the purpose of this change.

djcb:
Looks good. But @Chriss00 is right, that indentation seems a bit off. Perhaps a tabs-vs-spaces thing? We try to standardize on spaces (see the .editorconfig).

antoine-levitt:
Reverted the indentation

djcb:
Merged -- thanks!

antoine-levitt:
Thanks!

PR #1526, opened 2019-12-14
mu4e-util: allow ESC to exit a choice selection

seanfarley:
I occasionally find myself pressing escape to exit a menu choice in mu4e
and it's a bit unfortunate that nothing happens. As best I could tell,
since this is a C call, none of the keymaps are checked for
`(keyboard-quit)' so we need to explicitly check for ESC here.

----

#

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -253,6 +253,7 @@ trying an exact match."
     (while (not chosen)
       (message nil);; this seems needed...
       (setq choice (read-char-exclusive prompt))
+      (if (eq choice 27) (keyboard-quit)) ;; quit if ESC is pressed
       (setq chosen (or (member choice choices)
 		     (member (downcase choice) choices)
 		     (member (upcase choice) choices))))


seanfarley:
I'm not completely sure this is the best approach, so feedback is welcomed

theophilusx:
Isn't C-g the 'standard' way to quit something in Emacs?

How would using ESC affect other commands which use ESC like exiting
recursive edit, running M-x in a terminal without meta support (which is
the most common one I use ESC for)?

Tim

> I'm not completely sure this is the best approach, so feedback is welcomed
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.


-- 
Tim Cross

seanfarley:
Tim Cross <notifications@github.com> writes:

> Isn't C-g the 'standard' way to quit something in Emacs?

Yep, and C-g continues to work, this just adds another option.

> How would using ESC affect other commands which use ESC like exiting
> recursive edit, running M-x in a terminal without meta support (which is
> the most common one I use ESC for)?

Since `read-char-exclusive` is not a recursive edit (and I can't figure
out how it would be in a recursive edit but this is pushing my internal
emacs knowledge) I'm not sure this would mess anything up in this case?

djcb:
That looks useful, merged, thanks.

seanfarley:
Thanks!

PR #1529, opened 2019-12-20
Whitelist underscores if replying to mailing-list

jaknel:
When replying to a mailing list post from some_address@example.org, Mu4e incorrectly truncated the reply to down to just "some". Whitelisting underscores in the regex resolves the issue.


jaknel:
Moved the underscore to preserve - as the final character.

This actually fixed another issue where the mailing-list address was being included in both the TO and CC fields.  Feel free to close this PR, and commit the _ directly, if it actually solves the problem, or more changes are required.

djcb:
Okay, commit directly. Thanks for the fix!

PR #1530, opened 2019-12-22
mu4e: switch mu4e-contexts from defvar to defcustom

Vonfry:
In document, mu4e-contexts is described which can be customed by user, why we don't define it with defcustom to let `custom-set-variables` work before loading mu4e.


djcb:
Merged -- thanks!

PR #1532, opened 2019-12-27
mu configure.ac: allow emacs 28.xx to build mu4e

Bdanilko:
Changed autoconfig file to add support for emacs 28.xx versions building mu4e


djcb:
Didn't know 28 had started already... anyway, thanks!

PR #1535, opened 2019-12-31
mu4e: define mu4e-view-actions with defcustom instead of defvar

pedrormjunior:
It allows the customization directly from Emacs Custom mode.  It is
done in a similar way as for `mu4e-view-attachment-actions'.


djcb:
Merged, thanks!

PR #1541, opened 2020-01-12
mu4e: inform alternative for getting confirmation before sending

pedrormjunior:
The `message-send` command already uses the variable `message-confirm-send` to check if a confirmation should be asked before sending.  Then, setting this variable is a more straightforward alternative for getting those confirmations before sending.


djcb:
Looks good, thanks!

PR #1548, opened 2020-01-21
Avoid loading all gnus package

thierryvolpiatto:
when not necessary.

* mu4e/mu4e-view.el (mu4e~view-gnus): Require gnus-art here instead on
top of file.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -35,7 +35,6 @@
 (require 'mu4e-actions)
 (require 'mu4e-message)
 
-(require 'gnus-art)
 (require 'comint)
 (require 'browse-url)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
I noticed since some months a big slowdown when opening a mail the first time, this was due to gnus loading when mu4e-view is autoloaded, requiring gnus-art in the appropriate function fix this slowdown.
Also more strange I noticed sometimes a gnus buffer popping up in the background when lauching mu4e, maybe it is a bug in gnus itself?  I will see if it reproduce with this change.
Note: I am not using `mu4e-view-use-gnus`.

djcb:
Looks useful, thanks! Merged.

PR #1554, opened 2020-01-25
mu4e: use `mu4e~update-name` de facto

pedrormjunior:
Instead of hard coding the buffer name `" *mu4e-update*"`, use the buffer name defined in `mu4e~update-name`.  Also, the default name defined in `mu4e~update-name` were changed to match the previously hard-coded invisible buffer name.


djcb:
Oh thanks, merged.

PR #1555, opened 2020-01-27
Unquote lambda's

thierryvolpiatto:
* mu4e/mu4e-context.el (mu4e-context-switch):         Do it.
* mu4e/mu4e-contrib.el (mu4e-view-mode-hook):         Do it.
(mu4e-headers-mode-hook):                             Do it.
* mu4e/mu4e-icalendar.el (mu4e-icalendar-reply-ical): Do it.
* mu4e/mu4e-utils.el (mu4e~start):                    Do it.
(mu4e-split-ranges-to-numbers):                       Do it.

### DIFF
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -234,7 +234,7 @@ non-nil."
       (when (mu4e-context-enter-func context)
 	(funcall (mu4e-context-enter-func context)))
       (when (mu4e-context-vars context)
-	(mapc #'(lambda (cell)
+	(mapc (lambda (cell)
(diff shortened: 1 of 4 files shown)


Chris00:
[Movitation](http://endlessparentheses.com/get-in-the-habit-of-using-sharp-quote.html) (which should be mentioned in the commit message IMHO): [Under dynamic binding, `lambda` forms are self-quoting](https://www.gnu.org/software/emacs/manual/html_node/elisp/Anonymous-Functions.html#Anonymous-Functions)

djcb:
Ok thanks... merged.

PR #1556, opened 2020-01-27
Rewrite mu4e~main-bookmarks to have bookmarks aligned

thierryvolpiatto:
* mu4e/mu4e-main.el (mu4e~main-bookmarks): Do it.

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -108,31 +108,35 @@ clicked."
 
 (defun mu4e~main-bookmarks ()
   ;; TODO: it's a bit uncool to hard-code the "b" shortcut...
-  (mapconcat
-	  (lambda (bm)
-      (unless (plist-get bm :hide)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Before:
Capture d’écran_2020-01-27_08-44-34
After:
Capture d’écran_2020-01-27_08-47-16

djcb:
Oh, that's a nice improvement, thanks! I have to get used to `cl-loop` though...

thierryvolpiatto:

> Oh, that's a nice improvement, thanks!

Thanks.
What would be nice now is to update the unreads in realtime, actually we
have to close mu4e session and restart it to have the display updated,
sometimes it is even necessary to call `mu4e-update-index`.
At least the revert-buffer fn should do that.

> I have to get used to cl-loop though...

Yes, cl-loop is a nice tool.

-- 
Thierry

Get my Gnupg key:

PR #1557, opened 2020-01-27
Load gnus-art when compiling

Chris00:
Commit edce6354160eccaf7559e038996421f7e5a885ec only loads it
dynamically when executing the function mu4e~view-gnus but as a
consequence many variables are undefined at compile time and, more
importantly, it is no longer possible to override the variable
gnus-display-mime-function with a let binding before calling
gnus-article-prepare-display


djcb:
Merged, thanks.

PR #1562, opened 2020-01-29
Update bookmarks status from main buffer

thierryvolpiatto:
This PR allow refreshing the main buffer with `revert-buffer` (g). Before the unread count were not updated when hitting "g", now it does.
Also I tried clarifying the call to `mu4e~start` by separating it in two functions.
Same for the `revert-buffer-function`, it is now split in two functions.
Please have a look before merging.

Next step would be to refresh the unreads count after hitting "U" and update finish automatically without having to hit "g". Same after having read the unread and quitting `mu4e-headers` .

### DIFF
--- a/mu4e/mu4e-context.el
+++ b/mu4e/mu4e-context.el
@@ -239,7 +239,7 @@ non-nil."
 	  (mu4e-context-vars context)))
       (setq mu4e~context-current context)
       (unless (eq mu4e-split-view 'single-window)
-        (mu4e~main-view-real nil nil))
+        (mu4e~main-view-real-1))
       (run-hooks 'mu4e-context-changed-hook)
       (mu4e-message "Switched context to %s" (mu4e-context-name context)))
     context))
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1879,7 +1879,9 @@ other windows."
       ;; now, all *other* windows should be gone. kill ourselves, and return
       ;; to the main view
       (kill-buffer)
-      (mu4e~main-view))))
+      (if (get-buffer-window mu4e~main-buffer-name 'visible)
+          (mu4e~main-view-real-1 t)
+        (mu4e~main-view 'refresh)))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 (provide 'mu4e-headers)
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -145,46 +145,52 @@ clicked."
 ;; buffer switching at the end.
 (defun mu4e~main-view-real (ignore-auto noconfirm)
   (let ((buf (get-buffer-create mu4e~main-buffer-name))
-	       (inhibit-read-only t))
+        (inhibit-read-only t)
+        (pos (point)))
+    (message "Reverting %s..." mu4e~main-buffer-name)
+    ;; Refresh infos from server.
+    (mu4e~start-1)
     (with-current-buffer buf
       (erase-buffer)
       (insert
-        "* "
-	      (propertize "mu4e - mu for emacs version " 'face 'mu4e-title-face)
-	      (propertize  mu4e-mu-version 'face 'mu4e-header-key-face)
-        (propertize "; (in store: " 'face 'mu4e-title-face)
-        (propertize (format "%s" (plist-get mu4e~server-props :doccount)) 'face 'mu4e-header-key-face)
-        (propertize " messages)" 'face 'mu4e-title-face)
+       "* "
+       (propertize "mu4e - mu for emacs version " 'face 'mu4e-title-face)
+       (propertize  mu4e-mu-version 'face 'mu4e-header-key-face)
+       (propertize "; (in store: " 'face 'mu4e-title-face)
+       (propertize (format "%s" (plist-get mu4e~server-props :doccount)) 'face 'mu4e-header-key-face)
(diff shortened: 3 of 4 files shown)


thierryvolpiatto:
Update: After reading unread mails and quitting mu4e-headers, main buffer and unreads are updated correctly :-)

thierryvolpiatto:
Update: Counters are refreshed after updating mails and/or index.

thierryvolpiatto:

> @djcb commented on this pull request.
>
> Thanks -- seems useful...

Yes, otherwise counters are useless, now they are updated each time the
unread status change:
1) After updating mails ("U").
2) After reading unread mails and quitting mu4e-headers.

Before this change, I had to quit mu4e and restart it to have counters
updated which is inconvenient.

The only way I found to update the main buffer was to make a call to
`mu4e~start` from the revert-buffer-function of main mode.
Problem was that `mu4e~start` calls itself `mu4e~main-view`
inconditionally (not sure it is anymore true with your last change, I
have to look).

> but I think it would be cleaner to use mu4e-index-updated-hook from
> mu4e-main.

Ok thanks, I didn't know this hook.

> Also mu4e~main-view-real-1?

It is same as the old revert-buffer function but allows calling it with
a REFRESH argument which is needed (or not) in various places
i.e. allows two behaviors of the revert-buffer function.

See also `mu4e~start` and `mu4e~start-1` and let me know what you think.

What function otherwise would you use in the `mu4e-index-updated-hook` to update
counters? (or what this function if new would invoke?)

Thanks.

--
Thierry

Get my Gnupg key:

thierryvolpiatto:
Here the `mu4e~start` with last changes you made:
```elisp
(defun mu4e~start (&optional func)
[...]
  ;; if we're already running, simply go to the main view
  (if (mu4e-running-p)         ;; already running?
      (when func (funcall func)) ;; yes! run func if defined
    (progn
[...]
```
With this code, calling `mu4e~start` with no arg while mu4e is running is a noop, it does nothing
perhaps what you wanted was:
```elisp
(if (and func (mu4e-running-p))
    (funcall func)
  [...])
```
Also the progn is unneeded in the else clause.

thierryvolpiatto:

> but I think it would be cleaner to use mu4e-index-updated-hook from
> mu4e-main.

I had a look at this hook but don't understand how to use it to revert
main buffer, any help welcome.

Thanks.

PS: Otherwise this PR is working fine AFAICT as it is actually.

-- 
Thierry

Get my Gnupg key:

djcb:
I've updated things a bit, so you get fresh unread/all counts when calling mu4e, and added add `g` binding as well, that was indeed missing.

What's remaining is the updating bit -- that only really matters when pressing 'U' while in the main screen. I think it should be possible to use mu4e-index-updated-hook for that (or else we need to fix that!)... but anyway, I'll add that.

It might still be useful as optimization is having mu4e-index-updated-hook (and the :update / :remove messages) mark things as 'dirty'.. and skip the refresh when we're not dirty. But, at least for me, it works fine as it is.

thierryvolpiatto:

> I've updated things a bit, so you get fresh unread/all counts when
> calling mu4e,

Thanks, but still other things wrong IMHO, see below.

> and added add g binding as well, that was indeed missing.

No, g binding was here and correctly bound to revert-buffer as we derive
from special-mode and we have set revert-buffer-function in
mu4e-main-mode.
Calling revert-buffer should refresh the buffer.

(BTW same for "q" but this is unrelated).

> What's remaining is the updating bit

Not only, quitting mu4e-headers after reading a mail doesn't update.

> -- that only really matters when pressing 'U' while in the main
> screen. I think it should be possible to use mu4e-index-updated-hook
> for that (or else we need to fix that!)... but anyway, I'll add that.

Ok, good, thanks.

One thing also which is unclear is mu4e~start, its behavior have changed
at least 2 times this week, would need to clarify what it does (or
should do) in docstring or comment for further development, I tried to
do this along your changes in this PR (see docstring) but not sure now
if it's right (i.e. my docstring):

1) When mu4e is running and FUNC is defined:
master: (re)start mu4e and call FUNC in the filter process.
before: Call FUNC only.

2) When mu4e is running and FUNC is NOT defined:
master: (re)start mu4e.
before: Same as in master.

3) When mu4e is not running and FUNC is defined:
master: same as 1), and update contexts and requests.
before: same as master but always update contexts and requests.

4) When mu4e is not running and FUNC is NOT defined:
master: same as 2), and update contexts and requests.
before: same as master but always update contexts and requests.


> It might still be useful as optimization is having
> mu4e-index-updated-hook (and the :update / :remove messages) mark
> things as 'dirty'.. and skip the refresh when we're not dirty. But, at
> least for me, it works fine as it is.

You mean update counters when deleting messages?
This should be done in 4) (see below) by this PR.

So to resume, the requirements are:

1) Start mu4e with fresh info: ok.
2) Update main buffer with "g": ok but is should be revert-buffer that
do it.
3) Update main buffer after updating (after "U" finish or M-x
mu4e-update-index)
4) Update main buffer after reading unread mails and coming back from
mu4e-headers.


--
Thierry

Get my Gnupg key:

djcb:
I did a few more changes, I think they should address 1..4 (well, (2) works but does not use revert-buffer yet). I think the other things are working.

There's indeed a bit room for cleaning up the start-up process... I'll think a bit about that. Anyway, thank your!

thierryvolpiatto:

> I did a few more changes, I think they should address 1..4

4) is not working yet.

Thanks.

-- 
Thierry

Get my Gnupg key:

thierryvolpiatto:
I am now always using https://github.com/thierryvolpiatto/mu/tree/update_bmks_status
which IMO DTRT about revert-buffer and also works as expected, I am not sure using the
`mu4e-index-updated-hook` is the good approach specially in the mu4e source code for a specific problem.  Feel free to look at my branch.
Thanks.

Issue #1571, opened 2020-02-11
Offer to cleanup indentation


tarsius:
**Prelude:** I am excited that you have been working more on mu/mu4e recently and would like to help out with the elips side of things. Mu4e has been very stable for me in the past but recent changes brought some breaking changes with them. I'll investigate myself and if at all possible come up with solutions.

**And now:** While looking at recent changes I unfortunately ran into large accidental changes to whitespace. With your permission I would like to help out with that.

Currently the indentation is so inconsistent that I cannot even tell whether you want to use all spaces or the "tabs for indentation and spaces for alignment" style that is used in Emacs for elisp by default. I recommend you use only spaces, but the second approach works too, provided it is used consistently.

Once you have told me which indention style you intended to use, then I can:
1. ensure that you and all contributors will use that style going forward by setting `indent-tabs-mode` directory-locally, and
2. change all existing code to follow that style as well.

If you want I can also do something similar for other types of files.

djcb:
Hey Tarsius,

There have been some quite big changes in the dev series recently, including a bit of breakage for use-cases that aren't mine; sorry about that! But I'm trying to follow-up quickly!

As for the indentation, I'm using the `.editorconfig` in the project root, which is a cross-editor way of specifying some of the settings. For elisp, there's:

```
[*.el]
indent_style                = space
indent_size                 = 2
max_line_length             = 100
```
we could _also_ add dir-locals, as these settings shouldn't change too often.

The indentation is indeed inconsistent, and in some case just _wrong_, but it's not always to see from PRs...

tarsius:
> There have been some quite big changes in the dev series recently,
> including a bit of breakage for use-cases that aren't mine; sorry
> about that! But I'm trying to follow-up quickly!

No problem; I am happy to see you are actively working on this!

> [...] I'm using the `.editorconfig` in the project root, which is
> a cross-editor way of specifying some of the settings. For elisp,
> [...]

I think that *Emacs* lisp is almost always edited using Emacs, so configuring other editors is probably unnecessary in this case. ;-)

> The indentation is indeed inconsistent, and in some case just
> _wrong_, but it's not always to see from PRs...

I have some bad news for you.  The indentation is not wrong **despite** the use of `editorconfig` but **because** of the use of `editorconfig`.  Also its not the contributed changes that are intended wrong but the code that *you* write (because unlike most contributors you actually have `editorconfig' installed).

TODO link to pull-request

[...]

PR #1573, opened 2020-02-17
Keep editorconfig from misconfiguring indentation for Emacs lisp files

tarsius:
As mentioned in #1571 completely messes up indentation of Emacs lisp files.

The commit 173deff6a03bae1171e39ca8da916386ca9bbf67 ("editorconfig: Don't misconfigure indentation for Emacs lisp files") has more details.

In addition to removing the `.editorconfig` configuration this pr configures `.dir-locals.el` and then fixes the indentation.

And then it also fixes quite a few other assorted issues I ran into along the way.  Let me know if this does too much.  I can split this up into more pull-requests and explain in more details why I did what I did.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,2 @@
+((emacs-lisp-mode
+  (indent-tabs-mode . nil)))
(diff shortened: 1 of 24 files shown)


djcb:
Okay, merged-- thanks! It's quite a change from what it was, but little value in using non-standard indentation.

tarsius:
Piuuuuu.... thanks!!

PR #1574, opened 2020-02-17
Use outline-minor-mode section headings

tarsius:
(Don't merge this just yet, see bottom.)

`outline-minor-mode` is a mode that can be used to deal with sections in source files like one does in Org mode files.  The file is structures into a tree of sections, subsections and so on.  Sections can be collapsed and expanded, and the visibility of complete subtrees or all sections in the buffer can be easily by cycled.  This is useful for much the same reason that it is useful in Org mode buffers, or in Magit buffers.

However this assumes that the source files actually follow Emacs' conventions for section headers in source files, which (phrased in the context of lisp files) goes as follows:

- A comment line that begins with only one or two semicolons is just a plain old comment.
- A comment line that begins with three or more semicolons, followed by a space, followed by some text, is a section heading.
- The number of semicolons specifies the level of the section.  Three=section, four=subsection, five=subsubsection etc.

----

I would like to encourage you not only to merge this pr but to also start using `outline-minor-mode` yourself.  It is immensely useful for navigation (and especially when getting familiar with a unknown codebase, though that does not apply for you in this case.))

Here is my basic configuration, which you might want to adapt:

``` emacs-lisp
(add-hook 'emacs-lisp-mode-hook 'outline-minor-mode)

(use-package bicycle
  :demand t ; https://github.com/jwiegley/use-package/issues/705
  :after outline
  :bind (:map outline-minor-mode-map
              ([C-tab] . bicycle-cycle)
              ([S-tab] . bicycle-cycle-global)))
```

And to make the headings prettier you should also use:

``` emacs-lisp
(use-package outline-minor-faces
  :after outline
  :config (add-hook 'outline-minor-mode-hook
                    'outline-minor-faces-add-font-lock-keywords))

(use-package backline
  :after outline
  :config (advice-add 'outline-flag-region :after 'backline-update))
```

Disclaimer: I am the author of `bicycle`, `outline-minor-faces` and `backline`.

You can try out what this does in the `magit` repository, for example.  It makes heavy use of sections.

----

This converts existing section headings that previously did not conform to the Emacs conventions.

It also replaces section **separators** (things like `;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;`) into section *headings*.  Because I am not familiar with the code base, I often did not know what heading to use.  In such cases I used `;;; UNNAMED` as temporary heading.

You probably don't want to merge this without first adding an additional commit, which replaces these placeholders with the appropriate headings.

### DIFF
--- a/mu4e/mu4e-actions.el
+++ b/mu4e/mu4e-actions.el
@@ -26,6 +26,7 @@
 ;; manual)
 
 ;;; Code:
+
 (require 'cl-lib)
 (require 'ido)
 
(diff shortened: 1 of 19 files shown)


djcb:
Okay, merged locally (with some changed). Thanks!

djcb:
I'll check out those packages.

tarsius:
> Okay, merged locally (with some changed). Thanks!

Thanks for the meaningful headings! :stuck_out_tongue_winking_eye:

PR #1581, opened 2020-02-21
Fix revert buffer

thierryvolpiatto:
This is in continuation of #1562.
This PR allows using revert-buffer in mu4e-main instead of binding "g" to mu4e.
`mu4e~start` is no more using `mu4e-index-updated-hook`, the call is directly done in `mu4e-info-handler` which is cleaner.
Main buffer is correctly updated when quitting headers.
Mode-line is updated when changing context.


djcb:
Thanks -- merged.

PR #1584, opened 2020-02-22
mu4e: Fix quoting in M-q keybinding definition

hpdeifel:
In the keybinding defintions for mu4e-view-mode-map, the variable
mu4e-view-use-gnus was used quoted in an if condition, which means that the
condition always evaluated to true. Therefore, M-q would always be bound to
article-fill-long lines and never to mu4e-view-fill-long-lines.


djcb:

PR #1586, opened 2020-02-24
Add support for displaying maildirs and their counts

seanfarley:
This code could be merged as-is but is mostly a request for comment on this functionality. In a nutshell, this upstreams the functionality of [mu4e-maildirs-extension](https://github.com/agpchil/mu4e-maildirs-extension) using the new query count feature. Displaying maildirs is disabled by default and controlled via `mu4e-main-show-maildirs`.

The last two commits that align both the bookmark and maildir counts is probably overkill but it does look nice.

Missing are documentation, updating the changelog, etc. but I can add those if this feature is deemed worthy to upstream :-)

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -147,6 +147,38 @@ clicked."
                    "\n")))
 
 
+(defun mu4e~main-maildirs ()
+  "Return a string of maildirs with their counts."
+  (cl-loop with mds = (mu4e~maildirs-with-query)
+           with longest = (mu4e~longest-of-maildirs-and-bookmarks)
+           with queries = (plist-get mu4e~server-props :queries)
+           for m in mds
+           for key = (string (plist-get m :key))
+           for name = (plist-get m :name)
+           for query = (plist-get m :query)
+           for qcounts = (and (stringp query)
+                              (cl-loop for q in queries
+                                       when (string= (plist-get q :query) query)
+                                       collect q))
+           concat (concat
+                   ;; menu entry
+                   (mu4e~main-action-str
+                    (concat "\t* [j" key "] " name)
+                    (concat "j" key))
+                   ;; append all/unread numbers, if available.
+                   (if qcounts
+                       (let ((unread (plist-get (car qcounts) :unread))
+                             (count  (plist-get (car qcounts) :count)))
+                         (format
+                          "%s (%s/%s)"
+                          (make-string (- longest (length name)) ? )
+                          (propertize (number-to-string unread)
+                                      'face 'mu4e-header-key-face)
+                          count))
+                     "")
+                   "\n")))
+
+
 (defun mu4e~key-val (key val &optional unit)
   "Return a key / value pair."
   (concat
@@ -197,6 +229,9 @@ When REFRESH is non nil refresh infos from server."
        "\n"
        (propertize "  Bookmarks\n\n" 'face 'mu4e-title-face)
        (mu4e~main-bookmarks)
+       "\n"
+       (propertize "  Maildirs\n\n" 'face 'mu4e-title-face)
+       (mu4e~main-maildirs)
        "\n"
        (propertize "  Misc\n\n" 'face 'mu4e-title-face)
 
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -117,8 +117,7 @@ clicked."
 (defun mu4e~main-bookmarks ()
   ;; TODO: it's a bit uncool to hard-code the "b" shortcut...
   (cl-loop with bmks = (mu4e-bookmarks)
-           with longest = (cl-loop for b in bmks
-                                   maximize (length (plist-get b :name)))
+           with longest = (mu4e~longest-of-maildirs-and-bookmarks)
            with queries = (plist-get mu4e~server-props :queries)
            for bm in bmks
            for key = (string (plist-get bm :key))
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -110,7 +110,14 @@ clicked."
     (define-key map (kbd "RET") func)
     (put-text-property 0 (length newstr) 'keymap map newstr)
     (put-text-property (string-match "\\[.+$" newstr)
-                       (- (length newstr) 1) 'mouse-face 'highlight newstr)
+                       ;; only subtract one from length of newstr if we're
+                       ;; actually consuming the first letter (e.g.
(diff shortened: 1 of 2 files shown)


seanfarley:
Here are some screenshots to show what it all looks like:

Screen Shot 2020-02-24 at 2 49 35 PM

Screen Shot 2020-02-24 at 2 51 53 PM

djcb:
Nice, I like it!

However, at least for now, I'd like to limit this to just the maildirs in `mu4e-maildir-shortcuts`.

(Note that I changed the format of those shortcuts to 
```
   (:maildir <maildir> :key <key>)
```
and optional `:hide` and `:hide-unread`. so very similar to `mu4e-bookmark` (the old format is still supported, you need the function `mu4e-maildir-shortcuts` which does the conversions from the old format; so I think we can handle those in the very same way. We can perhaps even turn them on by default (or only if there are not too many?)

Maybe after that we can add _all_ maildirs, but that may be quite a bit (e.g., I have 88 -- doesn't fit on the screen and I don't want to run 2 * 88 queries). So we'd need some filtering.

Other small thing: you need to quote the maildirs in the queries, like `(format "maildir:\"%s\"" maildir)` (still not perfect, but at least handles the not uncommon case of maildirs with spaces in them).

What do you think?

seanfarley:

> Nice, I like it!

Woohoo, thanks!

> However, at least for now, I'd like to limit this to just the maildirs in `mu4e-maildir-shortcuts`.
>
> (Note that I changed the format of those shortcuts to
> ```
>    (:maildir <maildir> :key <key>)
> ```
> and optional `:hide` and `:hide-unread`. so very similar to `mu4e-bookmark` (the old format is still supported, you need the function `mu4e-maildir-shortcuts` which does the conversions from the old format; so I think we can handle those in the very same way. We can perhaps even turn them on by default (or only if there are not too many?)

Aha! Yeah, that makes sense. I'll update the code to use those.

> Maybe after that we can add _all_ maildirs, but that may be quite a bit (e.g., I have 88 -- doesn't fit on the screen and I don't want to run 2 * 88 queries). So we'd need some filtering.

Oh, yeah, that's a good point. I'm always worried about making these
kind of changes since I think I have a fairly small mailbox and am
unaware of how other people use theirs and what kind of edge cases
exist.

I totally agree about limiting this to `mu4e-maildir-shortcuts` for now;
that will simplify the code a bit about missing keys.

Do you think we need a config variable for turning this off/on? Or will
checking `mu4e-maildir-shortcuts` for nil suffice?

> Other small thing: you need to quote the maildirs in the queries, like `(format "maildir:\"%s\"" maildir)` (still not perfect, but at least handles the not uncommon case of maildirs with spaces in them).

Ah, good catch, thanks!

> What do you think?

All those suggestions make sense for me! Once you let me know about
`mu4e-main-show-maildirs` needing to exist or not, I can also work on
some of the NEWS entries.

seanfarley:
Sean Farley <sean@farley.io> writes:

> Dirk-Jan C. Binnema <notifications@github.com> writes:
>
>> Nice, I like it!
>
> Woohoo, thanks!
>
>> However, at least for now, I'd like to limit this to just the maildirs in `mu4e-maildir-shortcuts`.
>>
>> (Note that I changed the format of those shortcuts to
>> ```
>>    (:maildir <maildir> :key <key>)
>> ```
>> and optional `:hide` and `:hide-unread`. so very similar to `mu4e-bookmark` (the old format is still supported, you need the function `mu4e-maildir-shortcuts` which does the conversions from the old format; so I think we can handle those in the very same way. We can perhaps even turn them on by default (or only if there are not too many?)
>
> Aha! Yeah, that makes sense. I'll update the code to use those.
>
>> Maybe after that we can add _all_ maildirs, but that may be quite a bit (e.g., I have 88 -- doesn't fit on the screen and I don't want to run 2 * 88 queries). So we'd need some filtering.
>
> Oh, yeah, that's a good point. I'm always worried about making these
> kind of changes since I think I have a fairly small mailbox and am
> unaware of how other people use theirs and what kind of edge cases
> exist.
>
> I totally agree about limiting this to `mu4e-maildir-shortcuts` for now;
> that will simplify the code a bit about missing keys.
>
> Do you think we need a config variable for turning this off/on? Or will
> checking `mu4e-maildir-shortcuts` for nil suffice?
>
>> Other small thing: you need to quote the maildirs in the queries, like `(format "maildir:\"%s\"" maildir)` (still not perfect, but at least handles the not uncommon case of maildirs with spaces in them).
>
> Ah, good catch, thanks!
>
>> What do you think?
>
> All those suggestions make sense for me! Once you let me know about
> `mu4e-main-show-maildirs` needing to exist or not, I can also work on
> some of the NEWS entries.

Oh, one more question: what do you think about the logic to align both
the bookmarks unread count and the maildirs unread count? Keep it or is
it too much?

djcb:
Oh, great!

As for a configuration variable...  good question. For now, I guess we can try without it, and let people turn of individual shortcuts with `:hide t` if they need to. If that turns out to be too annoying for too many people we can think of something.

Aligning the unread-count for bookmars and maildirs is a nice touch, yeah!

seanfarley:
👍

seanfarley:
Ok, I finally got around to updating this PR 😬 Hopefully, I addressed all the feedback!

There was one small design decision I made: in the main view there is now a "Maildirs" section below "Bookmarks" which will be empty if the user does not set any maildir shortcuts. I modeled this after the bookmarks logic and hope that it won't be a problem.

Sorry for the delay!

djcb:
Great! I'll look at this post 1.4.x, hopefully soon.

scturtle:
This does not work well with CJK maildir names.
There is encoding problem in `query` field of `mu4e~server-props`.
E.g. `(decode-coding-string (apply #'unibyte-string (string-to-list (plist-get q :query))) 'utf-8)` is need to make it a correct UTF-8 string.

seanfarley:
Ah, thanks for the info!

seanfarley:
Ah, this is also a problem in `bookmarks` too. Trying to fix now.

seanfarley:
@djcb I put some fixes to the bookmarks section (an off-by-one error for the text property of mouse highlighting). Let me know if you want that in its own pull request separate from this one.

Also, @scturtle, I found it difficult to align the text of non-ASCII characters but that's due to my font. Here is a screenshot of my fix:

Screen Shot 2020-05-12 at 10 22 23 PM

djcb:
I've merged this locally, with some tiny changes. Thanks for contributing!

Issue #1600, opened 2020-02-29
Improved support for format=flowed


codesections:
I have recently been attempting to use mu4e to send and read emails with `Content-Type: text/plain; format=flowed` as described in [the FAQ](https://www.djcbsoftware.nl/code/mu/mu4e/Writing-messages.html).  Unfortunately, I have run into some issues.

## Issues reading format=flowed emails

By default, mu4e does not reflow emails.  The [RFC](https://www.ietf.org/rfc/rfc2646.txt) states that

> A series of one or more flowed lines followed by one fixed line is considered a paragraph, and MAY be flowed (wrapped and unwrapped) as  appropriate on display.

Despite the permissive language, this feature seems worth implementing if mu4e is interested in supporting `format=flowed` – my understanding is that the whole point of `format=flowed` is to support receiver reformatting, so leaving it out seems like a large omission.  Additionally, [at least some](https://vxlabs.com/2019/08/25/format-flowed-with-long-lines/) mu4e users intentionally send `format=flowed` emails with especially long lines on the expectation that receiving clients will flow them; not reflowing these emails is likely to create an especially bad reading experience.  

Additionally, both Mutt and Gnus reflow`format=flowed` emails by default (Mutt with the `reflow_text` and `reflow_wrap` properties and Gnus with `fill-flowed-display-column`).

### Potential fixes

Two simple workarounds exist.  First, the `mu4e-view-fill-long-lines` function lets users achieve something similar to reflowing text (although it ties the size of the text to `fill-column` – which is set for writing and thus may be smaller than ideal for reading to allow room for quoting.)

Second, users can switch to the experimental gnus viewer mode (with `(setq mu4e-view-use-gnus t)`).  As I mentioned, gnus already reflows `format=flowed` messages, which resolves the issue.

A longer-term fix would probably involve reflowing text based on the `fill-flowed-display-column` in the main mu4e viewer.  This seems especially appropriate given that mu4e already uses the similar `fill-flowed-encode-column` when sending `format=flowed` emails.

## Problems sending format=flowed emails

The problem with sending `format=flowed` emails arises is explained in the FAQ:

> The transformation of your message into the proper format is done at the time of sending. For this to happen properly, you should write each paragraph of your message of as a long line (i.e. without carriage return).

[...]

PR #1604, opened 2020-03-01
Improve support for format=flowed

codesections:
This PR revises the way mu4e handles format=flowed messages as discussed in #1600.  

This new code removes the previous requirement that users write each paragraph of their code as a single long line, and allows users to use `auto-fill-mode`, `fill-paragraph`, or any other function they might otherwise use to manage line length when writing their messages.

This change also removes the (somewhat confusing) distinction between `mu4e-compose-format-flowed` and `use-hard-newlines`.  Under this code, the user simply sets `mu4e-compose-format-flowed` whenever they want to use the `format=flowed` header and then controls the length of their lines with `fill-flow-encode-column`. 

Please see the individual commit messages for further explanation of implementation details and changes.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -198,9 +198,8 @@ This is a symbol, `new', `forward', `reply' or `edit'.")
 
 (defcustom mu4e-compose-format-flowed nil
   "Whether to compose messages to be sent as format=flowed.
-\(Or with long lines if variable `use-hard-newlines' is set to
-nil). The variable `fill-flowed-encode-column' lets you customize
-the width beyond which format=flowed lines are wrapped."
+The variable `fill-flowed-encode-column' lets you customize the
+width beyond which format=flowed lines are wrapped."
   :type 'boolean
   :safe 'booleanp
   :group 'mu4e-compose)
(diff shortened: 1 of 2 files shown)


mu4e/mu4e-compose.el
```diff
@@ -386,33 +457,13 @@ removing the In-Reply-To header."
           (define-key map (kbd "C-S-u")   'mu4e-update-mail-and-index)
           (define-key map (kbd "C-c C-u") 'mu4e-update-mail-and-index)
           (define-key map (kbd "C-c C-k") 'mu4e-message-kill-buffer)
-          (define-key map (kbd "M-q")     'mu4e-fill-paragraph)
+          (if mu4e-compose-format-flowed
+              (define-key map (kbd "C-c C-c" 'mu4e-fill-send-and-exit)))
```
djcb: Hmm, this would make things depend on the value of `mu4e-compose-format-flowed` the first time you use it (with no obvious way to change it later) - or am i misreading?

Also, could we keep `M-q` for filling too? That has the advantage that you can see the result.


Overall, I think what'd I'd prefer is having `mu4e-compose-fill-policy` defvar, with values nil, 'always, 'ask, and do the filling in `send-mode-hook`. 

What do you think?

codesections: > Overall, I think what'd I'd prefer is having mu4e-compose-fill-policy defvar, with values nil, 'always, 'ask, and do the filling in send-mode-hook.

We can definitely do that.  Here's my use-case, though: I would like to send ~90% of my emails with a format=flowed line length of 72, since that follows the spec and is the only one *guaranteed* not to display obnoxiously long lines in terminal clients.  However, for about 10% of my emails (primarily ones I send specifically to non-technical recipients/when I know they will be viewed on a cell phone) I'd like to set line length to 998 to take advantage of the [line-wrapping behavior](https://vxlabs.com/2019/08/25/format-flowed-with-long-lines/) of clients that don't implement format=flowed.

Given that use case, I wouldn't want mu4e to ask about line length _every_ time, since it's the same 90% of the time.  But I also would like a way to get it to ask when I _do_ want to set the length.  Accepting a prefix argument works very well for that.

I don't want to insist on implementing my use case or anything – I can always accommodate it my own `.init` file.  But I thought I'd mention it, since it seems (to me at least) like others might also like being able to change the behavior via a prefix argument.

How about this modification to your suggestion: bind `C-c C-c` to a new function `mu4e-send-and-exit`.  Without a prefix argument (or when`mu4e-compose-format-flowed` is nil), that function calls `message-send-and-exit`, but when called with a prefix argument (with `mu4e-compose-format-flowed` non nil), it prompts for a line length.  That keeps the prefix argument support that strikes me as useful, but also lets us use the `mu4e-compose-fill-policy` var and the `message-send-hook`.  (I assume that's what you meant by `send-mode-hook`, but let me know if there's a separate `send-mode-hook` I'm missing.)

Or, if you prefer, I'm happy to write it as you suggested and implement my own workflow in my `init` file.

> Also, could we keep M-q for filling too? That has the advantage that you can see the result.

That's a good point.  The reason I didn't bind `M-q` to `mu4e-fill-buffer-for-flowed` was to avoid shadowing `fill-paragraph`.  That wasn't a problem before – in fact, it was _good_ to shadow `fill-paragraph` when users were required to write message paragraphs as a single line, since using the regular `fill-paragraph` would break that.  But under this PR, users might use `fill-paragraph` to make their draft formatting better.

What if we bind `mu4e-fill-buffer-for-flowed` to `C-c M-q` and leave `M-q` free for `fill-paragraph`?

codesections:
After looking more into both the mu4e code and the underlying code in `flow-fill.el` and `mml.el`, I have decided that this PR is on the wrong track.  I am closing it and will open a different one that achieves the same functionality using built-in Emacs functionality rather than reinventing the wheel here.

PR #1605, opened 2020-03-04
Fix of the attachment actions

yellowmoneybank:
With the current version of the mu server, it is not possible open
attachments, because the server expects symbols as parameters, but
mu4e sends strings.
This change fixes this.


djcb:
Oh, good catch, merged. Thanks!

PR #1606, opened 2020-03-04
Added option to attachment-options to import gpg public key

yellowmoneybank:
This pull request adds an option to the attachment-options that allows
to import pgp-public-keys.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -1457,8 +1457,11 @@ If PIPECMD is nil, ask user for it."
   "Import MSG's attachment ATTACHNUM into the gpg-keyring."
   (interactive)
   (let* ((att (mu4e~view-get-attach msg attachnum))
-         (index (plist-get att :index)))
-    (mu4e~view-temp-action (mu4e-message-field msg :docid) index 'gpg)))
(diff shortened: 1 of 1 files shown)


mu4e/mu4e-view.el
```diff
@@ -1452,6 +1453,13 @@ If PIPECMD is nil, ask user for it."
          (index (plist-get att :index)))
     (mu4e~view-temp-action (mu4e-message-field msg :docid) index "diary")))
 
+(defun mu4e-view-import-public-key (msg attachnum)
+  "Import MSG's attachment ATTACHNUM into the gpg-keyring."
+  (interactive)
+  (let* ((att (mu4e~view-get-attach msg attachnum))
+         (index (plist-get att :index)))
+    (mu4e~view-temp-action (mu4e-message-field msg :docid) index 'gpg)))
```
djcb: Would be good to add a mime-type check here, so we can give an error unless the attachment is a public key. ie with `(plist-get att :mime-type)`

phikal: Won't epa throw an error if the attachment contents isn't a proper public key?

djcb:
Merged locally (with some minor changes). Thanks!

PR #1610, opened 2020-03-07
Merged crypto-policy configuration into one variable

phikal:
The current cryptography policy is a bit confusing.  The existence of the two variables `mu4e-compose-crypto-reply-plain-policy` and `mu4e-compose-crypto-reply-encrypted-policy` at first glance implies that you can configure how to respond to encrypted and plain emails, but on closer inspection `mu4e-compose-crypto-reply-plain-policy` reveals itself to also be responsible for forwarding mails, creating new messages, etc.

This changeset suggests merging the two variables into one -- `mu4e-compose-crypto-policy` -- while at the same time expanding it's granularity. So instead of implementing `mu4e-compose-crypto-forward-plain-policy`, `mu4e-compose-crypto-new-policy`, etc. everything regarding encryption and signing is centrally controlled by a customizable set.

To avoid conflicts with older configurations, I have added a translation layer in `mu4e~compose-handler`, that adds the corresponding new symbols to the new policy variable, based on the old the values of the old variables. In case these weren't customized, both should be nil now, meaning that the append has no consing overhead (be it ever so minimal), as `mu4e-compose-crypto-reply-encrypted-policy` has been changed from `sign-and-encrypt` to nil. 

To preserve the current defaults, `mu4e-compose-crypto-policy` only contains the symbols that were implied by `mu4e-compose-crypto-reply-encrypted-policy`. It might be worth considering expanding the default configuration, but I didn't want to make that decision in this pull request.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -137,7 +137,49 @@ Also see `mu4e-context-policy'."
   :safe 'symbolp
   :group 'mu4e-compose)
 
-(defcustom mu4e-compose-crypto-reply-encrypted-policy 'sign-and-encrypt
+(defcustom mu4e-compose-crypto-policy
+  '(encrypt-encrypted-replies sign-encrypted-replies)
+  "Policy to control when messages will be signed/encrypted.
+
+The value is a list, whose members determine the behaviour of
+`mu4e~compose-crypto-message'. Specifically, it might contain:
+
+- `sign-all-messages': Always add a signature.
+- `sign-new-messages': Add a signature to new message, ie.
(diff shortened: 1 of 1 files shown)


djcb:
At a glance, looks okay. But this will be something for post-1.4, as we are in deep-freeze mode.

phikal:
No problem. Just out of curiosity, is there a planned release date?

djcb:
Regarding the release: next few weeks, I think. There are still a few issues we're tracking.

djcb:
Merged locally & pushed. Thanks!

PR #1612, opened 2020-03-09
Format flowed

codesections:
This is a second approach to solving the format=flowed issue noted in #1600.  Unlike the first approach (#1604), this PR leverages the functions in fill-flowed.el and `message-mode` – as a result, this PR introduces less code and, hopefully, should make future maintenance easier.

For full documentation of the new format=flowed support, see the added `*info*` documentation.  But the short version is that `mu4e-compose-format-flowed now accepts four non-nil values

* `'manual` – sets the `f=f` header, but doesn't reformat lines
* `'auto` – sets the header and refills paragraphs based `fill-flowed-encode-column` 
* `'auto-long-lines` – sets the head, refills paragraphs, and adds a space to one-line paragraphs
* `t` – same as `'auto` (for backwards compatibility only)


djcb:
Thanks... overall looks good... I'm not going to pull this know since we're in deep feature-freeze for the 1.4 release... but I'll get to this after that.

mu4e/mu4e-compose.el
```diff
@@ -385,34 +396,93 @@ removing the In-Reply-To header."
...
+    (end-of-line 2)))
+
+(defun mu4e-compose-send-and-exit ()
+  "Send the current message and exit the buffer."
+  (interactive)
+  (mu4e-compose-ensure-final-newline)
+  ;; We have to make sure the message ends in a newline *before* calling
+  ;; message-send-and-exit (even though we check again later) to avoid
+  ;; message-send-and-exit messing with the formatting of the final paragraph
+  (advice-add #'fill-flowed-encode :around #'mu4e-compose-fill-flowed)
```
djcb, line 483: Can't we do this without our own functions?

mu4e/mu4e-compose.el
```diff
@@ -572,8 +650,7 @@ we can decide what we want to do."
       (sign (mml-secure-message-sign))
       (encrypt (mml-secure-message-encrypt))
       (sign-and-encrypt (mml-secure-message-sign-encrypt))
-      (message "Do nothing")))
-  )
+      (message "Do nothing"))))
```
djcb, line 653: Let's get rid of that message (it's old debugging stuff I think)

djcb:
Seems this has gotten stale, so let's close for now.

PR #1614, opened 2020-03-12
Adding support to query the headers of a message 

yellowmoneybank:
This pull-request adds support to search for headers in a message, specified
by its docid. This is necessary, if you want to add support for
protocols, that rely on adding header fields to mails (for example autocrypt).

The mu server is extended by the command "(header :docid <docid>
:header <header>)", and returns a message of the format "(:header
<header-fild> :docid <docid>  :value <header-field-value>)".

mu4e is extended by the function (defun mu4e~proc-header (docid
header) ...), which scans the message with the DOCID for a
HEADER-field.
The result of that function is delivered to the function registered as
"mu4e-search-header-field-func".

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -72,6 +72,9 @@
       "My bookmark" :query "banana OR pear" :key ?f)~ this format is a bit easier
       extensible.
 
+    - Same is true for ~mu4e-maildir-shortcuts~. The older format is still
+      recognized.
+
     - Improved handling of replies to mailing-lists.
 
     - For instance, we now recognize an attribute ~:hide t~, which will hide the
@@ -85,6 +88,15 @@
     - The mu4e <-> mu has been rewritten to communicate using s-expressions,
       with a repl for testing. If you have readline installed, you also get
       history etc.
+
+    - mu4e now defaults to _not_ doing a cleanup after indexing; see
+      ~mu4e-index-cleanup~; reason is that some users have slow (file)systems,
+      making the index operation unnecessary lengthy. More over, files that do not exist
+      are excluded during querying.
+
+      It is recommended to occasionally still clean up the database, perhaps by
+      closing mu4e and calling ~mu index~ from the command-line; this is esp. true
+      if other programs are working in your maildir, behind mu's back.
 
 *** toys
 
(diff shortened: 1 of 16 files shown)


djcb:
Hmm... appreciate the work, but not sure about this one. 

I'm not convinced `autocrypt` is worth supporting at this point. If `mutt` is the most prominent client supporting (with all due respect, I'm an old mutt user), it seems rather marginal -- not even `gnus` supports it. So the extra complexity seems unwanted.

Now, if we _do_ want to support it, it's probably better to add a few fields to the message sexp, using gmime's support. WDYT?

phikal:
Actually, the most prominent client is Thunderbird. @yellowmoneybank and I have been working on a MUA-generic autocrypt implementation (the MELPA pull request is [here](https://github.com/melpa/melpa/pull/6767)), and while I managed to circumvent the need for this specific patch, I do feel that it would be more efficient than what [I came up with](https://git.sr.ht/~zge/autocrypt/tree/master/autocrypt.el#L502).

Furthermore, it's not just Autocrypt. There are certainly many more headers that add-ons could make use of if there were a sensible way to query them.

---

Also, GMIME has some difficulty implementing Autocrypt, since to my knowledge it isn't responsible for sending or has the ability to interject while composing messages? If this isn't given, it's rather difficult to add the `Autocrypt` header or the `Autocrypt-Gossip` attachment that is used for multiple recipients.

djcb:
Ah, there's a 3rd-party extension for Thunderbird. Still I'd say it's rather marginal, but of course you have to start somewhere.

Anyway, I'm not in favor for querying of headers like this; the mu<->mu4e protocol is not designed for this, and it's not something I want to expose, either for autocrypt or other use-cases.

For the GMIME part, the idea would that in the sexps we send to mu4e when showing a full message, (mu-msg-sexp), we'd add the autocrypt header(s). I.e., you'd still get the same headers as in the 'query' case, but in a way that better fits mu4e.

djcb:
Thanks for the effort, but this doesn't apply anymore, and anyway, I wasn't quite happy with the approach (as discussed above). So closing this...

PR #1622, opened 2020-03-20
mu4e: extend mu4e-header-highlight-face to edge of window

tarsius:
Due to a breaking change we have to request this explicitly in
Emacs 27.  Earlier Emacs versions do the right thing by default.

If the effect of a face does not extend to the edge of the window,
then attributes such as the background color or underline cease to
be used for more that the width of a single character, which is not
appropriate for a face intended to highlight a complete line.

See https://debbugs.gnu.org/cgi/bugreport.cgi?bug=37774.


djcb:
Merged, thanks.

PR #1627, opened 2020-03-29
Misc fixes

haji-ali:
This fixes what I believe to be two bugs:

- The first is in `mu4e-info-handler` where `save-window-excursion` is now called instead of just selecting the `mainbuf` window. Previously, the main-mu4e window was selected (if visible) after `mu4e-update-index` is called (directly or through the update timer). The result was that the pointer would jump away from a currently selected window. 

- The second is in `mu4e-compose.el`. Previously, `(widen)` was called right after `(mu4e~compose-hide-headers)`. There were two problems: First the headers that were just hidden are immediately shown. Second if some headers were previously hidden before that hook is executed, some header fields were corrupted.


djcb:
Merged -- thanks!

Issue #1625, opened 2020-03-22
Message with inline ICS


Chris00:
## Expected or desired behavior

View the message

## Actual behavior

The [attached message](https://github.com/djcb/mu/files/4366483/test-ics.txt) (composed with Mu4e) triggers an error when displaying it with the Gnus mode (not the original one).
(I can send _privately_ the original message which also starts the debugger when “Enter Debugger on Error” is set.)

## Versions of mu, mu4e/emacs, operating system etc.

Emacs 26.3
Mu: commit 441f0053

Chris00:
Same with the latest Mu version (commit efd67102).

Chris00:
The problem is with `gnus-icalendar-with-decoded-handle`: it does
```elisp
  (let ((charset (make-symbol "charset")))
    `(let ((,charset (cdr (assoc 'charset (mm-handle-type ,handle)))))
```
and this returns an invalid charset.

djcb:
Any idea how to fix this?

Chris00:
On 25 March 2020 at 07:47 CET, Dirk-Jan C. Binnema <notifications@github.com> wrote:
>
> Any idea how to fix this?

I'll try to find some time to investigate in the coming days.  I see two possibilities: one could try to add the charset information to the handle (we try to determine it in `mu4e~view-gnus`) or we could advise the function (using `mu4e~view-rendering` to only trigger it for mu4e).

Chris00:
Temporary fix (a better solution must be found to use the charset in `mu4e~view-gnus`).
```elisp
(define-advice gnus-icalendar-event-from-handle
    (:filter-args (handle-attendee) fix-missing-charset)
  "Fix mu not passing the charset."
  (let* ((handle (car handle-attendee))
         (attendee (cdr handle-attendee))
         (buf (mm-handle-buffer handle))
         (ty (mm-handle-type handle))
         (rest (cddr handle)))
    ;; Put the fallback at the end:
    (setq ty (append ty '((charset . "utf-8"))))
    (setq handle (cons buf (cons ty rest)))
    (list handle attendee)))
```

djcb:
Ah, thanks. But should the advise perhaps have a guard so it only works when used in mu4e? Otherwise it could affect people trying to use the 'real' gnus after this is defined. Or?

Chris00:
I filed a bug against Emacs.  The following variant only apply the customization inside mu:
```elisp
(define-advice gnus-icalendar-event-from-handle
[...]

PR #1633, opened 2020-04-05
Fix displaying ical attachments with no charset

Chris00:
Fixes https://github.com/djcb/mu/issues/1625


djcb:
Merged, thanks!

alexmurray:
Since this was merged I am getting errors using `mu4e-view-use-gnus t` when displaying ical invitations using emacs-27 - I think the error is due to the use of `cdr` to get `attendee` from `handle-attendee` - this should instead be `cadr` as otherwise `cdr` returns `(list attendee)` rather than just `attendee`

PR #1637, opened 2020-04-08
Fix autoreconf

rotty:
Add an apparently missing m4 file from gnulib, and improve the .gitignore state of the m4/ subdirectory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,14 @@ install-sh
 depcomp
 libtool
 ltmain.sh
-m4/*
+
+# Added automatically by `autoreconf`
+m4/libtool.m4
+m4/ltoptions.m4
+m4/ltsugar.m4
+m4/ltversion.m4
+m4/lt~obsolete.m4
+
 missing
 nohup.out
 vgdump
--- a/m4/host-cpu-c-abi.m4
+++ b/m4/host-cpu-c-abi.m4
@@ -1,0 +1,675 @@
+# host-cpu-c-abi.m4 serial 13
+dnl Copyright (C) 2002-2020 Free Software Foundation, Inc.
+dnl This file is free software; the Free Software Foundation
+dnl gives unlimited permission to copy and/or distribute it,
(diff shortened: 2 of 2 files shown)


djcb:
What error did you get without this macro?

rotty:
Quoting from the commit message of the commit adding the file:

>  Add m4/host-cpu-c-abi.m4
> 
> The added file is taken from gnulib e313a53; adding it fixes the
> following warning:
> 
>     configure.ac:225: warning: gl_HOST_CPU_C_ABI_32BIT is m4_require'd but not m4_defun'd
> 
> The missing file also caused the generated `configure` script to
> contain an reference to `gl_HOST_CPU_C_ABI_32BIT`, instead of its
> expansion.

djcb:
Ah, but I wondered why I've never seen that warning... is that specific to 32-bit systems?

rotty:
This is on a 64-bit system. It happens also when building mu with the guix package manager (which uses the same autoconf/automake versions as Debian testing). The full trace is:

    configure.ac:225: warning: gl_HOST_CPU_C_ABI_32BIT is m4_require'd but not m4_defun'd
    m4/lib-prefix.m4:155: AC_LIB_PREPARE_MULTILIB is expanded from...
    m4/lib-link.m4:705: AC_LIB_LINKFLAGS_FROM_LIBS is expanded from...
    m4/guile-2.2.m4:132: GUILE_FLAGS is expanded from...
    ../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
    ../../lib/m4sugar/m4sh.m4:639: AS_IF is expanded from...
    configure.ac:225: the top level

So this seems something that is a side-effect of including `guile-2.2.m4`; conversely, I'm wondering why it doesn't happen on your system ;-), and that I appear to be the first one noticing.

djcb:
Ah, thanks. Probably some m4 file that I happen to already have on my system

djcb:
Merged, thanks!

PR #1646, opened 2020-04-19
~/.mu -> ~/.cache/mu

dschrempf:
Change man pages so that they honor the new mu config directory.

Question: `mu-script.1` Where are the scripts stored? I guess not in the `.cache` directory. I left `.mu/scripts`.


djcb:
Merged -- thanks! Good point about the guie scripts.... let me clear that up.

Issue #1657, opened 2020-04-22
Sort Order Reversed Within Threads


jterk:
## Expected or desired behavior

I typically sort mail in mu4e in chronological order - oldest at the top. My expectation is that this ordering will hold within a thread. I.e. (TN denotes relative timestamps):

* [T1] - Some Other Subject
* [T0] - Initial Email
  * [T2] - Re: Initial Email
  * [T3] - Re: Initial Email
* [T4] Yet Another Subject

## Actual behavior

Since updating to mu/mu4e 1.4.1 I observe the following sorting behavior, instead. Note that the thread as a whole is sorted correctly relative to other messages, but the messages within the thread are in the opposite order of what is expected (the most recent message in the thread is first):

* [T1] - Some Other Subject
  * [T3] - Re: Initial Email
  * [T2] - Re: Initial Email
* [T0] - Initial Email
* [T4] Yet Another Subject

This screenshot demonstrates the behavior. The thread beginning on 2019-07-25 is sorted correctly (oldest-to-newest) relative to the emails before (most recently updated on 2019-06-27, not visible here)  and after (2020-02-06), but the items within the thread are sorted newest-to-oldest

![Screen Shot 2020-04-22 at 09 51 43](https://user-images.githubusercontent.com/4620/80010901-f5562d00-847f-11ea-9595-8d7a91dfe1f4.png)

## Steps to reproduce

Receive some email, including some that should be threaded. In header view, sort by date (`O d`), toggle threading, if necessary (`P`).

## Versions of mu, mu4e/emacs, operating system etc.

* mu/mu4e: 1.4.1
* emacs: GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.5.0, NS appkit-1671.40 Version 10.14.4 (Build 18E226)) of 2019-08-30
* OS: Mac OS X 10.15.4

djcb:
The order has changed indeed, and this is per design (e.g. series of patches are easier to read chronologically); after some discussion on the mailing-list (see "Changing the way threads are sorted") it seemed that most people preferred the new order.

jterk:
I see - I find it very surprising that the sorting within a thread is exactly the opposite of the sort order I choose.

czan:
[...]

PR #1663, opened 2020-04-29
mu: always sort descending within threads

czan:
My fix for https://github.com/djcb/mu/issues/1657


djcb:
Okay, tested this, looks good (and, in fact, a fix) -- so merging. Thanks!

PR #1666, opened 2020-04-30
mu4e-proc: restore rename behaviour in mu4e~proc-move

alexmurray:
This appears to have been inadvertently changed in
a5dccffcdfa3aeb1908832bf467f90db82b7f82c which results in errors as noted
in https://groups.google.com/d/msg/mu-discuss/iLR_9Cw14Ps/EdA2pctoBQAJ


djcb:
Thanks! Solved (but in a slightly different way).

PR #1672, opened 2020-05-01
mu-store: enlarge buffer for strftime

krietvel:
On Debian 9 random memory was printed with the original buffer size
and valgrind complained about the use of uninitialized memory.


djcb:
Merged.. bedankt.

The second parameter to strftime should ensure we don't write outside the buffer, but if it's too small, the contents are "undefined" according to the man page.  Anyway, 64 chars should be plenty.

PR #1673, opened 2020-05-04
Is it :hide-unread or :show-unread

sje30:
I think the docstring is correct and the variable setting is wrong -- the code elsewhere  (mu4e~start) seems to check hide-unread property.

Once the discrepancy between doc string and code is fixed, the info file will need updating, as this variable is shown verbatim in 7.2 Bookmarks node.


djcb:
Merged, thanks. Some left-overs from an earlier version.

PR #1683, opened 2020-05-08
mu4e: change spacing in main view for info section

manuel-arguelles:
Change the spaces to tab to be consistent with the other entries in the
main view.


djcb:
Merged -- thanks!

PR #1685, opened 2020-05-09
Make mu4e~view-message available to mu4e-view-mode-hooks.

derek-upham:
In the current `master` branch, setting up an `m4ue-view` buffer runs the `mu4e-view-mode-hooks` entries on a buffer that is only partway through initialization: the buffer is missing the `mu4e~view-message` buffer-local variable with the actual message.  The code sets that variable after entering `mu4e-view-mode`.  This ordering means the current message is unavailable to the hooks, and the hooks can't adjust their behavior based on the message properties.

To fix that bug, it is not enough to just set `mu4e~view-message` before entering `mu4e-view-mode`.  Major mode changes clobber buffer-local variables, as part of standard Emacs behavior ([ref](https://www.gnu.org/software/emacs/manual/html_node/elisp/Creating-Buffer_002dLocal.html#Creating-Buffer_002dLocal)).  Marking the `mu4e~view-message` variable as `permanent-local` makes that variable persists across the major mode change, giving us the behavior we want.  (Persistance is a global property of `mu4e~view-message`, so this alters the buffer-local behavior of `mu4e~view-message` in all buffers and across all major mode changes.  Given how tightly coupled the variable and the major mode are, this is unlikely to affect anyone.)

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -194,8 +194,14 @@ The first letter of NAME is used as a shortcut character."
 
 ;;; Variables
 
+;; It's useful to have the current view message available to
+;; `mu4e-view-mode-hooks' functions, and we set up this variable
+;; before calling `mu4e-view-mode'.  However, changing the major mode
+;; clobbers any local variables.  Work around that by declaring the
+;; variable permanent-local.
 (defvar-local mu4e~view-message nil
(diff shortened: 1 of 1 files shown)


djcb:
Oh, that's a useful change; merged. Thanks!

PR #1686, opened 2020-05-10
Fix mu4e-headers-mark-or-move-to-trash()

gspe:
Fix mu4e-headers-mark-or-move-to-trash() to pick right trash folder when
:no-trash-flag option is used in a context.


djcb:
Great -- thank you! Merged.

PR #1690, opened 2020-05-11
mu4e: fix trashed messages face

gspe:
This fix the look and feel of trashed messages:
- messages moved to trash folder with trashed flag
- messages moved to trash folder but without trashed flag

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -703,8 +703,12 @@ found."
 (defun mu4e~headers-line-apply-flag-face (msg line)
   "Adjust LINE's face property based on FLAGS."
   (let* ((flags (mu4e-message-field msg :flags))
+         (msg-dir (mu4e-message-field msg :maildir))
+         (trash-dir (mu4e-get-trash-folder msg))
          (face (cond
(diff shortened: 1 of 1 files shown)


djcb:
I can see whyyou might  want that, but no -- the face is a reflection of the 'trashed' 'T' flag on messages, regardless what the current value for the trash dir is; and I'd like to keep it that way. If somehow messages ended up in the trash-dir without that flag, I want to make that obvious. (Same is true for e.g. the Replied flag and the Sent-dir).

So have to respectfully decline this.

gspe:
Ok I see your point.
What do you think, instead, to use a different face for messages in trash dir?

I ask because I have to use the `:no-trash-flag` for my email provider, fastmail, without this flag trashed email will be deleted directely on IMAP and with this flag, in my today bookmark trashed email are undistinguished from seen one...

tsdh:
@gspe  Hi Guiseppe, I also use fastmail with mbsync and mu4e and have the problem that trashed mails are deleted.  Therefore, I've removed the `+T` flag from the `trash` entry in `mu4e-marks`.  That's obviously a bad workaround given that it'll break again on every update.  So I wonder what this mysterious `:no-trash-flag` is?  I couldn't find it in the code...

PR #1694, opened 2020-05-14
Fix #1687: Respect `no-trash-flag` when trashing email

### LINKED ISSUE #1687: "mark for trash" key should bind to `mu4e-headers-mark-or-move-to-trash`?

Select a message, mark it with <kbd>d</kbd>, if the current context has `no-trash-flag`, the message should be moved to trash folder without flagging.
The same for when using <kbd>backspace</kbd>.

## Actual behavior

Only <kbd>backspace</kbd> works as expected, marking a message <kbd>d</kbd> would call directly `mu4e-headers-mark-for-trash` and thus ignore `no-trash-flag` of the current context.

`mu` version `1.4.4`.

## Any other detail

I am not sure if it is intentional, otherwise I think this line https://github.com/djcb/mu/blob/master/mu4e/mu4e-headers.el#L983 should be 
```
'("Mark for trash" .  mu4e-headers-mark-or-move-to-trash))
```

vqhuy:
Fix #1687.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -980,7 +980,7 @@ Also see `mu4e-view-mark-or-move-to-trash'."
             (define-key menumap [mark-untrash]
               '("Mark for untrash" .  mu4e-headers-mark-for-untrash))
             (define-key menumap [mark-trash]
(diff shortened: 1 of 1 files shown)


djcb:
Thanks for the effort -- fixed it in a slightly different way (some more cases needed to be handled)

PR #1696, opened 2020-05-15
mu-server: construct contact address from name and email

remvee:
Use `name` and `email` to construct a contact address to avoid bad
`full_address` members.  This change also makes the mu4e address
completion more compatible with the output of `mu cfind`.

### DIFF
--- a/mu/mu-cmd-server.cc
+++ b/mu/mu-cmd-server.cc
@@ -401,8 +401,15 @@ contacts_handler (Context& context, const Parameters& params)
                 if (after > ci.last_seen)
                         return;
 
+                /* using name and email yields prettier results than full_address */
+                auto full_email_address{std::string{ci.full_address}};
+                if (!ci.name.empty()) {
+                        const auto address{std::string{ci.name} + " <" + std::string{ci.email} + ">"};
+                        full_email_address = address;
+                }
+
                 Sexp::List contact;
-                contact.add_prop(":address", Sexp::make_string(ci.full_address));
+                contact.add_prop(":address", Sexp::make_string(full_email_address));
                 contact.add_prop(":rank",    Sexp::make_number(rank));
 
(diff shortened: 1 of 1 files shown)


remvee:
More explanation: some addresses in my mailbox have a lot of variants and some of them are bad.  The annoying case which comes up a lot is where some e-mail was sent to or from `<test@example.com>` and others like `Test <test@example.com>`, which is obviously the same address.  This may cause mu4e auto complete to yield `test@example.com` instead of the preferred `Test <test@example.com>`.

I did manage to reproduce this in a test case in `test-mu-contacts.cc` but I am not comfortable enough with C++ to fix this problem on a more fundamental level.

djcb:
`full_email` already contains the name (if available), so it's a bit unclear to me what this changes. Can you explain? Thanks.

remvee:
First I noticed the autocompleted addresses in mu4e were a bit off sometimes (after moving my configuration to another machine and reindexing my maildir).  So I investigated and was very surprised `mu cfind` got me better results than whatever `mu4e` automcomplete got me.  It seems, my c++ experience is close to zero, the indexer code populates that `full_email` with whatever it sees first for the given email address which in some cases may be some empty or some horrible application generated thing.  Since `cfind` doesn't use `full_email` at all and it yields better results (at least for me), I altered `mu server` to do the same as `mu cfind`

remvee:
Rebased to master.

jpggithub:
Clearly, it's a very annoying bug that occurs everyday for several contacts.
I'm not a programmer so I can't help so much here but thanks for that PR!

djcb:
Thanks for the effort, but this commit doesn't apply anymore, and in any case it was at best papering over an issue; closing this. If you still see issues with 1.5.x, please file new issues -- thanks!

PR #1700, opened 2020-05-18
Strip properties from mu4e-completing-read-function results.

derek-upham:
Some completion engines (like "flx") decorate the strings that they
return.  If MU4E passes such a string down to MU, the "format" call
preserves the text properties in the generated S-expression, producing
an invalid query.  MU4E itself has no interest in those decorations,
so strip them out as early as possible from all prompts that use
mu4e-completing-read-function.


djcb:
Makes sense. Merged -- thanks.

PR #1701, opened 2020-05-19
Missing space between sig verification and signers

mrvdb:
Resolution for a tiny annoyance:  adds a space between the verification status of the signature and the signers


djcb:
Makes sense. Merged -- thanks!

Issue #1698, opened 2020-05-17
browse-url-mail does not write body


Chris00:
## Expected or desired behavior

`(browse-url-mail "mailto:test@mu4e.org?subject=SUBJECT&body=MAIN TEXT")` should write “MAIN TEXT” in the message (with `(setq mail-user-agent 'mu4e-user-agent)`).

## Actual behavior

“MAIN TEXT” is written in the current window: `*scratch*` if you execute it from there but usually in the `*Article*` buffer — which is read only, so an error is triggered.

## Versions of mu, mu4e/emacs, operating system etc.

Mu4e 1.4.6 (commit 45bcad1f).

Chris00:
The problem is with the line

https://github.com/djcb/mu/blob/c283cd173d2c5e342d8540d374a838fadc5cdad9/mu4e/mu4e-compose.el#L896

If I understand well, `mu4e~start` makes the call asynchronous so the current buffer keeps being the current one until `*draft*` is ready.  (Switching to `*draft*` before the call does not help because `browse-url-mail` will write to the buffer before it is ready.)

djcb:
Oh, my comment wasn't added... 

But indeed, `compose-mail` (which what we end up with) has some implicit expectations which are not met in the mu4e case, and one is that it's immediately dropped in the mail buffer (another one is that the message uses 'mail-header-separator, but at least mu4e is using that).

Chris00:
Maybe we should replace (through _advising_) `browse-url-mail` (the one that has expectations about `compose-mail`), triggered when `(eq mail-user-agent 'mu4e-user-agent)`?

averter:
With commit 286225b the email is almost perfect. The message body is the only thing missing. For some reason it gets written at the end of the org file which contains the `mailto:` link

Chris00:
May you explain what has https://github.com/djcb/mu/commit/286225be8f712186644bd4eb20d8573054f81499 do with it?

averter:
Sorry, I misread the issue. I was at 1.4.5 and in there not only the MAIN TEXT but also the SUBJECT and TO fields were empty. The update to the latest commit at least fixed those.

PR #1703, opened 2020-05-19
Advise browse-url-mail to insert the email body

Chris00:
Without this, calling
`(browse-url-mail "mailto:test@mu4e.org?subject=SUBJECT&body=MAIN TEXT")`
would not insert “MAIN TEXT” in the email body.

Fixes https://github.com/djcb/mu/issues/1698

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -71,6 +71,7 @@
 (require 'mail-parse)
 (require 'smtpmail)
 (require 'rfc2368)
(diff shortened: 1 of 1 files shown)


djcb:
Ah, thanks! Fixed this in a slightly different way.

alexmurray:
Can this also be merged on the release branch and a new stable release? Otherwise mu4e users can't easily report bugs in emacs itself.

djcb:
Yeah, that's the plan.

PR #1705, opened 2020-05-23
Allow ';' as a Maildir flags seperator

nickg:
Isync uses this by default on Windows where ':' is an invalid character
in file names. Also try to preserve the existing separator character
when generating a new file name.


djcb:
Looks good! Merged --thanks.

PR #1715, opened 2020-05-31
Fix #1714 void symbols link and query when org-store-link

### LINKED ISSUE #1714: The recently commits on master breaks org-store-link in mu4e
This commit:
```
7ed8d509 * mu4e-org: support more attributes for capturing
```

It removed let-binding caused void variables.

stardiviner:
----

#


djcb:
Thanks -- fixed.

PR #1731, opened 2020-06-19
mu4e: Delay the execution of workaround for desktop.el

cireu:
Some Emacs user may never use desktop.el, We don't need to require desktop.el
for them.


djcb:
Looks useful, but would like to avoid that `defvar`... can you use `(eval ...` in the hook instead? Thanks.

cireu:
Done, please check

djcb:
Merged -- thanks!

Issue #1729, opened 2020-06-13
Accepting an Outlook invite doesn't mark as accepted


behaghel:
## Expected or desired behavior

When I click the [ Accept ] button and send the response straightaway, on reception, I expect the organiser's calendar to update my status as invitee and mark me as accepted.

## Actual behavior

From a google calendar invite, this works. From an outlook (office 365) invite, this doesn't. The response is received with the attached ics marking attendance as accepted and yet outlook still keep me as "didn't respond".

## Steps to reproduce

open email, go on button [ Accept ], click enter, then send with C-c C-c, then check on outlook calendar.

## Versions of mu, mu4e/emacs, operating system etc.

- mu4e 1.4.1
- emacs 28.0.50

## Any other detail

I have tried to figure out if something was wrong with the text/calendar part by tweaking the `*CAL*` buffer but without being an expert, I don't think so. Tried also to figure how does outlook deal with it itself but it seems to deal with this through API and not attachment. I am more than happy to help in any way I can.

## Something I Discovered

Unrelated and may or may not be a bug. If both emails for organiser and attendee are listed as personal addresses in your mu, it then creates two ATTENDEE;PARTSTAT: lines one for each which then break the updating of the organiser calendar. In fact I have even managed to get 3+ of these lines by inviting more emails that I own.

Chris00:
@djcb May you assign me to this issue?
@behaghel I you have more time to investigate what is wrong with outlook (maybe try to remove the multipart??), that will be helpful.  I only have easy access to OWA (Outlook Web Application); if you have too may you check whether the behavior differs?

behaghel:
@Chris00 I am really keen to help. I suggest I start by inviting you on an email address of your choice. That would discard an issue with my setup. What do you think?

Chris00:
Use christophe+troestler → umons+ac+be (replace `+` by `.` and ` → ` by `@`).

Chris00:
I have tested with a colleague who uses Outlook and she says it works.

behaghel:
<img width="1028" alt="Screenshot 2020-06-16 at 21 55 49" src="https://user-images.githubusercontent.com/251832/84832805-d611f100-b025-11ea-9cc0-f38248a0c61c.png">

Thanks for allowing this test. I have the same behaviour with your config as with mine: it doesn't work. Not sure where we go from there...

Chris00:
Second reply sent.

behaghel:
\o/ Victory! \o/

[...]

PR #1732, opened 2020-06-19
Fix accepting ical RSVP with Outlook

Chris00:
- The multipart/alternative was problematic with Outlook (see https://github.com/djcb/mu/issues/1729).  
- Since a text reply is possible with the ical attachment, the original email quotation is saved to the kill-ring.


djcb:
Merged, thanks!

PR #1740, opened 2020-06-29
mu4e/mu4e-utils: only set so-long-mode once

stsquad:
Ideally we should separate the log buffer creation code so this van be
done a bit more cleanly. For now however only set so-long-mode once
otherwise you end up spamming the messages with constant:

  Changed to so-long-mode (from fundamental-mode) on account of line length.  C-c C-c to revert. [36 times]

As the messages keep rolling in.


djcb:
Ah yes.. thanks!

PR #1742, opened 2020-06-30
Don't declare mu_str_size_s() with the const attribute.

daybeforenet:
When this function is declared const or pure, clang at -O1 or higher optimizes
away the call to mu_str_size_s() inside mu_str_size(), so that it ignores its
argument and returns whatever is in mu_str_size_s()'s static buffer.

Found when test-mu-str failed while testing an update of mu in OpenBSD's ports tree.


djcb:
Ah, good catch. Merged, thanks.

Issue #1745, opened 2020-07-03
mu4e-view-image-max-height specified in pixels, but handled in 'lines'


mrvdb:
The default for `mu4e-view-image-max-height` is `600` which suggests that it is a "number of pixels"  as is the case with `mu4e-view-image-max-width` 

However, in `mu4e-display-image` the height is treated as a "number of lines" value.

We should either clarify this in the documentation strings and adjust to a more reasonable default or change the implementation.

mrvdb:
nevermind, read the code wrong.

There are issues with the height though, but this issue is not it ;-)

PR #1746, opened 2020-07-03
Satisfy both max height and width for images in view

mrvdb:
- introduce private function mu4e~image-width-scale:
    determine the width to use for proportional scaling based on the image width, height and the max
    restrictions.
- use it in mu4e-display-image


mrvdb:
This fixes what I intented to report in #1745 but failed to word properly there.

djcb:
Looks good - merged. Thanks!

PR #1747, opened 2020-07-03
NEWS.org, mu4e-vars.e.: fix spelling mu4e-contact-rewrite-function

dato:
mu4e-contact-rewrite-function was obsoleted in 1.4, but the entry in
NEWS.org entry, and the make-obsolete-variable call, referred to it as
mu4e-contacts-rewrite-function. (Should be "contact", not "contacts".)

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -100,7 +100,7 @@
       get the contacts that have changed since the last round.
 
       We also moved sorting the contacts to the mu-side, which speeds things up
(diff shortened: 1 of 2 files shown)


dato:
This seems trivial, but I literally almost missed fixing my setup because I blindly grep'd my .emacs with the mentioned variables in the NEWS file, as given.

djcb:
Merged -- thanks!

PR #1750, opened 2020-07-06
mu4e/mu4e-utils: save our position before deleting old data

stsquad:
Once your log buffer has filled up every new entry causes you to jump
back to the top of the log. Stop this happening by wrapping the
clean-up in save-excursion.


stsquad:
Hmm thinking about it I think the whole bit of character movement needs to be saved.

stsquad:
OK did the mini re-factor and wrapped the save excursion around all point stuff.

PR #1756, opened 2020-07-12
mu4e-main: fix hide-fully-read semantics; make defcustom

progfolio:
Reverse the semantics of `mu4e-main-hide-fully-read' so that t hides and nil shows.
The behavior now matches the variable name.
Additionally define the variable with defcustom so users may customize
it through the customization system.

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -39,9 +39,10 @@
   This also hides the warning if your `user-mail-address' is not
 part of the personal addresses.")
 
-(defvar mu4e-main-hide-fully-read nil
-  "When set to t, do not hide bookmarks or maildirs that have
-no unread messages.")
+(defcustom mu4e-main-hide-fully-read nil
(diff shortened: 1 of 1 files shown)


progfolio:
Alternatively, we could move `mu43-main-hide-fully-read` to `mu4e-vars.el` and rename it to `mu4e-hide-fully-read` or something similar. Happy to make that change if it's preferable to the defcustom in `mu4e-main.el`

djcb:
The `defcustom` change is fine, but the other I don't think is an improvement... so if you can just do the `defcustom`, I can merge. Thanks!

progfolio:
>  but the other I don't think is an improvement...

I'm not sure which change you would like reverted. Are you referring to the reversal of the behavior?:

```emacs-lisp
unless (or (plist-get bm :hide)
                      (and mu4e-main-hide-fully-read (eq unread 0)))
```

progfolio:
Closing this as I'm not sure what you wanted changed.

PR #1757, opened 2020-07-12
refactor: access server properties through generic function

progfolio:
New function, `mu4e~server-prop', allows accessing server properties
generically, rather than defining a function per property.

This should scale better as it doesn't repeat any logic per property.
If necessary, a property-specific function could easily be defined via `apply-partially'.


djcb:
Appreciate the effort, but don't think this is really an improvement. So, respectfully going to close this.

PR #1776, opened 2020-07-29
mu4e-headers: allow pipe through shell from headers view

### LINKED ISSUE #1752: Pipe through shell in headers view?

Looking at the keybindings help it seems to say that you should be able to pipe a message through a shell command, but trying to do so results in "| is undefined". It does work in message view, just not headers view. I don't know if it is just a documentation error, or if it meant to be able to pipe from the headers view.

## Actual behavior

emacs reports "| is undefined"

## Steps to reproduce

Open mu4e, jump to a folder, move to desired message, press '|'

Fedora 32, emacs 26.3-2, maildir-utils 1.4.10-1

## Any other detail

If it is supposed to function in headers view, I don't know if it is as simple as the following (which seems to work):

```
diff --git a/mu4e/mu4e-headers.el b/mu4e/mu4e-headers.el
index bf7430d5d3c2..4b0611fd298a 100644
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -919,6 +919,8 @@ after the end of the search results."
           (define-key map "$" 'mu4e-show-log)
           (define-key map "H" 'mu4e-display-manual)

+          (define-key map "|" 'mu4e-view-pipe)
+
           ;; menu
           ;;(define-key map [menu-bar] (make-sparse-keymap))
           (let ((menumap (make-sparse-keymap)))
@@ -942,6 +944,7 @@ after the end of the search results."
                                            (and (boundp 'mu4e-headers-show-threads)
                                                 mu4e-headers-show-threads))))

+            (define-key menumap "|" '("Pipe through shell" . mu4e-view-pipe))
             (define-key menumap [sepa1] '("--"))

             (define-key menumap [execute-marks]  '("Execute marks"

```
For my use case I ended up creating a custom header action so I could mark multiple messages and then do a single action command, but thought I would report this in case we are supposed to be able to pipe from the headers view.

Regards,
Jerry

snits:
Being able to pipe through shell from the headers view is
convenient for some use cases, so wire it up to work.

Resolves #1752

Signed-off-by: Jerry Snitselaar <jsnitsel@redhat.com>

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -919,6 +919,8 @@ after the end of the search results."
           (define-key map "$" 'mu4e-show-log)
           (define-key map "H" 'mu4e-display-manual)
 
+          (define-key map "|" 'mu4e-view-pipe)
+
           ;; menu
           ;;(define-key map [menu-bar] (make-sparse-keymap))
           (let ((menumap (make-sparse-keymap)))
@@ -942,6 +944,7 @@ after the end of the search results."
                                            (and (boundp 'mu4e-headers-show-threads)
                                                 mu4e-headers-show-threads))))
 
+            (define-key menumap "|" '("Pipe through shell" . mu4e-view-pipe))
             (define-key menumap [sepa1] '("--"))
 
             (define-key menumap [execute-marks]  '("Execute marks"


djcb:
Merged -- thanks!

PR #1777, opened 2020-08-03
Add Org entry when accepting event

jesusbv:
When clicking "Accept" event, if the entry is new
it is not added to the Org file, besides having
gnus-icalendar-org-enabled-p set to true.

This modification changes that, so that a new entry
is added or an existing one is modified.


djcb:
Any comments, @Chris00?

jesusbv:
Hi, any feedback on this ?

djcb:
Merged -- thanks!

PR #1796, opened 2020-09-08
Guile 3.0 support

dannyob:
Hullo!

GNU Guile released [version 3.0 in January](https://www.gnu.org/software/guile/news/gnu-guile-300-released.html); this patch should allow mu to build with either v2.2 or v3.0 of that language. (The bindings don't seem to have needed any changes, but I hacked the build to be more flexible.)

I'm not the greatest coder, but this builds and passes all the tests on my machine. I use Guix, which uses Guile a lot, and let me switch between building against both versions.

There were a couple of comments in the original code talking about how hacky the solution was for 2.2, and how its path needed to be hard-coded  to avoid a `make distcheck` error.  I didn't see any problems and this new code seems to build okay, so I took out those comments too. 

This patch is for the v1.14.x branch.

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -108,6 +108,10 @@
     - The mu4e <-> mu interaction has been rewritten to communicate using
       s-expressions, with a repl for testing.
 
+*** guile
+
+    - guile 3.0 is now supported; guile 2.2 still works.
+
 *** toys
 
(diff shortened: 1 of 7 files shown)


djcb:
Thanks, I'm interested in this, but not for 1.4.x, so I'll close this one (and we continue with #1797.

PR #1797, opened 2020-09-08
Guile 3.0 update, based on master.

dannyob:
This is the same set of patches as pull #1796 , except using master as the base. I just cherry-picked them, so you can accept either and then apply them to the other branch. Let me know if there was a smarter way to do this -- I'm eager to learn!

### DIFF
--- a/NEWS.org
+++ b/NEWS.org
@@ -149,6 +149,10 @@
     - The mu4e <-> mu interaction has been rewritten to communicate using
       s-expressions, with a repl for testing.
 
+*** guile
+
+    - guile 3.0 is now supported; guile 2.2 still works.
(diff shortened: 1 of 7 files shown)


djcb:
This looks useful, thanks! 

However guile is only as soft requirement, and this seems to make it a hard requirement.. i.e. from the CI:  
   https://github.com/djcb/mu/pull/1797/checks?check_run_id=1084030032

dannyob:
Huh! I'll take a look at that!

djcb:
Merged this locally, with a fix configure.ac. Thanks!

dannyob:
Thank you!

PR #1803, opened 2020-09-22
Properly align variable-pitch faces in header view

legalnonsense:
This change allows the use of variable-pitch fonts in the header view, because it aligns columns using the :align-to value of the display text property rather than relying on spaces. Since the new function needs to know the field, I had to change the call to `mu4e~headers-truncate-field` (it looks like this changed recently since my version already had FIELD as a parameter).


djcb:
Nice work! I merged it locally, with some minor changes (field-id vs field).

PR #1807, opened 2020-10-02
Remove configuration check for timegm()

juanjosegarciaripoll:
timegm() is no longer used in the mu source base. I am trying to find what are the minimal changes to allow configuring this code in mingw/msys (which is my preferred basis for compiling Emacs itself).


djcb:
Merged -- thanks!

PR #1808, opened 2020-10-04
Allow use of Maildir FAT extension changing the Maildir info separator

juanjosegarciaripoll:
When working on Windows systems, mbsync and other software uses ';' as separator for Maildir flags, because ':' is not allowed in file names. Change the function that creates names for draft and sent emails, so that it depends on a global customizable variable that determines this separator.


djcb:
That seems useful. Thanks!

PR #1812, opened 2020-10-10
Add reference about read-mail-command to mu4e.texi.

lockywolf:
As far as I understand, this is a legitimate usage of `mu4e`, isn't it?

### DIFF
--- a/mu4e/mu4e.texi
+++ b/mu4e/mu4e.texi
@@ -3225,6 +3225,15 @@ you can do so by adding the following to your configuration:
 (setq mail-user-agent 'mu4e-user-agent)
 @end lisp
 
+Similarly, to specify @t{mu4e} as your preferred method for reading mail,
+customize the variable @code{read-mail-command}.
+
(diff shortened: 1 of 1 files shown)


mu4e/mu4e.texi
```diff
@@ -3464,7 +3474,7 @@ or create a new one.
 @section Hydra
 
 People sometimes ask about having multi-character shortcuts for
-bookmarks; and easy way to achieve this, is by using anm emacs package
+bookmarks; and easy way to achieve this, is by using an emacs package
 called Hydra@footnote{@url{https://github.com/abo-abo/hydra}}.
```
djcb, line 3478: ah, while you are changing this, can you also change 'and easy way' into 'an easy way'? Thanks!

lockywolf:
I think that this pull request can be merged, if I find some other places the documentation can be improved, I'll open another one.

djcb:
Merged, thanks!

PR #1818, opened 2020-10-19
Check for nil before calling mu4e-user-mail-address-p

wavexx:
Got a few errors in `*mu4e-headers*` while skimming through a junk folder due to a nil :from address.
I'd rather have mu4e-user-mail-address-p not called in such cases, so we filter it earlier.

There's a similar second instance in mu4e-view.el:272 , when constructing "from-to", but I'm not sure if that's applicable? (I couldn't select a message that would trigger it).


djcb:
Merged, thanks. 

(I'll do some follow commits, since e.g. `mu4e-user-mail-address-p` is actually deprecated)

wavexx:
On Mon, Oct 19 2020, Dirk-Jan C. Binnema wrote:
> (I'll do some follow commits, since e.g. `mu4e-user-mail-address-p` is
> actually deprecated)

Fully aware, but was to lazy to update my bloated config so far ;)

PR #1835, opened 2020-11-13
Use functions instead of lambdas in add-hook calls

thierryvolpiatto:
This is cleaner and help when debugging to use functions in hooks instead of lambdas.


djcb:
Merged, thanks!

PR #1836, opened 2020-11-15
Disable `mm-verify-option` let-bounded in `mu4e~view-gnus`

thierryvolpiatto:
When this option is set to 'known and the needed key is not yet in the user's
keyring, mu4e may hangs forever unable to verify email so lets user set this
variable globally to 'never to prevent this issue.


djcb:
Merged, thanks!

PR #1837, opened 2020-11-15
Use readdir(3) instead of deprecated readdir_r(3) in mu-maildir.c

eklitzke:
I was originally looking at this code because clang was producing an alignment warning in `dirent_new`, but once I took a look I realized this code can be simplified a lot. Usage of `readdir_r(3)` is discouraged according to the [glibc docs](https://www.gnu.org/software/libc/manual/html_node/Reading_002fClosing-Directory.html), which explain the problems with `readdir_r(3)` better than I can here. There's an even more detailed explanation of the issue [in this blog post](https://elliotth.blogspot.com/2012/10/how-not-to-use-readdirr3.html). Basically all modern implementations of libc use a pointer to data within the `DIR*` handle for the return value of `readdir(3)` (including on bsd/macos), so as long as that `DIR*` handle isn't used my multiple threads concurrently there aren't thread safety issues.

I think there's also a bug in the original implementation because it allocates memory using `g_new0` but frees it using `g_slice_free1`, but with the change I've made here that code gets deleted anyway.

### DIFF
--- a/lib/mu-maildir.c
+++ b/lib/mu-maildir.c
@@ -438,22 +438,6 @@ process_dir_entry (const char* path, const char* mdir, struct dirent *entry,
 }
 
 
-static const size_t DIRENT_ALLOC_SIZE =
-	offsetof (struct dirent, d_name) + PATH_MAX;
-
-static struct dirent*
-dirent_new (void)
-{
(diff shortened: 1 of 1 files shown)


djcb:
Merged, thanks! 

Note that the code here is mostly replaced by mu-scanner, but always good to fix a warning.

PR #1838, opened 2020-11-16
No need to call `read-only-mode` when buffer is already read-only

thierryvolpiatto:
In mu4e~view-gnus, mu4e-view-mode derive from gnus-article-mode which derive
from gnus-mode which derive from special-mode:
(get 'mu4e-view-mode 'mode-class) => special.


djcb:
Merged, thanks.

PR #1839, opened 2020-11-16
Avoid looping through buffer-list when not needed

thierryvolpiatto:
Actual code loop through buffer-list even when `message-kill-buffer-on-exit` is nil.


djcb:
Merged, thanks!

PR #1850, opened 2020-12-04
Do not use shell commands to extract version number

juanjosegarciaripoll:
Avoid using 'head', 'sed' or other POSIX commands to extract the version number when Emacs can do it. This allows installing just 'mu', without the extra unix tools in a Windows environment, or run 'mu' from MSYS but do not expose the rest of the MSYS environment (needed if one wants to rely on Windows cmd.exe shell for other tasks, for instance).


djcb:
Merged, thanks!

PR #1868, opened 2020-12-20
Completion for search query keywords and fixed or determinable values

tsdh:
This pull requests adds TAB-completion for search query keywords (like `from:`, `subject:`, etc.), and also for their values in case they are fixed (like with `flag:`) or can be determined at runtime (like for `maildir:`, and `mime:`).

I didn't know which Emacs version mu4e requires so I've `fboundp`-checked the needed `minibuffer-with-setup-hook` macro (available since Emacs 25.1) and fall back to "no completion" if that's undefined.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1551,6 +1551,43 @@ or `past'."
 (defvar mu4e~headers-search-hist nil
   "History list of searches.")
 
+(defconst mu4e--search-query-keywords
+  '("and" "or" "not"
+    "from:" "to:" "cc:" "bcc:" "contact:" "date:" "subject:" "body:"
+    "list:" "maildir:" "flag:" "mime:" "file:" "prio:" "tag:" "msgid:"
+    "size:" "embed:"))
+
+(defun mu4e--search-query-competion-at-point ()
+  (cond
+   ((not (looking-back "[:\"][^ \t]*"))
+    (let ((bounds (bounds-of-thing-at-point 'word)))
+      (list (or (car bounds) (point))
+            (or (cdr bounds) (point))
+            mu4e--search-query-keywords)))
+   ((looking-back "flag:\\(\\w*\\)")
+    (list (match-beginning 1)
+          (match-end 1)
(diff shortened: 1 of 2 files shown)


mu4e/mu4e-headers.el
```diff
@@ -1570,7 +1607,16 @@ searching. If SHOW is non-nil, show the message with MSGID."
           (if edit
               (read-string prompt expr)
             (or expr
-                (read-string prompt nil 'mu4e~headers-search-hist)))))
+                (if (fboundp 'minibuffer-with-setup-hook)
+                    ;; Available since Emacs 25.1.
+                    (minibuffer-with-setup-hook
```
djcb, line 1612: We require emacs 25.3, so no need for the check.

mu4e/mu4e.texi
```diff
@@ -1910,9 +1910,16 @@ see @ref{Sorting and threading}.
...
-look at some examples here; you can consult the @code{mu-query} man page
-for the details.
+subject to the shell's interpretation before @t{mu} sees them}. You can
+consult the @code{mu-query} man page for the details.
+
+Additionally, @t{mu4e} supports @kbd{TAB}-completion for queries. There
+there is completion for all search keywords such as @code{and},
+@code{from:}, or @code{date:} and also for certain values, i.e., the
+possible values for @code{flag:}, @code{prio}, @code{mime:}, and
+@code{maildir:}.
```
djcb, line 1920: `prio:` I guess?

djcb (commented):
Nice work! A few comments, if you can update for those, I'll merge it. Thanks!

mu4e/mu4e-headers.el
tsdh: Done!

mu4e/mu4e.texi
tsdh: Yes, right.

tsdh:
@djcb I've fixed your two points, rebased again onto master, and pushed.

djcb:
Thanks! Merged.

PR #1871, opened 2020-12-24
Make libreadline really optional.

juanjosegarciaripoll:
Add a --disable-readline argument to the 'configure' script, removing the need of libreadline which pulls ncurses and a lot of other files, not needed when mu is only used from inside emacs.


djcb:
Merged, thanks!

PR #1884, opened 2021-01-01
Add message counts to `mu4e-ask-bookmark`

ebellani:
This is specially useful for using bookmarks if one does not frequent the main
view window a lot. (For instance mu4e-split-view is set to 'single-window mode)


djcb:
Sorry for not responding earlier;  

To get up-to-date message counts without slowing things down would be quite hard (i.e., `j` would imply a bunch of queries. That could be cached, but that would add a bit too much complexity imho.

So thank you but have to decline this.

PR #1887, opened 2021-01-04
Fix 'mu4e-interrupt-update-mail alias definition

wavexx:
Fixes build during byte-compilation.


tsdh:
For the context, the WHEN argument in all obsoletion macros is now required since a recent change to emacs master (aka 28).  Its omission has been declared problematic and generated a warning since emacs 23.1.

epaulson10:
I opened #1895 to fix this before I saw this PR. I’ll close mine to let you have the glory, though I suggest changing the when value to “1.0-alpha0” which is the first tag that contains the commit that added the alias.

djcb:
Thanks, merged this locallly (with a small change for the version, as epaulson10 suggest).

PR #1894, opened 2021-01-10
Add a hook on switching from mu4e-compose-mode to org-mode

stardiviner:
I want to enable some settings locally only when `mu4e-compose-mode` switching to `org-mode`.
Here is an example:
```elisp
(when (boundp 'org-mu4e-compose-switch-org-mode-hook)
    (add-hook 'org-mu4e-compose-switch-org-mode-hook
              (lambda ()
                (with-eval-after-load 'company
                  (add-to-list 'company-backends 'company-ispell 'append)))))
```


djcb:
Thanks for the effort, but org-mu4e-compose has been obsolete for many years, so I'm not going to add any changes for that, sorry...

PR #1895, opened 2021-01-10
Add missing when parameter to define-obsolete-function-alias

epaulson10:
Emacs 28 is making this 'when' argument mandatory, as the
'make-obsolete' and related functions without a 'when' were marked as
obsolete in Emacs 23.1.

I used 'git tag --contains b6985e13c' to determine the value of
'when'.


epaulson10:
Closing because #1887 beat me to fixing this bug.

PR #1899, opened 2021-01-13
Do not show trailing whitespace when viewing messages

Chris00:


djcb:
Hmm, isn't `show-trailing-whitespace` is `nil` by default?

Chris00:
It is.  I saw that being done in Gnus — maybe an artifact of the past before the variable was buffer local?

Chris00:
Specifically, you can [find it here](https://github.com/emacs-mirror/emacs/blob/master/lisp/gnus/gnus-art.el#L4521).  Maybe you would prefer it to be applied only when `mu4e-view-mode` derives from `special-mode` since it will be applied anyway in the other case?

Chris00:
Now that `(setq show-trailing-whitespace nil)` is part of [`gnus-article-mode`](https://github.com/emacs-mirror/emacs/blob/master/lisp/gnus/gnus-art.el#L4451), it will eventually apply to mu4e.  Closing.

Issue #1886, opened 2021-01-04
Accents in view mode


myravian:
Hi there,

## Expected or desired behavior

In the view-mode, the accents should be properly displayed. 

## Actual behavior

The accents are not shown properly in text view-mode (non html). All the possible UTF-8 settings in .emacs have been enabled, and accents show up when writing emails or using html-mode view. The problem is just with text view-mode. 

Is this solved by the fix to https://github.com/djcb/mu/issues/1823 maybe?

## Steps to reproduce

For bug-reports, please describe in as much details as possible how
one can reproduce the problem*.

## Versions of mu, mu4e/emacs, operating system etc.

Ubuntu 19.04, mu/mu4e 1.10

Chris00:
@myravian Solutions [are being discussed](https://github.com/djcb/mu/issues/1823).  May you please chime in and let us know whether they fix your examples too?  Thanks.

Issue #1823, opened 2020-10-28
Bad encoding in mu4e-view with `mu4e~view-gnus`


thierryvolpiatto:
## Expected or desired behavior

Have accentuated characters correctly encoded.

## Actual behavior

Accentuated characters are not encoded. 

e.g. `pièce => pi\303\250ce`

This happen only in the gnus view, mu4e~view-internal is encoding properly.

Possibles solutions:

1) Force reencoding instead of limiting to message with a content-type header:

```diff
diff --git a/mu4e/mu4e-view.el b/mu4e/mu4e-view.el
index d415d254..16ac7b66 100644
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -390,11 +390,9 @@ article-mode."
     (switch-to-buffer (get-buffer-create mu4e~view-buffer-name))
     (erase-buffer)
     (insert-file-contents-literally path)
-    (unless (message-fetch-field "Content-Type" t)
-      ;; For example, for messages in `mu4e-drafts-folder'
-      (let ((coding (or (default-value 'buffer-file-coding-system)
-                        'prefer-utf-8)))
-        (recode-region (point-min) (point-max) coding 'no-conversion)))
+    (let ((coding (or (default-value 'buffer-file-coding-system)
+                      'prefer-utf-8)))
+      (recode-region (point-min) (point-max) coding 'no-conversion))
     (setq
      gnus-summary-buffer (get-buffer-create " *appease-gnus*")
      gnus-original-article-buffer (current-buffer))

```
2) Use insert-file-contents (not litteraly)

```diff
diff --git a/mu4e/mu4e-view.el b/mu4e/mu4e-view.el
index d415d254..83b75113 100644
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -389,7 +389,7 @@ article-mode."
                                             gnus-buttonized-mime-types)))
     (switch-to-buffer (get-buffer-create mu4e~view-buffer-name))
     (erase-buffer)
-    (insert-file-contents-literally path)
+    (insert-file-contents path)
     (unless (message-fetch-field "Content-Type" t)
       ;; For example, for messages in `mu4e-drafts-folder'
       (let ((coding (or (default-value 'buffer-file-coding-system)

```
In this case it is probably not needed to recode region as insert-file-contents is already doing it.

## Versions of mu, mu4e/emacs, operating system etc.

Last Mu from Git, Emacs-27, Linuxmint.

[...]

PR #1900, opened 2021-01-14
Fix the decoding of accents in Gnus view

Chris00:
Followup on https://github.com/djcb/mu/issues/1823
Fixes https://github.com/djcb/mu/issues/1886


djcb:
Merged -- thanks!

Issue #1857, opened 2020-12-13
[Feature request] Contact list can be sorted by frequency


rhgg2:
## Expected or desired behavior

Mu4e contact list can be sorted in a manner which favours frequency of contact

## Actual behavior

Mu4e contact list is sorted in a manner which favours most-recent contact.

## Versions of mu, mu4e/emacs, operating system etc.

mu/mu4e 1.4.13, Emacs 26.2

## Any other detail

I just upgraded from mu 1.0.x. The contact handling has changed substantially, and in particular, it seems that mu now only passes to mu4e a list of pre-sorted contacts, rather than passing data allowing mu4e to do the sorting itself.

However, the sorting algorithm which mu uses is less useful than the one which mu4e used previously. The current sorting algorithm is, I think, in ContactInfoLessThan in mu-contacts.cc and is that:

(1) personal comes first  
(2) then more recent comes first  
(3) then more frequent comes first  
(4) then alphabetical

On the other hand, mu4e (in the removed function mu4e~sort-contacts) used to sort as follows:

(1) personal comes first  
(2') then, if one contact was last mentioned in the last 15 days, and the other was not, that one comes first  
(3) then more frequent comes first

Now (2') is much superior to (2), because (2) makes it almost impossible for (3) to be triggered, unless you happened to receive an email from two different contacts at exactly the same time. 

Would it be possible to implement (2') in mu-contacts.cc as an alternative sorting algorithm for contacts in mu-server and mu-clist (maybe with the hard coded 15 days made into an option?)

djcb:
I think it's bit arbitrary whether "more recent" or "more frequent" is more important; and the way completion in emacs works is that many of the completion methods completely ignore the order provided by mu4e, so many users don't even notice.

Anyway, i'm not against 2', but it would become the one and only sorting method, we have too many options already.

rhgg2:
Thanks for the reply. I use helm, which quite definitely takes the list in the order it is provided and matches the first item which fits in that order. The upshot is that my email completion has become rather less useful with the new version (though lots of other things have become much better.) With the old mu4e sort algorithm I didn't even have to check whether the autocomplete was correct, as it would always get the intended contact first time. Now it only gets it right about 50% of the time.

PR #1910, opened 2021-01-22
Make order relation on contacts antisymmetric

rhgg2:
The behaviour suggested in #1857 (and the old behaviour of mu4e when that did the sorting) is that if exactly one of ci1 or ci2 was seen in the last 15 days, then that should be selected as first in the order by ContactInfoLessThan. As it stands, we return the most recent of ci1 and ci2 as least in the order only when ci1 was seen in the last 15 days. So if ci2 was seen in the last 15 days, but ci1 was not, it will fall through to a comparison on frequency; while if ci1 was seen in the last 15 days, but ci2 was not, it will always return ci1 as least. This is not antisymmetric in ci1 and ci2.

### DIFF
--- a/lib/mu-contacts.cc
+++ b/lib/mu-contacts.cc
@@ -87,7 +87,7 @@ struct ContactInfoLessThan {
                 if (ci1.personal != ci2.personal)
                         return ci1.personal; // personal comes first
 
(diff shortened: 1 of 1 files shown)


djcb:
Yeah, not quite sure what I was thinking... thanks for the fix.

Issue #1912, opened 2021-01-25
Feature request: save ical events to date-tree


plantarum:
I am using `mu4e-icalendar` to capture calendar invites to my org files, with the following config:

```
;; Use gnus article view
(setq mu4e-view-use-gnus t)

(require 'mu4e-icalendar)
(mu4e-icalendar-setup)

(setq mu4e-icalendar-diary-file "~/org-roam/diary")
(require 'org-agenda)
(setq gnus-icalendar-org-capture-file "~/org-roam/meetings.org")
(setq gnus-icalendar-org-capture-headline '("vcal"))
(gnus-icalendar-org-setup)

```

This works well and does exactly what I expect. However, events are added in the order I process them, into a single tree, `vcal`. Is it possible to store them instead in a date-tree structure? i.e., instead of:

```
* vcal
** event one
** event three
** event two 
```

I'd like this:

```
* 2021

** 2021-W04

*** 2021-01-26 Tuesday

**** Event three

*** 2021-01-27 Wednesday

**** Event one

*** 2021-01-29 Friday

**** Event two

```

My objective is to maintain a single, chronological listing of all my events, including things captured from vcal invites, and also regular org captures.

Is there any way to do this with through configuring mu4e? If not, do you have any suggestions as to how to attack the problem of integrating ical capture with org's date-tree storage system?

plantarum:
I figured out how to change the capture template that is used, see #1913. Unfortunately, when using a datetree target, the date is set as the date of capture (i.e., today), not the date of the event. I'll keep looking into that.

plantarum:
I figured out how to do this by advising the gnus-icalendar function. I've added an explanation to the new documentation, if you think it will be generally useful.

PR #1913, opened 2021-01-26
Documented icalendar capture template

plantarum:
It's not documented in gnus-icalendar, so the explanation comes straight from my reading of the code.

Being able to customize the capture template partially addresses the issue I raised in #1912. gnus-icalendar supports this, in an undocumented way, and the code comments suggest the details may change eventually.


plantarum:
This now completely addresses #1912.

plantarum:
Sorry, I don't understand why the checks have failed. I only changed one file, the info source, and the error appears to reference something else.

djcb:
Merged -- thanks!

PR #1921, opened 2021-02-03
mu4e-jump-to-maildir: offer editing the query when a prefix arg is given

tsdh:
I have many maildirs (because I also use other clients, like a mobile one, which haven't the searching super-powers of mu) and configured shortcuts for them.  Frequently, I want to read such a maildir but do not want to see all messages but just unread messages or messages with attachments.  For such purposes, I've added an optional prefix arg to `mu4e-jump-to-maildir` which allows the `maildir:"/foo/bar"` query to be refined/extended before execution.  So it's basically the same as for bookmarks, i.e., `B` instead of `b`.

If you want, I can also add a new command bound to `J` which calls `mu4e-jump-to-maildir` with non-nil prefix arg so that `j` and `J` are really analogous to `b` and `B`.

I've adapted docstring and texi docs accordingly.

I also refactored the now two usages of the "read a query with completion" feature into a new separate function `mu4e-read-query`.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1542,7 +1542,16 @@ or `past'."
      (pop mu4e~headers-query-future))))
 
 
-;;; Interactive functions
+;;; Reading queries with completion
+
+(defun mu4e-read-query (prompt &optional initial-input)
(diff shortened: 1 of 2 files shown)


djcb:
Looks useful, thanks!

PR #1931, opened 2021-02-17
Read query with completion also when editing a bookmark

tsdh:
This wraps up some change that somehow didn't get applied when merging PR #1921
and also offers completion when editing a bookmark (which I've forgotten in
that PR).


djcb:
Merged, thanks!

PR #1934, opened 2021-02-20
mu4e-compose-context-switch for context switching in mu4e:compose buffers

### LINKED ISSUE #1772: Switching context on a message draft

When I am editing a message, if I switch context between accounts I expect the `From:` field to be updated accordingly and the message draft to be moved to the correct folder associated with the context.

## Actual behavior

The `From:` field is not updated and the message stays in the draft folder of the original context.

## Steps to reproduce

- Have multiple contexts in `mu4e-contexts`
- Create a new draft
- Call `mu4e-switch-context` and switch to a different context
- The context is switched (as evidenced by the modeline)
- The message stays in the original context draft folder and the `From` field is not updated.

Emacs: 26.3
mu: 1.5.5

haji-ali:
Fixes #1772 by implementing context switching in `mu4e-compose` which also changes the `From` and moves the draft to the correct folder.

This mostly works. However, after switching context `mu4e~proc-move` opens a message view even though I pass `t` to `no-view` (I would rather not open the view).  

One also has to press `E` to edit the viewed message. I want to call `mu4e~proc-compose` at the end to open the newly moved message automatically, but I can't find a way to get the doc-id from the message-id.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -446,6 +446,7 @@ removing the In-Reply-To header."
   (setq mu4e-compose-mode-map
         (let ((map (make-sparse-keymap)))
           (define-key map (kbd "C-S-u")   'mu4e-update-mail-and-index)
+          (define-key map (kbd "C-c C-;")   'mu4e-compose-context-switch)
           (define-key map (kbd "C-c C-u") 'mu4e-update-mail-and-index)
           (define-key map (kbd "C-c C-k") 'mu4e-message-kill-buffer)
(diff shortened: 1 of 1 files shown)


djcb:
Looks good! Merged this locally with some minor changes.

haji-ali:
That's great @djcb.

Just one question. I want to add a call to `mu4e~proc-compose` on the newly moved message at the end of the context switching so that one can continue with editing the message. However, I can't seem to find a way to get a doc-id (which `mu4e~proc-compose` accepts) out of a message-id (which I have). Is there a way to do this?

PR #1937, opened 2021-02-23
Fix error in mu4e-view-gnus-save-mime-parts

thierryvolpiatto:
where `handle' may not be a list (here a marker when I had error).

So ensure it is a list.


djcb:
Merged -- thanks!

PR #1950, opened 2021-03-16
Fix warnings in mu4e-icalendar

Chris00:


mu4e/mu4e-icalendar.el
```diff
@@ -172,13 +171,10 @@ response in icalendar format."
       ;; also trash the message (thus must be appended to hooks).
       (add-hook
        'message-sent-hook
-       #'mu4e~icalendar-setup-sent-hook-fn
+       (defun mu4e~icalendar-trash-hook()
+         (setq mu4e-sent-func (mu4e~icalendar-trash-message original-msg)))
        t t))))
```
djcb, line 176: Hmmm, the manual describes the return value of `defun` as 'undefined', so let's not depend on that. 

Also, I don't quite understand this, `mu4e-sent-func` should be a function, but here it's set to the result of `mu4e~icalendar-trash-message`. And where do we get the `original-msg` from?

Chris00: I can use a lambda instead — but [this was changed not to](https://github.com/djcb/mu/commit/0109172ad4daddd0fd691828da9934be45888d17#diff-bea9db4624a70008cf0cdc8b59dcb83bfa6833541da1ec1b176c76918c9c74d9) and this is the problem with the variable `original-msg` being not defined.  Note that `mu4e~icalendar-trash-message` [returns a function](https://github.com/djcb/mu/blob/master/mu4e/mu4e-icalendar.el#L128).  Here we get `original-msg` from the lexical environment (because lexical binding is enabled).

Chris00: I pushed a new proposal.

djcb: Ah, I see... I missed the `lambda`-part in `mu4e~icalendar-trash-message`, and forgot about the lexical binding. Sorry for the changes behind your back.

Still if you could make it a lambda rather than a defun, then we're future-proof - thanks!

Chris00: I used an auxiliary function `mu4e~icalendar-trash-message-hook` — the closure captures less of the environment (just `original-msg` and nothing else).

djcb: Yeah looks good, thanks!

djcb:
Merged, thanks.

PR #1952, opened 2021-03-18
mu4e/view-gnus: Make Gnus email clickable links work

Chris00:
The `From` link triggers _reply_ and the email links launch _mu4e_ (**not** `message-mail`).


djcb:
Ah, that is useful. Merged, thanks.

PR #1953, opened 2021-03-23
Add ivy completion when saving attachments

ahttraga:
[ivy](https://github.com/abo-abo/swiper) is an alternative completion framework to *helm*. I added an extra conditional to use `completing-read` to save attachments (`mu4e-view-save-attachments` in `mu4e-view-gnus.el`) when *ivy* is in use.


djcb:
Looks useful -- merged, thanks!

djcb:
Hmm, ivy cannot handle multiple selections (like helm does), so we can't really use it here.

PR #1957, opened 2021-03-31
mu4e/mu4e-headers: remove excess message output

stsquad:
When marking threads as read things are slowed down by echoing the
thread path to the mini-buffer. I assume this is left over debug but
if needed for something else should probably be a log call.


djcb:
Oh yeah that was debugging out. Good catch!

Issue #1928, opened 2021-02-15
C-x 4 m doesn't open in new window using mu4e-user-agent


sje30:
## Expected or desired behavior

With 

    (setq mail-user-agent 'mu4e-user-agent)
    (compose-mail-other-window)

should open *draft* in new window.  (The desired behaviour is observed with
message-user-agent .)

## Actual behavior

*draft* appears in current window.

## Any other detail

1.4.13 with Emacs 27.1

The problem is that mu4e/mu4e-compose.el is defined as follows

    (defun mu4e~compose-mail (&optional to subject other-headers _continue
                                        _switch-function yank-action _send-actions _return-action)

where you can see _switch-function (and several other parameters) are
ignored.  Would it make sense to make mu4e~compose-mail more similar
to message-mail in this regard?

djcb:
Sure, that'd be useful.

PR #1960, opened 2021-04-03
Respect switch-function in compose-mail.

sje30:
Emacs has several standard keybindings

C-x m   compose-mail
C-x 4 m compose-mail-other-window
C-x 5 m compose-mail-other-frame

This patch fixes the creation of new mail buffers to respect the
latter two keybindings, C-x 4 m and C-x 5 m.

Note that there is already the variable mu4e-compose-in-new-frame
which if true opens in a new frame.  That will still work for C-x m
and C-x 5 m, but if the user runs C-x 4 m, it switches to other-window
as it assumes the keybinding takes precedence.  This behaviour can be
changed within mu4e~draft-open-file.

See also #1928


djcb:
Merged, thanks!

Issue #1956, opened 2021-03-30
Incompatibility with org-msg package


baracunatana:
## Expected or desired behavior
I can accept an ical invitation while having the org-msg package activated

## Actual behavior
org-msg takes over and the invitation reply message isn't formatted correctly

## Steps to reproduce
* Install mu 1.5.11
* Install latest org-msg
* Try to reply to a calendar invitation

## Versions of mu, mu4e/emacs, operating system etc.
Emacs 28.0.5 with mu 15.11

## Any other detail
From the gnus view, if I try to reply to a calendar invitation while having org-msg-mode active, the reply message isn't formatted correctly and the invitation isn't properly accepted/declined. Is you deactivate org-msg-mode, the reply goes as expected, so this is definitely a compatibility issue.

For now, I fixed it with the following lines in my config file:

    (defun j/deshabilitar-org-msg (orig-fun &rest args)
    (let ((activo org-msg-mode))
      (org-msg-mode -1)
      (apply orig-fun args)
      (if activo
        (org-msg-mode))))

    (advice-add 'gnus-article-press-button :around #'j/deshabilitar-org-msg)

Now, I'm not sure if I should report this here or in org-msg. Any advice on this regard (or in a more elegant way to solve this) is appreciated.

djcb:
@Chris00: any idea?

Chris00:
I'll install org-msg and see what I can do.

Chris00:
I submitted [some](https://github.com/jeremy-compostella/org-msg/pull/91) [PR](https://github.com/jeremy-compostella/org-msg/pull/93) and [issue](https://github.com/jeremy-compostella/org-msg/issues/92).

Chris00:
The project is interesting but it does not mix well with [MML](https://www.gnu.org/software/emacs/manual/html_node/emacs-mime/MML-Definition.html) directives.  I'll keep working on it to see the best way to solve the issue.

djcb:
Oh, great work, @Chris00!

Chris00:
This has been fixed on the OrgMsg side as well.  We can close this issue.

PR #1962, opened 2021-04-05
mu4e/compose: attach buffers & files setting any property

Chris00:
This fixes the fact that `:disposition` was advertised but not queried.  In addition, one has the possibility to add a `:description` and to attach buffers.  This may be used in the resolution of https://github.com/djcb/mu/issues/1956.


djcb:
Merged, thank you!

PR #1965, opened 2021-04-06
Mention texinfo build dependency

### LINKED ISSUE #1964: Please add texinfo as build dependency in the docs
I'm building mu from git on a current Ubuntu (on WSL2, so the distro might slightly differ from 'plain' Ubuntu). The `texinfo` package is not installed by default, but is a build dependency for `mu` (see also #1210). The docs have [a list of build dependencies on Ubuntu](https://www.djcbsoftware.nl/code/mu/mu4e/Installation.html#Installation), I think that list should include `texinfo`. 

Thanks!

tinloaf:
Fixes #1964

### DIFF
--- a/mu4e/mu4e.texi
+++ b/mu4e/mu4e.texi
@@ -273,7 +273,7 @@ apply to you, you can follow either @ref{Building from a release tarball} or
 @subsection Dependencies for Debian/Ubuntu
 
 @example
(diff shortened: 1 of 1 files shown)


djcb:
Thanks -- but this should not be necessary anymore.

PR #1970, opened 2021-04-09
skip confirmation when executing marks

Gallipo:
When running `u4e-mark-execute-all` with any prefix-arg you can skip the confirmation-prompt. The potential for accidentally hitting this is quite low.


djcb:
Merged, thanks.

PR #1976, opened 2021-04-12
mu4e-mark: Allow executing marks without passing a prefix argument

Chris00:


djcb:
Hmm, calling this without a prefix argument works just fine... did you you mean calling it non-interactively?

Chris00:
Sometimes the following is triggered (with debugger on);
```
Debugger entered--Lisp error: (wrong-number-of-arguments (1 . 1) 0)
  mu4e-mark-execute-all()
  mu4e-view-marked-execute()
  funcall-interactively(mu4e-view-marked-execute)
  call-interactively(mu4e-view-marked-execute nil nil)
  command-execute(mu4e-view-marked-execute)
```
Sometimes it works...  A bit puzzled.

Chris00:
I recompiled it and now I only have the error...

djcb:
Guess the "interactiveness" doesn't propagate.. anyway, the optional should fix that, thanks!

Issue #1977, opened 2021-04-12
Point is moved in the active frame when (mu4e~proc-index) finishes


Chris00:
## Expected or desired behavior

The point in the active frame is not moved when the frame is not the one where `*mu4e-headers*` is.

## Actual behavior

The point in the active frame is moved, regardless of which frame is currently active.

## Steps to reproduce

Open two frames.  Run mu in one and update the index.  While the update takes place, focus the second frame.  At the end of the update, the cursor in the active buffer of that frame will be moved (towards the beginning).

## Versions of mu, mu4e/emacs, operating system etc.

Mu: commit b825b563
Emacs 27.1

djcb:
Hmmm, I think I've seen something like this, but am unable to reproduce with these steps.

Chris00:
The culprit is [this line](https://github.com/djcb/mu/blob/master/mu4e/mu4e-headers.el#L843):
```elisp
(set-window-point (get-buffer-window) pos)
```
Not sure what it is for.

Chris00:
`(get-buffer-window)` returns `nil` which says to `set-window-point` to use the currently selected one.

PR #1978, opened 2021-04-12
mu4e-headers: Set the point in the correct frame after search

Chris00:
Fixes https://github.com/djcb/mu/issues/1977


djcb:
Oh, great, thanks!

PR #1986, opened 2021-04-22
Remove unneeded `or` clause in mu4e-view-save-attachments

thierryvolpiatto:
Was not removed with commit 55d8d20771405465f4d75649eab7b8a462cc4e31.


djcb:
Merged, thanks!

Issue #1984, opened 2021-04-22
mu crash during error printing


derek-zhou:
I am seeing segfaults in mu, and I traced it down to this line:
https://github.com/djcb/mu/blob/11fc99168b5c28313af884f6b80ea526e1dc1b30/lib/utils/mu-error.hh#L96
Supposedly, it should call:
https://github.com/djcb/mu/blob/11fc99168b5c28313af884f6b80ea526e1dc1b30/lib/utils/mu-utils.cc#L202
but on my machine (ia32 linux), it is calling:
https://github.com/djcb/mu/blob/11fc99168b5c28313af884f6b80ea526e1dc1b30/lib/utils/mu-utils.cc#L214
Therefore, it crashes. 
This problem only happens in ia32, not in AMD64. It is likely the `va_list` type is not a pointer type in AMD64. This issue can be simplified to the following example:
``` c++
#include <stdio.h>
#include <stdarg.h>

void format(const char *fmt, ...) {
  va_list args;
  va_start (args, fmt);
  vprintf(fmt, args);
  va_end (args);
}

void format(const char *fmt, va_list args) {
  vprintf(fmt, args);
}

int main() {
  format("%s\n", "hello");
}
```
The above program will crash in ia32 Linux, and will work fine in AMD64 Linux. 
Using C++ function overloading together with `va_list` has high risk. I propose to change second `format` to `vformat` to avoid any confusion. I can make a PR for this.

djcb:
Oh, good catch! Yes, a PR would be great.

derek-zhou:
I would love to; however, I have trouble building mu at git HEAD. automake does not work anymore, so I use meson. `meson builddir` went smoothly, but `meson compile` gives me this:
```
derek@roastidious:~/projects/mu$ cd builddir/
derek@roastidious:~/projects/mu/builddir$ meson compile
ninja: Entering directory `.'
[36/110] Compiling C++ object lib/libmu.a.p/mu-store.cc.o
../lib/mu-store.cc: In member function ‘void Mu::Store::set_dirstamp(const string&, time_t)’:
../lib/mu-store.cc:484:71: warning: format ‘%zx’ expects argument of type ‘size_t’, but argument 4 has type ‘time_’ {aka ‘long int’} [-Wformat=]
         const std::size_t len = g_snprintf (data.data(), data.size(), "%zx", tstamp);
                                                                       ^~~~~  ~~~~~~
[80/110] Generating mu4e_context_el_el with a custom command
[...]

PR #1987, opened 2021-04-22
seperate Mu::format and Mu::vformat

derek-zhou:
All the call site that issues a `va_list` in the second argument are converted to use `vformat`.
There is also a unrelated type cast to suppress a compiler warning.


derek-zhou:
forgot to mention this is for:
https://github.com/djcb/mu/issues/1984

djcb:
Merged, thanks!

PR #1991, opened 2021-04-26
main-view: make the bullet char a variable

seanfarley:
There has been some code from Doom to prettify the main view of mu4e
from an asterisk to a unicode symbol. This has been difficult because
`mu4e~main-redraw-buffer' isn't easy to advise or hook into since it
directly inserts an "*".

This patch creates a new variable `mu4e-main-bullet-char' so that
third-parties can use their own bullet character.

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -81,6 +81,10 @@ no unread messages.")
   "Keymap for the *mu4e-main* buffer.")
 
 (defvar mu4e-main-mode-abbrev-table nil)
+
+(defvar mu4e-main-bullet-char "*"
+  "Character to use for bullet points in the mu4e-main view.")
+
 (define-derived-mode mu4e-main-mode special-mode "mu4e:main"
   "Major mode for the mu4e main screen.
 \\{mu4e-main-mode-map}."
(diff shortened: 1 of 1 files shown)


seanfarley:
Here is a screenshot:

Screen Shot 2021-04-25 at 9 24 02 PM

djcb:
Thanks for the effort, but this is a bit too much code churn for something fairly unimportant (to me at least), also, this doesn't really scale for all the other characters people may want some replacement for. So sorry, have to decline this.

seanfarley:
> Thanks for the effort, but this is a bit too much code churn for something fairly unimportant (to me at least), also, this doesn't really scale for all the other characters people may want some replacement for. So sorry, have to decline this.

I can understand that but there's a lot of effort going on in third-party repos to try and theme some of `mu4e` (mostly with unicode). Is there some path forward with theming so that the monkey-patching can stop?

djcb:
The best would be for people care less about tweaking such things, but I have no illusions in that regard :-)

What I _might_ do at some point (time permitting) is make the main screen an org-mode file (well, org-mode with a few mu4e-specifics), and allow people to use a different one from the default.

peterhoeg:
With all due respect, this PR is self-contained and while I don't particularly care very much about how this looks either, that doesn't change the fact that a lot of people do. To me this PR seems like an easy way to make things customizable for those who care with very little down-side.

LemonBreezes:
> The best would be for people care less about tweaking such things, but I have no illusions in that regard :-)
> 
> What I _might_ do at some point (time permitting) is make the main screen an org-mode file (well, org-mode with a few mu4e-specifics), and allow people to use a different one from the default.

Any information on this? Has this idea been cancelled?

djcb:
No nothing planned for this, and don't think it will change in the foreseeable future.

PR #2005, opened 2021-05-02
Insert text without its text properties when replying

### LINKED ISSUE #2004: Disable buttons properties when replying

Do not have buttons properties when editing a reply.

## Actual behavior

The function `mu4e-view-message-text` is returning `(buffer-string)`, and `mu4e~draft-cite-original` use this function when replying.
When a message contain a button e.g. `[[End of PGP Encrypted Part]]` this button text is inserted in reply buffer.
Even when deleting this text the text properties stay (probably because of some sticky property), in consequence when the cursor is on the position of button the keymap prop remain and for example RET continue scrolling instead of inserting a new line.
A possible solution is inserting text without text properties like this:

```diff
diff --git a/mu4e/mu4e-view-gnus.el b/mu4e/mu4e-view-gnus.el
index 308aa6b7..d919e1ed 100644
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -70,7 +70,7 @@ etc."
     (insert-file-contents-literally
      (mu4e-message-field msg :path) nil nil nil t)
     (mu4e~view-render-buffer msg)
-    (buffer-string)))
+    (buffer-substring-no-properties (point-min) (point-max))))
 
 (defun mu4e-action-view-in-browser (msg)
   "Show current message MSG in browser, if it contains an html body."
```

## Steps to reproduce
Reply to a mail with such buttons and move your cursor on button press RET.

Mu: master
Emacs: 28

## Any other detail

E.g. are you using the gnus-based message view?

Yes (buttons are gnus buttons).

thierryvolpiatto:
Fix #2004.

### DIFF
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -70,7 +70,7 @@ etc."
     (insert-file-contents-literally
      (mu4e-message-field msg :path) nil nil nil t)
     (mu4e~view-render-buffer msg)
-    (buffer-string)))
+    (buffer-substring-no-properties (point-min) (point-max))))
 
 (defun mu4e-action-view-in-browser (msg)
   "Show current message MSG in browser, if it contains an html body."


djcb:
Ah yeah, actually it shouldn't include those buttons in the first place, but this fixes the immediate problem. Thanks!

PR #2007, opened 2021-05-02
Fix #2006 by using cl-loop in mu4e-action-view-in-browser instead of seq

### LINKED ISSUE #2006: mu4e-action-view-in-browser is not loaded because of seq not required (probably)

Be able to add to my config:

```elisp
(add-to-list 'mu4e-view-actions
            '("ViewInBrowser" . mu4e-action-view-in-browser) t)
```
and hit "a V" to browse a mail in browser.

## Actual behavior

Actually I have an error something like "void function mu4e-action-view-in-browser".
If I modify mu4e-action-view-in-browser function to not use seq like this:

```diff
diff --git a/mu4e/mu4e-view-gnus.el b/mu4e/mu4e-view-gnus.el
index d919e1ed..6cd064fb 100644
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -79,12 +79,9 @@ etc."
     (insert-file-contents-literally
      (mu4e-message-field msg :path) nil nil nil t)
     (run-hooks 'gnus-article-decode-hook)
-    (let ((header (mapconcat 'identity
-            (seq-filter (lambda(hdr) hdr)
-                        (seq-map (lambda(field)
-                                   (when-let ((val (message-fetch-field field)))
-                                     (format "%s: %s" (capitalize field) val)))
-                                 '("from" "to" "cc" "date" "subject"))) "\n"))
+    (let ((header (cl-loop for field in '("from" "to" "cc" "date" "subject")
+                           when (message-fetch-field field)
+                           concat (format "%s: %s\n" (capitalize field) it)))
           (parts (mm-dissect-buffer t t)))
       ;; If singlepart, enforce a list.
       (when (and (bufferp (car parts))
```

and recompile mu, I have no error.
An alternative would be to require seq but it is much cleaner to use cl-loop here (simpler as well).

For bug-reports, if applicable, include error messages, emacs stack traces, example messages
etc. Try to be as specific as possible - when do you see this happening? Does it happen always?
Sometimes? How often?

## Steps to reproduce

Try to add browser action to mu4e-view-actions and restart emacs, then try "a V" from a mail buffer.

mu: master
emacs: 28

## Any other detail

E.g. are you using the gnus-based message view?

yes.

thierryvolpiatto:
Using cl-loop is much simpler and readable, than using mapconcat+seq+seq.
Also it avoids requiring seq (which is not done actually).

### DIFF
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -79,12 +79,9 @@ etc."
     (insert-file-contents-literally
      (mu4e-message-field msg :path) nil nil nil t)
     (run-hooks 'gnus-article-decode-hook)
-    (let ((header (mapconcat 'identity
-            (seq-filter (lambda(hdr) hdr)
-                        (seq-map (lambda(field)
-                                   (when-let ((val (message-fetch-field field)))
-                                     (format "%s: %s" (capitalize field) val)))
-                                 '("from" "to" "cc" "date" "subject"))) "\n"))
+    (let ((header (cl-loop for field in '("from" "to" "cc" "date" "subject")
+                           when (message-fetch-field field)
+                           concat (format "%s: %s\n" (capitalize field) it)))
           (parts (mm-dissect-buffer t t)))
       ;; If singlepart, enforce a list.
       (when (and (bufferp (car parts))


djcb:
Thanks, generally not a fan of CL-style `loop`, but here it does make things clearer.

Issue #1995, opened 2021-04-26
incorrect thread parent selected when actual parent doesn't exist and messages have identical subjects


nicolasavru:
I have the following threaded set of messages (digits were replaced with #s):
2021-04-22 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
2021-04-19 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####

What is particularly bad about this threading is how messages from 2021-04-18 are "replies" to a message from 2021-04-19.

They have these references headers (respectively):
<no References header>
<no References header>
References: <+1646#######.0de1dfc5259c49347621104b1bee58fc0197a0b9@txt.voice.google.com>
References: <+1646#######.0de1dfc5259c49347621104b1bee58fc0197a0b9@txt.voice.google.com>
References: <+1646#######.0de1dfc5259c49347621104b1bee58fc0197a0b9@txt.voice.google.com>

They are all from the same sender and have the same subject.

Expected behaviour is one of the following threading arrangements:
The two messages without References headers in their own threads, the oldest of the messages sharing a References header becomes the parent.
2021-04-22 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
2021-04-19 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####

or

The two messages without References headers in their own threads, the messages sharing a References in a thread with a dummy parent.
2021-04-22 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
2021-04-19 S                 Foo Bar (SMS)     New text message from Foo Bar (###) ###-####
<dummy>
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####
2021-04-18 S                 Foo Bar (SMS)     |>New text message from Foo Bar (###) ###-####
[...]

PR #2008, opened 2021-05-02
threads: fix flag handling of orphan children

nicolasavru:
Even if a container has no query_match, it may still have children that need to
have flags applied.

This was broken by 6ff9829ce1e57e9f7fb4933d1389f4fd81dbd5dc.

Fixes: #1995


djcb:
Oh yes, good catch! Thanks.

PR #2016, opened 2021-05-20
mu4e-view: remove <home> and <end> bindings

Profpatsch:
Users usually have `<home>` and `<end>` bound in their configuration,
for Spacemacs the default is “move-beginning-of-line” and “move-end-of-line”.

The mu4e view mode should not rebind basic navigation keys like these.


djcb:
Merged, thanks. Wonder how they got in in the first place!

PR #2021, opened 2021-05-28
Add support for bug-reference auto-setup in mu4e-view-mode

tsdh:
The bug-reference mode in Emacs 28 has support for several kinds of auto-setup, one of them being for mail customizable by the variable `bug-reference-setup-from-mail-alist`.  Add mu4e support for that so that users can simply do

``` emacs-lisp
(add-hook 'mu4e-view-mode-hook #'bug-reference-mode)
```

and have it working.

Also squash one byte-compiler warning about the (at compile-time) undefined variable `gnus-article-buffer`.

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -540,6 +540,7 @@ Or go to the top level if there is none."
     (with-current-buffer  (mu4e-get-headers-buffer)
       mu4e~headers-last-query)))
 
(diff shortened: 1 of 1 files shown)


tsdh:
I have nothing to do with the failure on Mac OS, AFAICT...

djcb:
That seems useful. Merged, thanks!

Issue #2019, opened 2021-05-25
the license notices in mu4e files are wrong.


smohaorg:
mu4e files have a license which is not related to mu4e.

In the opening section of all files, it says,

> ;; This file is not part of GNU Emacs.
> 
> ;; GNU Emacs is free software: you can redistribute it and/or modify
> ;; it under the terms of the GNU General Public License as published by
> ;; the Free Software Foundation, either version 3 of the License, or
> ;; (at your option) any later version.
> 
> ;; GNU Emacs is distributed in the hope that it will be useful,
> ;; but WITHOUT ANY WARRANTY; without even the implied warranty of
> ;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
> ;; GNU General Public License for more details.
> 
> ;; You should have received a copy of the GNU General Public License
> ;; along with GNU Emacs.  If not, see <http://www.gnu.org/licenses/>.
> 

Which should be changed to
```

;; This file is not part of GNU Emacs.

;; mu4e is free software: you can redistribute it and/or modify
;; it under the terms of the GNU General Public License as published by
;; the Free Software Foundation, either version 3 of the License, or
;; (at your option) any later version.

;; mu4e is distributed in the hope that it will be useful,
;; but WITHOUT ANY WARRANTY; without even the implied warranty of
;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
;; GNU General Public License for more details.

;; You should have received a copy of the GNU General Public License
;; along with mu4e.  If not, see <http://www.gnu.org/licenses/>.

```

original report: https://social.linux.pizza/@redstarfish/106295917286141430

djcb:
Ah... can you make a PR?

PR #2022, opened 2021-05-29
mu4e/*: fix license notice at the start of source files

smohaorg:
At the start of mu4e/* files, the license has been wrongly attributed
to GNU Emacs instead of mu4e.

Fixes https://github.com/djcb/mu/issues/2019


djcb:
Merged -- thanks!

PR #2025, opened 2021-06-01
fix wrong configuration key reference

peterhoeg:
Looks like a copy/paste error.


djcb:
Oh, good catch! Thanks, merged.

Issue #1955, opened 2021-03-29
Provide a way to get all unread emails when running functions on `mu4e-index-updated-hook`?


rayw000:
It may be useful if `mu4e` could provide a built-in way to get all unread emails when running functions on `mu4e-index-updated-hook`. I didn't find it. I notice that [`mu4e-alert`](https://github.com/iqbalansari/mu4e-alert/blob/91f0657c5b245a9de57aa38391221fb5d141d9bd/mu4e-alert.el#L232-L254) starts its own `mu` process to retrieve emails like
```emacs-lisp
(defun mu4e-alert--get-mu-unread-emails-1 (callback)
  "Invoke a mu to get interesting emails and call CALLBACK with the results.

This is used internally by `mu4e-alert--get-mu-unread-emails' which throttles
the requests for unread emails."
  (let ((default-directory (expand-file-name "~")))
    (set-process-sentinel (apply #'start-process
                                 "mu4e-alert-unread-mails"
                                 (mu4e-alert--get-mail-output-buffer)
                                 mu4e-mu-binary
                                 (append (list "find"
                                               "--nocolor"
                                               "-o"
                                               "sexp"
                                               "--sortfield=d"
                                               "-z"
                                               (format "--maxnum=%d" mu4e-alert-max-messages-to-process))
                                         (when mu4e-headers-skip-duplicates
                                           (list "-u"))
                                         (when mu4e-mu-home
                                           (list (concat "--muhome=" mu4e-mu-home)))
                                         (split-string mu4e-alert-interesting-mail-query)))
                          (mu4e-alert--get-mail-sentinel callback))))
```

If we have more than one functions from different packages hooked on `mu4e-index-updated-hook`, each of them needs mail list,  we may need to start a process in every of them. Perhaps there are considerations that we don't temporarily store unread emails? Thank you!

djcb:
This doesn't really fit into how mu4e works, so I don't think this will be added anytime soon. Sorry!

PR #2026, opened 2021-06-02
Add new hook: mu4e-new-incoming-mails-hook

rayw000:
Functions hooked on this hook will be invoked when there are new incoming mails.
Meta information of new mails will be passed as argument to hooked functions.
This would be useful when you implement something like mail auto tagging etc.

This PR is inspired by https://github.com/iqbalansari/mu4e-alert

Fix: #1955


djcb:
Thanks for the effort, but I don't want mu4e to fork calls to mu binary; so have to decline this.

PR #2028, opened 2021-06-02
Fix bug with helm completing-read when saving attachments

andersjohansson:
Previously helm-comp-read-use-marked was bound also when completing on
a directory for saving attachments (when using a prefix argument). This
returned a list with the selected directory, which caused an error.

### DIFF
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -476,7 +476,6 @@ containing commas."
   (let* ((parts (mu4e~view-gather-mime-parts))
          (handles '())
          (files '())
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Anders Johansson ***@***.***> writes:

> Previously helm-comp-read-use-marked was bound also when completing on
> a directory for saving attachments (when using a prefix argument). This
> returned a list with the selected directory, which caused an error.

Good catch, you can use as well `(or (car-safe dir) dir)` to avoid
creating another let clause.

Thanks.

-- 
Thierry

djcb:
Merged, thanks!

PR #2036, opened 2021-06-11
Ensure mu4e~headers-move honor line-move-ignore-invisible

thierryvolpiatto:
This change is needed because forward-line doesn't honor this variable, more
generally visual lines.

Using `next-line` instead of `forward-line` allows this but it is more focused
on interactive use, so use `line-move` which handles visual lines without
warnings and return 0 or 1 just like `forward-line`.


djcb:
Merged -- thanks!

PR #2045, opened 2021-07-06
Fix #2044: broken header movement with visual-line-mode

### LINKED ISSUE #2044: mu4e-view-headers-next in mu4e-view-mode is broken with visual-line-mode
When visual-line-mode is enabled, it is impossible to view the next message from the message view buffer when the current header is visually wrapped. I believe this problem was introduced by c3c8e534 where forward-line was changed to line-move (which moves by visual lines by default, whereas the former does not).

**To Reproduce**
1. Enable visual-line-mode or global-visual-line-mode in the mu4e-headers buffer.
2. Resize your window so that the headers are wrapped.
3. Open a message.
4. Try to use mu4e-view-headers-next to switch to the next message.
5. Observe that the cursor moves to the next visual line, which is still the same message, and then jumps back to the beginning of the previous line.

**Environment**
OS: Arch Linux / Debian 11 (reproduced on both)
Emacs: 27.2 / 27.1
mu/mu4e: 1.5.13 r7034556a

DiGitHubCap:
This fixes the issue introduced by 50f6f539 where header navigation
would break when `visual-line-mode` was enabled. Previously,
`forward-line` was used in `mu4e-view-headers-next`, which disregarded
visual lines and moved by logical lines, but this was changed to
`line-move`, which moves by visual lines when `line-move-visual` is
non-nil (the default when `visual-line-mode` is enabled). Thus, when
the current header line was wrapped and a message was open in the
split view, `mu4e-view-headers-next` would move to the next visual
line in the headers buffer (the same message), and then jump back to
the start of the previous line, preventing switching to the next
message.

This would also throw off navigation when `mu4e-view-headers-next` was
used with a prefix argument, since it would move by visual lines and
not headers.

`line-move-visual` is therefore set to nil before using `line-move' to
prevent these issues.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1846,7 +1846,8 @@ docid. Otherwise, return nil."
   (cl-flet ((goto-next-line
              (arg)
              (condition-case _err
-                 (and (line-move arg) 0)
+                 (and (let (line-move-visual)
+                        (line-move arg)) 0)
                ((beginning-of-buffer end-of-buffer)
                 1))))
     (let* ((_succeeded (zerop (goto-next-line lines)))


thierryvolpiatto:
DiGitHubCap ***@***.***> writes:

> line-move-visual is therefore set to nil before using `line-move' to
> prevent these issues.

No, this fix the visual-mode issue but always move by logical lines, no
more by visual lines which was the purpose of this change to be able to
skip invisible lines with mu4e-thread-folding.
IOW with this change it is the same as using forward-line as before.

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/djcb/mu/pull/2045
>
> Commit Summary
>
>   • Fix #2044: broken header movement with visual-line-mode
>
> File Changes
>
>   • M mu4e/mu4e-headers.el (3)
>
> Patch Links:
>
>   • https://github.com/djcb/mu/pull/2045.patch
>   • https://github.com/djcb/mu/pull/2045.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.*


-- 
Thierry

DiGitHubCap:
> No, this fix the visual-mode issue but always move by logical lines, no more by visual lines which was the purpose of this change to be able to skip invisible lines with mu4e-thread-folding. IOW with this change it is the same as using forward-line as before.

Moving by visual lines and skipping invisible lines are two separate things, I think. The variable `line-move-ignore-invisible` should control whether line-move ignores invisible lines, and it is `t` by default. Have you actually tested whether this fix will break skipping invisible lines with mu4e-thread-folding?

Alternatively, using `forward-visible-line` might work if this doesn't work for you.

thierryvolpiatto:
DiGitHubCap ***@***.***> writes:

> The variable line-move-ignore-invisible should control whether
> line-move ignores invisible lines, and it is t by default.

Oh yes, you are right line-move-ignore-invisible seems enough, so yes,
let-binding line-move-visual is working, I was confused by the
documentation which is unclear for all this (line-move etc...) at least
for me.

Thanks.


> Have you actually tested whether this fix will break skipping
> invisible lines with mu4e-thread-folding?

Just tested, it is working, thanks.

-- 
Thierry

djcb:
Merged, thanks!

PR #2046, opened 2021-07-06
Assign `list-buffers-directory`

nagy:
for mu4e-header buffers, this displays the search query in
`M-x list-buffers` instead of showing an empty path column.


djcb:
Merged, thanks.

Issue #1911, opened 2021-01-23
mu4e-headers-toogle-include-related does not work anymore


nils-schween:
Hi everyone,

I just pulled the latest changes and unfortunately `mu4e-headers-toogle-include-related` does not work anymore.
I expected it to also include my answers in the threads. Typing `W` does not change anything. It "reloads" the headers, but the shown emails are they same. Probably you can reproduce this behaviour by pulling the latest changes and typing `W` in the header view.
Are you experiencing the same?
Thank you,
Nils

djcb:
It works for me, e.g.
```
lib/index % mu find hello | wc -l
5187
lib/index % mu find hello --include-related | wc -l
9396
```
so some more detail is needed. Does running regenerating the database help?

nils-schween:
Thanks for your quick reply.
Hmm ... that's strange. If I execute the same commands, I also do get different numbers. And second one is higher. Nonetheless, in header view in a thread, I do not see any emails I sent. I checked the `*Messages*`buffer and I do get the following error:
`
error in process sentinel: mu4e-alert--parse-mails: Invalid read syntax: ") or . in a vector"
error in process sentinel: Invalid read syntax: ") or . in a vector"
`
Deactivating mu4e-alert makes the error message disappear. But still, the threads do not include my own mails. Hmm ..

djcb:
The latter seems like a problem with mu4e-alert I guess... 

Anyway, can you try to re-generate the database, i.e., with `mu init`?

nils-schween:
Oh ... I directly edited my post, since I had forgotten to say, that I did re-generate the database. Probably you haven't seen the updated post. Unfortunately, it did not help. Still no email, which I had sent, in my threads.

djcb:
And... how many messages are in your inbox? Does it make a difference when you press `Q` (for a full-search rather than only 500)?

If that doesn't make a difference, it's best trying to reproduce with command-line searches, something like
```
mu find maildir:/inbox --include-related 
```
(change "/inbox" as needed).

nils-schween:
I just checked it, it does make a difference. If I toggle a full-search with `Q`, I can see my messages again.	
Here a rough list of my mailboxes and their respective size:
```
/maildir1/INBOX              (0/864)
/maildir1/otherFolder1     (0/126)
/maildir1/otherFolder2     (0/406)
/maildir1/otherfolder3      (0/56)
/maildir1/otherfolder4      (0/446)
[...]

PR #2048, opened 2021-07-11
query: fix include-related.

### LINKED ISSUE #1924: Missing rest of thread context when searching for (from:self)

When I search for threads I'm participating in with:

    (recip:alex.bennee OR recip:bennee.com)
  
I expect the rest of the thread to show up.
  
## Actual behaviour

Please describe the behavior you are actually seeing.

I get a mixed picture. For some threads it seems to show up but for some others I just get the email I sent. The email has a proper In-reply-to header and if I search for that id I get the whole thread in question.

This appears to be a result of the recent threading changes.

Current mu git HEAD, emacs native-comp branch, Linux Debian 10

nicolasavru:
mu-query.cc:
- make_related_enquire: don't include first query in qvec, we already have all
  thread IDs we need to query in thread_ids.
- run_related: always sort first query by date, as explained in the code
  comment.
- run_related: include qflags (in particular ascending vs descending) in
  leader_qflags.
- run_threaded: don't limit results to maxnum, that results in threads
  potentially being cut off.

mu-server.cc:
- output_sexp: don't limit results to maxnum so as to match the behaviour of
  mu find (and avoid cutting off threads).

Fixes #1924 and #1911.

### DIFF
--- a/lib/mu-query.cc
+++ b/lib/mu-query.cc
@@ -45,12 +45,11 @@ struct Query::Private {
 
         Xapian::Enquire make_enquire (const std::string& expr,
                                       MuMsgFieldId sortfieldid, QueryFlags qflags) const;
-        Xapian::Enquire make_related_enquire (const Xapian::Query& first_q,
-                                              const StringSet& thread_ids,
+        Xapian::Enquire make_related_enquire (const StringSet& thread_ids,
                                               MuMsgFieldId sortfieldid, QueryFlags qflags) const;
 
         Option<QueryResults> run_threaded (QueryResults&& qres, Xapian::Enquire& enq,
-                                            QueryFlags qflags, size_t max_size) const;
+                                           QueryFlags qflags) const;
         Option<QueryResults> run_singular (const std::string& expr, MuMsgFieldId sortfieldid,
                                            QueryFlags qflags, size_t maxnum) const;
         Option<QueryResults> run_related (const std::string& expr, MuMsgFieldId sortfieldid,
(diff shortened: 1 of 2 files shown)


lib/mu-query.cc
```diff
@@ -192,12 +190,12 @@ Query::Private::run_related (const std::string& expr, MuMsgFieldId sortfieldid,
...
-        const auto leader_qflags{QueryFlags::Leader};
+        const auto leader_qflags{QueryFlags::Leader | qflags};
         const auto threading{any_of(qflags & QueryFlags::Threading)};
 
         // Run our first, "leader" query
         DeciderInfo minfo{};
-        auto enq{make_enquire(expr, sortfieldid, leader_qflags)};
+        auto enq{make_enquire(expr, MU_MSG_FIELD_ID_DATE, leader_qflags)};
         const auto mset{enq.get_mset(0, maxnum, {},
                                      make_leader_decider(leader_qflags, minfo).get())};
```
nicolasavru, line 201: It just occurred to me... shouldn't we be gathering message IDs here, not thread IDs? Also, how are thread IDs populated in the db (as in, how are they generated, what actually are they) if threading is calculated after this point?

nicolasavru: Ah, I think I remember how this works now... there should be a 1:1 correspondence between message IDs and thread IDs, so this should work...

djcb:
I've played with for the last days, and it worked well for me (though admittedly, I never really encountered the original issue).

Anyway, merged.... Thanks for working on this!

PR #2049, opened 2021-07-12
Fix mu4e-headers-prev with invisible lines

thierryvolpiatto:
Skip invisible text at BOL possibly hidden by
the end of another invisible overlay covering
previous EOL.

This patch affects people using mu4e-thread-folding-mode but change nothing
when using mu4e as it is.


djcb:
Thanks, merged.

PR #2050, opened 2021-07-13
Allow a function as value of mu4e-split-view

tsdh:
This allows for splitting depending on geometry using `split-window-sensibly`
or more advanced home-brewn functions.  As a starter, which already suits my
needs better than the previously supported values, I use the following config:

```elisp
(defun th/mu4e-split-view ()
  (let* ((headers (mu4e-get-headers-buffer))
         (headers-win (and headers (get-buffer-window headers)))
         (view (mu4e-get-view-buffer))
         (view-win (and view (get-buffer-window view))))
    (cond
     (view-win view-win)
     ((one-window-p) (split-window-sensibly))
     ;; Use the tallest window.
     (t (car (sort (delq headers-win (window-list))
                   (lambda (a b)
                     (> (window-height a) (window-height b)))))))))
(setq mu4e-split-view #'th/mu4e-split-view)
```


tsdh:
I'll close this PR and come back later with a similar change to master.  For the time being (and my convenience), I've based this branch on the 1.6 release branch.

PR #2054, opened 2021-07-23
mu4e: Add support for gemini/gopher in mu4e~view-beginning-of-url-regexp

milouse:
This allow such URLs to be clickable and openable through Elpher (for example).

Gemini is a new kind of "Web" built around SSL and a light text markup language (learn more at https://gemini.circumlunar.space/). Gopher is an older internet protocol, still in use by some. Both protocol may be browse with Elpher, the same way than EWW or other web browser.

This commit make it possible to have links to these networks in email clickable again.

However, maybe it could be a better idea in the long term to follow values of `thing-at-point-uri-schemes` (in `thingatpt` package)?

### DIFF
--- a/mu4e/mu4e-view-common.el
+++ b/mu4e/mu4e-view-common.el
@@ -494,9 +494,10 @@ list."
   "Keymap used for the urls inside the body.")
 
 (defvar mu4e~view-beginning-of-url-regexp
-  "https?\\://\\|mailto:"
-  "Regexp that matches the beginning of http:/https:/mailto:
-URLs; match-string 1 will contain the matched URL, if any.")
+  "\\(?:https?\\|gemini\\|gopher\\|finger\\)://\\|mailto:"
+  "Regexp that matches the beginning of http:/https:/mailto: or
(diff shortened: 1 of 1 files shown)


djcb:
mu4e uses thing-at-point for URIs, but the we filter out anything that's not mailto/http/https before that.
what does emacs do with gemini: links?

milouse:
Currently, Emacs won’t do anything with gemini: links.

After a second look (mainly at thingatpt comment about `thing-at-point-uri-schemes` which states that protocol listed there are here only because they are recognised standard by IANA, and https://gitlab.com/gemini-specification/protocol/-/issues/6 where people thought it is urgent to wait before trying to published something about gemini), I think it is not yet the good time to add gemini support in various packages.

Also the next version of browse-url will make it more easily to declare a new protocol, thus we should wait for things to be clarified before adding anything not well defined.

However, finger and gopher are well established standard and are mentioned in the browse-url package. I’ll look forward to propose a patch to support them (through gophie or elpher) more easily. When done, it would be the right time to open again this MR.

Thank you very much for your quick answer !

PR #2077, opened 2021-08-03
Revive Xwidget view

AriT93:
extracted the 3 methods needed for viewing as message in an xwidget
in local testing this does not seem to cause any issues with the new
gnus view

### DIFF
--- a/mu4e/mu4e-view-xwidget.el
+++ b/mu4e/mu4e-view-xwidget.el
@@ -1,0 +1,104 @@
+;;; mu4e-view-xwidget -- part of mu4e, the mu mail user agent -*- lexical-binding: t -*-
+
+;; Copyright (C) 2011-2020 Dirk-Jan C. Binnema
+
+;; Author: Dirk-Jan C. Binnema <djcb@djcbsoftware.nl>
+;; Maintainer: Dirk-Jan C. Binnema <djcb@djcbsoftware.nl>
+
+;; This file is not part of GNU Emacs.
+
+;; mu4e is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


cfriesicke:
This PR has been open for a few months, and so far there is no discussion yet whether the feature will be included. Are there plans to integrate these changes into a future release?

Personally, I would like to have the Xwidget view feature; it feels a bit more "integrated" than spawning an external browser on top of Emacs. I tested the code provided in the above PR on the latest Emacs from git (29.0.50) after applying the code to the latest stable release of mu (1.6.10). I can confirm that everything seems to work, both in mu4e's "old" view and in the new "gnus" view.

So far, the only two things that maybe could be added is (a) some code that will `require` the new file/feature so that it can be discovered after loading mu4e, and (b) the new file needs to be added to the build configuration for the feature to install. I have already made these changes in my local repo; would it help if I provide this as well?

djcb:
Appreciate the work, but things work a little different in 1.7.x. It's perfectly possible to make an "action" for this, and use is with mu4e (esp 1.6.x).

I pushed a change for 1.7 to implement the feature; that seems to work, though not for all messages.

cfriesicke:
Thank you for reviving the xwidgets feature in the 1.7 branch!

I pulled the newest commit from master and it seems to work! As you already said, maybe not for all mails. I had one from Amazon where webkit throws some XML parsing error. But having that experimental feature is already great! I will try to test in the next days/weeks.

AriT93:
yay!  thank you for that.  I've just been loading this separately.  I'm still on 1.6 but will give 1.7 a try later this week.

djcb:
It's available in 1.8.x. Give it a try :-)

PR #2081, opened 2021-08-08
Customize the mu4e org link description

danielfleischer:
- There's a default function that extracts the subject but users can define their own function. 
- Added a mention in the documentation.
- Based on code from v1.4.15.


djcb:
Merged, thanks!

Issue #2013, opened 2021-05-11
[mu4e rfe] Toggle html view / part in new gnus backend


seanfarley:
With the old `mu4e-view-mode`, pressing `h` would toggle the html (using `shr` / `eww`) of the email. That seems to be missing with the gnus backend.

I did a little spelunking of the gnus code but I couldn't wrap my head around how it displays the html part.

djcb:
You can toggle it by clicking on the mime part in "Attachments:". A patch for hooking up `h` with that would be welcome.

seanfarley:
> You can toggle it by clicking on the mime part in "Attachments:". A patch for hooking up `h` with that would be welcome.

I tried but have no idea how to do it :-(

seanfarley:
🎉

PR #2087, opened 2021-08-14
mu4e: Add back the `h' keybinding to toggle HTML view in a gnus article

milouse:
This MR adds a simple function, which loops over the different parts of a message and makes use of gnus internal function to toggle the HTML part of it.

Fixes: https://github.com/djcb/mu/issues/2013


djcb:
Merged, thanks!  Note: I did modify a some impl details in a follow up commit, but it's your idea that counts :)

PR #2093, opened 2021-08-16
Create numbered backup of attached file when already exists #2090

### LINKED ISSUE #2090: [mu4e bug] mu4e-view-save-attachments directly overwrite files with the same name
**Describe the bug**
The function to save attachment in the new gnus view, `mu4e-view-save-attachments`, does not offer to change file name etc. when file with the same name exists, and just overwrite it.

**To Reproduce**
Open an email with an attachment; save the attachment; save for a second time.

**Environment**
Macos, Emacs 28, mu/mu4e 1.6.3

**Checklist**
- [ ] you are running vanilla emacs (i.e. without Doom, Evil, Spacemacs etc.) (otherwise, please try to reproduce without those
- [ ] you are running mu4e without any third-party extensions (otherwise, please try to reproduce without those)
- [ ] you are running either the latest 1.4.x release, or a 1.5.11+ development release (otherwise, please upgrade).

thierryvolpiatto:
Fix #2090.

### DIFF
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -506,13 +506,14 @@ containing commas."
                    do (mm-save-part-to-file
                        h (let ((file (expand-file-name f dir)))
                            (if (file-exists-p file)
-                               (let (newname
-                                     (count 1))
-                                 (while (progn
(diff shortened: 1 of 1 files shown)


djcb:
Merged -- thanks.

PR #2109, opened 2021-08-25
fix: mu4e-view-save-attachments void variable

JakDar:
This `_` looks like typo (correct me if I'm mistaken 😅).
`_arg` instead of `arg` here caused `Symbol’s value as variable is void: arg` for me when invoking `mu4e-view-save-attachments` .


djcb:
Don't think it's a _typo_, I was meant to signify that `_arg` was _unused_. However, that's not (or no longer) the case, so indeed it should be `arg` now. Thanks, merged!

PR #2110, opened 2021-08-25
Support maildir shortcuts without keys

### LINKED ISSUE #2106: [mu4e rfe] `mu4e-maildir-shortcuts' :key support string (or support without key)
**Is your feature request related to a problem? Please describe.**
I used to be a user of the mu4e-maildirs extension thing. It would list *all* maildirs. I am trying to use mu4e with it's builtin support for such somewhat now, and I've ran into an issue. I'll describe sort of why it is like that and then explain what I am wanting to do.

Lets say my email address is llmII@domain.tld and I run my own email server. Let's say I sign up for facebook. Since I own my email server I actually create an alias to llmII@domain.tld that is facebook@domain.tld. Now imagine I do this for every place I sign up. Now imagine I have a sieve script that recognizes "hey, this isn't to the main mail address" and takes the user part and automatically creates a folder for that place (now I've got my INBOX, and a facebook folder, and a github folder, and a google folder, and and and...).

So, I would like to list all these folders in the mu4e main view... but I have over 26 folders (we're going to pretend it won't let us use anything but alphabetic keycodes for purpose of the example). I write a bit of code to iterate over all folders each time mu4e finishes processing a mu response and generate an entry in `mu4e-maildir-shortcuts'. Lets say I have folders a maximum of 3 levels deep. Let's also say I have 2 email accounts, say another that is llmII@gmail.com. I define them in mbsync as llmII and gmail so they look like "$maildirroot/{llmII,gmail}/*". When I generate the shortcuts, I'd like for the idea was to take the first letter of the maildir root for the account ('l' or 'g'), then for the folder ('g' for github, 'f' for facebook) with a fallback of "try next character in second or 3rd part of path, until fail, then random assign).

Currently I can't do this, I'll have to pick and choose and somewhere I'll lose out on a mailbox I'd like to watch at least for unread/count purposes.

**Describe the solution you'd like**
I'd like to have `mu4e-maildir-shortcuts' either have ":key" optional or support using a string of keys.

llmII:
This adds support for `mu4e-maildir-shortuct` to include maildirs
without a `:key` for jumping to.

Fixes `mu4e~main-maildirs` to not error upon a maildir shortcut without
a key.

Fixes `mu4e-ask-maildir` to not include maildir shortcuts without keys.

Resolves #2106.

Justification:

Allows one to display more maildir's, their counts (unread/total), in
the mu4e main view. Before this if a user had more maildirs than the
total amount of displayable and usable single key shortcuts they would
have to pick and choose which ones they want displayed be they important
to them or not.

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -169,7 +169,9 @@ clicked."
            with longest = (mu4e~longest-of-maildirs-and-bookmarks)
            with queries = (plist-get mu4e~server-props :queries)
            for m in mds
-           for key = (string (plist-get m :key))
+           for key = (if (plist-get m :key)
+                         (string (plist-get m :key))
+                       " ")
            for name = (plist-get m :name)
            for query = (plist-get m :query)
            for qcounts = (and (stringp query)
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -364,8 +364,12 @@ all maildirs under `mu4e-maildir'."
     (if (not (mu4e-maildir-shortcuts))
         (substring-no-properties
          (funcall mu4e-completing-read-function prompt (mu4e-get-maildirs)))
-      (let* ((mlist (append (mu4e-maildir-shortcuts)
-                            '((:maildir "ther"  :key ?o))))
+      (let* ((mlist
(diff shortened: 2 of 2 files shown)


djcb:
Thanks! This has been now implemented in slightly different way.

Issue #2099, opened 2021-08-20
[mu4e rfe] Add a hook that is almost like `mu4e-index-updated-hook` except is guaranteed to run.


lockywolf:
**Is your feature request related to a problem? Please describe.**

I generally use mu from two interfaces, mu console, and mu4e. I fetch new mail with mbsync, and then I have two options:
1. If Emacs is running, I use `emacsclient  --eval '(mu4e-update-index)'
2. If Emacs is not running, I use mu index

The second operation is synchronous. Bash waits for it to finish, and then I can display new messages using the "notify-send" interface.

The first operation is not just anynchronous, but is not even guaranteed to produce any result, if there are no new messages.
If there _are_ new messages, I have a function in `mu4e-index-updated-hook`, that would create a lockfile, which bash would wait to appear. But `mu4e-index-updated-hook` is not run when there are no new messages.

I am, therefore, asking to add another hook `mu4e-index-updated-guaranteed-hook`, which would be run regardless of whether there are any new messages.

**Describe the solution you'd like**

The `mu4e-index-updated-hook` already exists. There should be not too hard to add one more hook nearby, with no "messages updated" condition.

**Describe alternatives you've considered**

1. Waiting for the xapian index to change. Not reliable, since some other thread may have run `mu4e-update-index`, before  me, and created a race condition.
2. Waiting for mu server to become multiuser. Not something that will happen soon.
3. Adding one more script to 'only show messages'. This leads to proliferation of scripts, and requires having a common configuration file for the "fetch script" and "notify script", hence increased complexity.
4. Only ever checking mail with  Emacs. Not really a good solution, since mail is not only used by Emacs.
5. Asking for an option to run mu4e-update-index synchronously. Not very nice, since it would freeze Emacs.

**Additional context**

I am monitoring the mail folder using IMAP IDLE, and when new messages appear, I fetch them with  mbsync and update index with mu index or with Emacs/mu4e-update-index.

llmII:
The way I did this was something like the below:

``` elisp
(defvar mu4e-finished-processing-hooks nil
 "Whenever mu4e finishes processing, regardless of index updated, trigger these.")

(defadvice! mu4e-finished-processing-handler (info)
   :after #'mu4e-info-handler
   (unless (eq (plist-get info :status) 'running)
     (run-hooks 'mu4e-finished-processing-hooks)))
```

[...]

PR #2111, opened 2021-08-25
Add new hook `mu4e-processing-finished-hook`.

llmII:
Fix: #2099

Serves as a way to guarantee a function gets ran anytime mu4e finishes
handling data from mu related to index operations, regardless of the
index actually being updated.

An example use case is when making use of goimapnotify to trigger when mu4e does an index by using file watchers. If goimapnotify is watching several boxes (different or same account(s)) it may generate that trigger multiple time (one user used file deletion as a trigger, I use file change and appending a single character as a trigger). Rather than creating hold down timers, what I do is just add a hook to the new hook that will remove itself once it runs, and the code that adds it refuses to add it if it's already present. This means that if goimapnotify triggers multiple times while mu4e is indexing, then mu4e will reindex after indexing, but not attempt to do so during an index. With my configuration this comes in handy as a way to manage indexing without lots of complexity.

I also use this to sync (in my doom config) `+mu4e-personal-addresses` which is used for selecting who you're sending as from a reply. The idea is to filter all contacts for contacts that fit my definition of "who am I" and add them to that list for selection upon reply later (which is an automatic selection).

### DIFF
--- a/mu4e/mu4e-utils.el
+++ b/mu4e/mu4e-utils.el
@@ -620,6 +620,12 @@ messages, it depends on `mu4e-index-updated-hook'. This can be
 used as a simple way to invoke some action when a message
 changed.")
 
+(defvar mu4e-processing-finished-hook nil
+  "Hook run when mu4e has finished processing data from the mu
+server. This can be used as a simple way to be certain something
+is run every single time mu4e finishes processing mu server
+data.")
+
 (make-obsolete-variable 'mu4e-msg-changed-hook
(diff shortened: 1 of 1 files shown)


llmII:
> Is this related to #2136 as well?

No - if I'm reading correctly they want a hook called once per mail processed. This is a hook that gets called once after mu handles indexing, regardless of if new messages were found during indexing or not. It only gets called after, not during, and not per message.

Issue #2116, opened 2021-08-28
[mu bug]


sje30:
**Describe the bug**

Pressing 'e' in an *Article* buffer, generates the message:

No attached files found

despite there being attachments that I can view when clicking on their buttons of the Attachments: line in the header.

Investigating this a bit deeper, I see the mime part structure differs to what is expected.  running

    (pp (mu4e~view-gather-mime-parts))

```
((7 #<buffer  *mm*-155504>
     ("application/octet-stream"
     (name . "Important Information Document.pdf"))
    base64 nil
    ("attachment")
    nil nil nil)
 (6 #<buffer  *mm*-298653>
    ("application/octet-stream"
     (name . "Primary IPID.pdf"))
    base64 nil
    ("attachment")
    nil nil nil)
 (5 #<buffer  *mm*-380027>
    ("application/octet-stream"
     (name . "Policy Wording.pdf"))
    base64 nil
    ("attachment")
    nil nil nil)
 (4 #<buffer  *mm*-158383>
    ("application/octet-stream"
     (name . "Policy Pack.pdf"))
    base64 nil
    ("attachment")
    nil nil nil)
 (2 #<buffer  *mm*-984499>
    ("text/plain"
     (charset . "utf-8"))
    base64 nil
    ("inline")
    nil nil nil))
```

where the (attachment) section is empty; compared with a normal message which has the filename element wihtin the attachment.

```
"((4 #<buffer  *mm*-216735>
    (\"application/pdf\"
     (name . \"UKRI-BBSRC_Scientific_Computing_in_the_Biosciences_Summary.pdf\"))
    base64 nil
    (\"attachment\"
     (modification-date . \"Wed, 18 Aug 2021 11:18:00 GMT\")
     (creation-date . \"Wed, 18 Aug 2021 11:18:00 GMT\")
     (size . \"55150\")
     (filename . \"UKRI-BBSRC_Scientific_Computing_in_the_Biosciences_Summary.pdf\"))
    \"UKRI-BBSRC_Scientific_Computing_in_the_Biosciences_Summary.pdf\" nil nil)
 (2 #<buffer  *mm*-226420>
    (\"text/plain\"
     (charset . \"us-ascii\"))
    quoted-printable nil
    (\"inline\")
    nil nil nil))
"

```

[...]

PR #2118, opened 2021-08-30
Attachments Saving Improvment

danielfleischer:
Different MIME configurations need a fallback logic. 

Instead of looking for the attachment "fname" in the structure
 ```elisp
("attachment" . (filename "fname"))
```
we find the same file name in the structure
```elisp
("mime-type" . (name . "fname"))
```
by searching, since we don't know the MIME type beforehand.

Should solve #2116 .

### DIFF
--- a/mu4e/mu4e-view-gnus.el
+++ b/mu4e/mu4e-view-gnus.el
@@ -491,7 +491,10 @@ containing commas."
                    #'completing-read-multiple))
         dir)
     (dolist (part parts)
-      (let ((fname (cdr (assoc 'filename (assoc "attachment" (cdr part))))))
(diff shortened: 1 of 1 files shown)


djcb:
Thanks!

You just got in the middle of the `master` refactoring, so the merge doesn't quite apply... can you make it instead use `release/1.6` as the target branch?

danielfleischer:
I see there's a lot of going on. Done.

It will have to be moved later from `mu4e-view-gnus.el` to `mu4e-view.el`.

djcb:
Merged -- thanks! I'll "port" it to mu4e-view in the new setup.

danielfleischer:
Great, you should probably close issue #2116.

PR #2120, opened 2021-08-31
Remove unnecessary faces

andersjohansson:
With the removal of the old mu4e-view, most of these faces were only used in compose, where it seems more reasonable to not remap the faces from message-mode (which can be suitably customized by users in any case).

The only face that was actually used in more places was `mu4e-header-key-face` which perhaps could be removed as well, but I didn’t do that for this initial suggestion.


djcb:
Thank you, this has been done!

andersjohansson:
Thanks, It could of course be argued that the face `mu4e-contact-face' and the function `mu4e~compose-remap-faces` could also be removed (see remaining diff). I don’t know if this would surprise any users.

Issue #2124, opened 2021-09-03
[mu4e rfe] Add regex option for `mu init --my-addresses=`


wilhelmy:
**Is your feature request related to a problem? Please describe.**
I have many email addresses in the style of `foo+bar@example.com` and `foo-blah@example.com`, which I use for sorting them into separate maildirs using sieve and being able to see which website leaked my data, among other things. I would like mu and mu4e to be able to recognize all of my email addresses without specifying all of them individually, since there are many.

**Describe the solution you'd like**
`mu init` has a `--my-address=` parameter which can be specified multiple times. I propose also adding an option to pass a regular expression.

**Describe alternatives you've considered**
I've considered passing them all, but it's annoying because I keep specifying new ones, and then would have to rebuild the index if I'm not mistaken.

ninadpchaudhari:
A new Mu user, but the insight might help. 
I do the same thing with Gmail. It just forwards the email as `foo+bar@gmail.com` to `foo@gmail.com` and adds a label of `bar` to it. Maybe this could be done in your mail server too. I imagine, forwarding emails and then seeing them at the same place is much better than accessing them individually.

Just a bad example (  I think ), if we consider email as an address and have a analogy of houses, these aliases are essentially apartment numbers in a big house! The mail is still delivered in the same mail box, then the roommates divide the same by themselves!

djcb:
It's already there; `mu init '--my-address=/foo-.*@example\.com/'`.

Chris00:
BTW, what kind of regexp is accepted?  Emacs/Perl/... ?  If I understand well, you use `g_regex_match` which accepts Perl regexes?

wilhelmy:
Thanks so far, this is great news! The manpage `mu-index(1)` (I'm running 1.6.5) says the following:
```
              <my-email-address> can be either a plain e-mail address (such  as  foo@example.com),  or  a
              regular-expression  (of  the  'Basic  POSIX'  flavor),  wrapped in 0 (such as 0foo-.*@exam-
              ple\.com).
```
The zeros are kind of an artifact of a syntax error in `mu-init.1.gz`, it's meant to be a slash. From what I can see this expression used is also not a BRE (Basic POSIX — deprecated AFAIK?) regex but ERE (Extended).

djcb:
The syntax error was fixed earlier this week. You can add anchors if you want or not, depending on what you want to match.

[...]

PR #2125, opened 2021-09-04
Document the use of regular expressions for --my-address

Chris00:
See https://github.com/djcb/mu/issues/2124


djcb:
Ah, thanks, but I already did the same this morning :-)

PR #2128, opened 2021-09-07
mu4e: Update the instructions to initialize mu

Chris00:
Using “mu init --maildir=~/Maildir” the yield

$ mu index
error: '~/Maildir' is not readable: No such file or directory

### DIFF
--- a/mu4e/mu4e.texi
+++ b/mu4e/mu4e.texi
@@ -424,13 +424,13 @@ pass that to the other commands as well.
 Assuming that your maildir is at @file{~/Maildir}, we issue the
 following command:
 @example
-  $ mu init --maildir=~/Maildir
+  $ mu init --maildir=$HOME/Maildir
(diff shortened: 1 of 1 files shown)


djcb:
Hmmm... at least `zsh` handles the '~'...; what shell are you using?

Chris00:
I use bash.  `~` seems not to be expanded in all situations:

```
$ echo x=~/Maildir
x=/home/trch/Maildir
$ echo --maildir=~/Maildir
--maildir=~/Maildir
$ echo --maildir= ~/Maildir
--maildir= /home/trch/Maildir
```

djcb:
I checked and actually `mu` is supposed to do the expansion itself when the shell doesn't... but it didn't work in the `meson` case... until (hopefully) now. Can you check? Thanks!

Chris00:
It works.

PR #2129, opened 2021-09-08
mu4e-headers: Avoid the obsolete command warning

Chris00:


djcb:
Those commands should go to the `mu4e-search-minor-mode`. Not sure if that works yet, but at least that's the correct place.

So instead of having to duplicate the command in various major-modes (header/view/main), they just enable the search minor mode. It's work-in-progress though.

Chris00:
That's a good idea.  I activated `mu4e-search-minor-mode` in header mode (I like to live dangerously 😀).  I was expecting that it overrides the <kbd>g</kbd> key but [it doesn't](https://github.com/djcb/mu/blob/master/mu4e/mu4e-search.el#L436-L445).  I guess it is because, in view mode, <kbd>g</kbd> is bound to `mu4e-view-go-to-url`.  So, to which map will <kbd>g</kbd> be attached?

djcb:
There are a few clashes (esp. when more things will be move to minor-modes) , we can solve those on a case-by-case basis.... for 'g' I think it's 'go-to-url' is more useful. Perhaps we need _two_ bindings in the minor mode map.

djcb:
Merged, thanks.

Issue #2132, opened 2021-09-12
[mu4e bug] Wrong buffer display after sending an email


Chris00:
**Describe the bug**
When composing an email (say replying to a message) when the `*mu4e-headers*` and `*Article*` buffers are displayed in the same frame, the `*draft*` buffer replaces the `*Article*` one (so far, so good).  When the email is sent, the `*mu4e-headers*` buffer takes the place of the `*draft*` one — thus 2 `*mu4e-headers*` buffers are displayed in a split view.  I'm expecting the `*Article*` buffer to replace back the `*draft*` one.

**To Reproduce**

Just send an email.

**Environment**
Emacs 27.1
Mu devel (commit #191f0f47)

**Checklist**
- [x] you are running vanilla emacs (i.e. without Doom, Evil, Spacemacs etc.) (otherwise, please try to reproduce without those
- [x] you are running mu4e without any third-party extensions (otherwise, please try to reproduce without those)
- [x] you are running either the latest 1.6.x release, or a 1.7.x development release (otherwise, please upgrade).

theophilusx:
Confirm I see same behaviour with latest dev HEAD.

djcb:
I think this has been with us for a long time.

The related code hasn't changed in ages (`mu4e-message-kill-buffer`), perhaps something changed with the gnus-view?

thierryvolpiatto:
I confirm the bug as well.  The reason is that `mu4e-get-view-buffer` returns a buffer-name instead of returning a buffer object, so `buffer-live-p` in `mu4e~switch-back-to-mu4e-buffer` fails.
Also using `gnus-article-buffer` is not the best thing to do if gnus is used along mu4e e.g. for news, in this case it would be better to use a `defconst` like `mu4e-headers-buffer-name` and use `get-buffer` just like `mu4e-get-headers-buffer` does.
Also both docstrings are wrong and misleading (hence this bug) because they say "get the name of the buffer" whereas they should be "get buffer...".

```diff
diff --git a/mu4e/mu4e-helpers.el b/mu4e/mu4e-helpers.el
index 7b533f89..7cf029a5 100644
--- a/mu4e/mu4e-helpers.el
+++ b/mu4e/mu4e-helpers.el
@@ -111,15 +111,16 @@ buffer list.")
   "Name of the buffer for message headers.")
 (defconst mu4e-embedded-buffer-name " *mu4e-embedded*"
   "Name for the embedded message view buffer.")
[...]

PR #2137, opened 2021-09-23
Fix `mu4e-get-view-buffer` (#2132)

thierryvolpiatto:
It actually returns a buffer-name whereas it should return a buffer
object which make `mu4e~switch-back-to-mu4e-buffer` fail because
`buffer-live-p` expects a buffer object.


djcb:
Merged, thanks!

PR #2141, opened 2021-09-25
Fix declare-function statements referencing the now-gone mu4e-utils

tsdh:
This fixes some `declare-function` statements referencing the now-gone `mu4e-utils.el`.

BTW, I've already sent you this patch by mail on August, 31st, but never got a response.  Did your spam filter classify my mails as spam, or is `git send-email` simply a way to contribute that you are not interested in, or is that fine but should go to the mailinglist rather than your private mail address to allow for broader discussion?

### DIFF
--- a/mu4e/mu4e-message.el
+++ b/mu4e/mu4e-message.el
@@ -31,21 +31,21 @@
 (require 'flow-fill)
 (require 'shr)
 
-(declare-function mu4e-error "mu4e-utils")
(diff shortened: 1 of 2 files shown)


djcb:
Merged, thanks.

tsdh:
@djcb Hey Dirk, thanks for merging!

What about the question about the patch gone unseen when I've sent it per mail?

PR #2142, opened 2021-09-25
Allow a function as mu4e-split-view value

tsdh:
This allows to have a function returning a window as value of
`mu4e-split-view`.  I'm using this patch since quite some time locally with the
following function.

```el
(defun th/mu4e-split-view ()
  (let* ((headers (mu4e-get-headers-buffer))
         (headers-win (and headers (get-buffer-window headers)))
         (view (mu4e-get-view-buffer))
         (view-win (and view (get-buffer-window view))))
    (cond
     ;; There's a view window, just use it.
     (view-win view-win)
     ;; Just one window, split sensibly, i.e., horizontally or
     ;; vertically depending on geometry.
     ((one-window-p) (split-window-sensibly))
     ;; Otherwise, use the tallest window.
     (t (car (sort (delq headers-win (window-list))
                   (lambda (a b)
                     (> (window-height a) (window-height b)))))))))

(setq mu4e-split-view #'th/mu4e-split-view)
```

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1269,16 +1269,16 @@ of `mu4e-split-view', and return a window for the message view."
       (kill-buffer (mu4e-get-view-buffer)))
     ;; get a new view window
     (setq mu4e~headers-view-win
-          (let* ((new-win-func
-                  (cond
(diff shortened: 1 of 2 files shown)


djcb:
Merged, thanks.

Issue #2139, opened 2021-09-24
Wrong buffer name for `gnus-article-buffer`


thierryvolpiatto:
**Describe the bug**

While looking at bug #2132 and writing PR #2137, I noticed that the value of `gnus-article-buffer` may change if I use gnus along mu4e for e.g. looking at News with Gmane.
After viewing Gmane news with gnus the `gnus-article-buffer` value become something like "*Article Gmane emacs-devel*", this because gnus modify original value in `gnus-article-setup-buffer`.

**To Reproduce**
1) Open News in gnus
2) Open any mail with mu4e
See the value of  `gnus-article-buffer` is no more "*Article*".

**Environment**

OS: LinuxMint

**Checklist**
- [X] you are running vanilla emacs (i.e. without Doom, Evil, Spacemacs etc.) (otherwise, please try to reproduce without those
- [X] you are running mu4e without any third-party extensions (otherwise, please try to reproduce without those)
- [X] you are running either the latest 1.6.x release, or a 1.7.x development release (otherwise, please upgrade).

thierryvolpiatto:
To fix it we can set it back to original value in `mu4e-view`, I can make a PR on top of #2137 once it will be merged:

```diff
diff --git a/mu4e/mu4e-view.el b/mu4e/mu4e-view.el
index 7227645e..25515b77 100644
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -622,6 +622,7 @@ marking if it still had that."
 
     (when (bufferp gnus-article-buffer)
     (kill-buffer gnus-article-buffer))
+    (setq gnus-article-buffer mu4e-view-buffer-name)
   (with-current-buffer (get-buffer-create gnus-article-buffer)
     (let ((inhibit-read-only t))
       (remove-overlays (point-min)(point-max) 'mu4e-overlay t)

```

djcb:
IIRC there were (are?) some parts in the Gnus code that expected "*Article*", so simply renaming it led to errors.

thierryvolpiatto:
"Dirk-Jan C. Binnema" ***@***.***> writes:

> IIRC there were (are?) some parts in the Gnus code that expected "Article", so simply renaming it led to errors.

Do you mean "Article" or "*Article*" the default value is the later and
after grepping gnus code I see no place where it can led to errors
e.g. string= etc... And because "*Article*" is anyway the default value
when Gnus start I don't see where it could clash.
[...]

PR #2150, opened 2021-09-30
Reset `gnus-article-buffer` in mu4e-view (#2139)

thierryvolpiatto:
This is needed when gnus is used along mu4e for other purpose like
reading News, in this case gnus rename `gnus-article-buffer` and we
end up with an unrelated buffer-name in mu4e.


djcb:
Merged, thanks!

PR #2152, opened 2021-10-03
mu4e-search-edit: Fix obsolete alias; update documentation

progfolio:
Replace obsolete variable alias with obsolete function alias for mu4e-headers-search-edit.
Update docstring for mu4e-view-search-edit to reference the new function.


djcb:
Merged, thanks.

Issue #2167, opened 2021-10-28
mu find with dashes in subject fail


c3d:
**Describe the bug**

On my system, `mu find` does not find anything in a subject if the query contains a dash sign (`-`). I could not find anything in `man mu-query` suggesting that the dash sign would have any special meaning.

The following shows some of the queries I just ran:

```
% mu --version
mu (mail indexer/searcher) version 1.7.3
[...]
% mu find "s:kata" | grep "kata-containers" | wc
   89290 1585726 13579143
% mu find "s:/kata.containers/" | wc
   93076 1659167 14116532
% mu find "s:/kata[-]containers/" | wc
   89290 1585726 13579143
% mu find kata-containers | wc
   42146  753417 6657244
% mu find "s:kata-containers" | wc
error: no matches for search expression
       0       0       0
```

The messages are found with a regexp like `kata.containers` or `kata[-]containers`, but they are not found with the straight `kata-containers` text.

**To Reproduce**

See the commands above.

**Environment**

Tested on macOS (BigSur 11.6.1) and Linux (Fedora 34). In both cases, `mu` had been built from source.

**Checklist**
- [X] you are running either the latest 1.6.x release, or a 1.7.x development release (otherwise, please upgrade).

lockywolf:
I tried regexps, but still failed to find anything with dashes.

This is a critical bug for me, because, like... a quarter of search queries I am doing are for dashed constructions. In particular:
```
Signed-off-by: Random J Developer <test@example.com>
```

Is there _any_ way I can work around this bug now?

djcb:
I'd just search without the dashes.

stsquad:
I use the ADJ keyword to match stuff close by:

```
              (:name "Mail addressed to me with git tags"
               :query "recip:/alex.bennee/ AND \( \(reviewed ADJ by\) OR \(signed ADJ off ADJ by\) \)"
               :hide-unread t
               :key ?g)
```

djcb:
There's no `ADJ` keyword though (`mu` doesn't use Xapian's query parser). You can just leave them out.

djcb:
Note, this case works with the new query parser (mu 1.11.20 and a bit before):
```
$ mu find subject:kata-containers
```
works just fine. It does _not_ work with the same word in the body-text though, since we only have the indexed form there.

PR #2168, opened 2021-10-28
parser: Deal with dash-separated query values

c3d:
Xapian uses `-` as a word separator, so a query like `mu find s:kata-containers`
would find no results.

This commit brute-forces it by replacing `-` with a `[-]` regexp and
transparently turning the query into a regexp query when this situation occurs.

Fixes: #2167

Signed-off-by: Christophe de Dinechin <christophe@dinechin.org>


wavexx:
I've been bitten by the word separator, however I'm not sure turning the search into a regexp is a good idea. It it would unsuspectingly turn a fast query into a slow one.

Wouldn't it make more sense to turn this transparently into a phrase search instead?

this-that => "this that"

djcb:
Thanks for the effort! #2167 has been fixed now in a different way, so  closing this.

Issue #1083, opened 2017-05-22
Messages without a Date header treated as 1970-01-01


yurkobb:
## Expected or desired behavior

I've noticed that some messages (namely notifications from drupal about available updates) arrive without a Date field. Mu then treats them as having arrived on 1970-01-01. However, it would probably make more sense to infer the date from one of the `Received:` headers to display them among recent messages.

## Actual behavior
Mu treats `Date`-less messages as having arrived on 1970-01-01, and as a result they don't show up in recent mail views.

## Steps to reproduce

Put a message without a `Date:` field into your maildir's `new` folder. Update your INBOX header view.

## Versions of mu, mu4e/emacs, operating system etc.
mu4e 0.9.18 C

djcb:
RFC 2822 specifies that messages should have a Date: to be proper messages (it's one of very view required fields), so I suppose Drupal is at fault here.

Of course mu *could* try to guess the date nevertheless, but it's a bit too much work for a corner-case. Probably best to complain with the Drupal folks.

Chris00:
It was not said clearly by the @yurkobb but maybe the problem he faces is that such messages do not appear in his inbox as they are older than the 500 first ones.  If so, one solution would be to use a date in the future when the `Date` field is absent.

yurkobb:
> RFC 2822 specifies that messages should have a Date: to be proper messages

I suspected that it is a violation of the standard. But should `mu` produce some kind of error then, instead of silently treating it as the unix epoch?

Currently, the behavior visible for the user is that some kind of interpretation still happens (date = 1970), although it may not be intentional. (An alternative interpretation of "no date" could be to always sort it newest, rather than oldest, although that is weird too).

djcb:
We could add a warning in the log, but generally we try to show messages, correct or not (and the incorrect ones are a tiny subset, at least in the mails that I receive).

Anyway, instead of inventing fragile hacks, it's must preferable to fix the problem at the source, by getting Drupal to send messages with a Date: header. Do you have a link to the bug report?

yurkobb:
I don't use Drupal anymore, so before submitting a bug report I would have to check the bug is still there.

I _think_ I got a couple other cases with messages arriving without a Date field, but have since deleted them.

Anyway, this is not a big deal, mu4e is highly usable for me as is.

[...]

PR #2175, opened 2021-11-03
Use file timestamp for indexing non-RFC-822-compliant messages where the Date: field is missing

berke:
This is a proposed fix for the closed issue : https://github.com/djcb/mu/issues/1083


djcb:
Hmmm... I think it's better to have a known value for "No date" than to invent a confusing/wrong version.

berke:
> Hmmm... I think it's better to have a known value for "No date" than to invent a confusing/wrong version.

Maybe you can add a flag that records where the date value came from: the Date: header, or the file timestamp.  If it's not the mail header, the flag can be displayed next to the date in mu4e.

djcb:
We're going to fix this in a different way (if at all), but thank you for the effort!

PR #2180, opened 2021-11-13
Ignore existing compiled files when byte-compiling

wavexx:
In the new build system we call batch-byte-compile manually for incremental compilation. Sadly the default behavior of emacs is to load the byte-compiled version of a source file, even if it refers to an older version. After a git update, this results in emacs loading the existing (old) source when other sources are compiled, which is incorrect.

A warning is emitted though:

```Source file ‘~/src/mu/build/mu4e/mu4e-view.el’ newer than byte-compiled file; using older file```

A simple fix is to use load-prefer-newer in the rule to always avoid loading stale byte-compiled files.

### DIFF
--- a/mu4e/meson.build
+++ b/mu4e/meson.build
@@ -98,6 +98,7 @@ foreach src : mu4e_all_srcs
 		command: [emacs,
 			  '--no-init-file',
 			  '--batch',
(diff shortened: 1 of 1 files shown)


djcb:
Merged, thanks!

PR #2191, opened 2021-12-15
Add eldoc support in headers views

tsdh:
When using a side-by-side split layout, i.e., headers left of message view, it
easily can happen that parts of the headers (like a longer subject) are not
visible without horizontal scrolling.  So eldoc to the rescue, which can echo a
customizable message (see `mu4e-headers-eldoc-format`) containing stuff like
subject, from, to, date, etc. in the echo area.

This PR just adds a suitable documentation function buffer-locally.  It does not enable eldoc.


djcb:
Looks useful, thanks. Merged.

PR #2194, opened 2021-12-25
utils: validate string before g_utf8_next_char()

l2dy:
> Before using this macro, use `g_utf8_validate()` to validate strings that may contain invalid UTF-8.

https://docs.gtk.org/glib/func.utf8_next_char.html


djcb:
Merged (outside github), thanks!

Issue #2182, opened 2021-11-21
RFE: Set up external links for HTML manual


josephmturner:
There's a broken link in the online manual:
[6.5 Signing and encrypting](https://www.djcbsoftware.nl/code/mu/mu4e/Signing-and-encrypting.html#Signing-and-encrypting) links to [(emacs-mime)Composing)](https://www.djcbsoftware.nl/code/mu/emacs-mime/Composing.html#Composing), which is broken.

Apologies if this is not the correctly place to report this.

djcb:
It's a link to an external manual, which sadly does not automaticallywork for the generated html docs. I suspect it could be fixed by setting up something like this:
  https://www.gnu.org/software/texinfo/manual/texinfo/html_node/HTML-Xref-Configuration.html

Volunteers welcome :)

josephmturner:
There no longer appears to be a broken link [here](https://www.djcbsoftware.nl/code/mu/mu4e/Signing-and-encrypting.html#Signing-and-encrypting). Shall we close this issue?

Thank you for mu4e! Great daily driver. :)

PR #2212, opened 2022-02-20
Fix broken link  in the online manual

averter:
Solves issue #2182 Please confirm that it works since I am not familiar with texi.


djcb:
Closing this... but yeah, something with htmlxref.cnf. would be welcome :)

PR #2224, opened 2022-03-25
Include <functional> for std::invoke

wavexx:
Fixes a build failure under gcc 12


djcb:
Thanks - was a good fix, but no longer needed.

PR #2225, opened 2022-03-26
meson: Install mu4e/mu4e-about.org

PRESFIL:
`mu4e` expects this file when activating the [A]bout link in `mu4e-main`
view, but when building the `mu-git` package from the AUR, this file
did not appear in the resulting package.

Now this file is not lost.


djcb:
Merged, thanks!

PR #2235, opened 2022-04-18
Last updated timestamp in main screen

danielfleischer:
Based on the `:tstamp` key in `mu4e-index-update-status` plist. 

Looks like this:

example


djcb:
Merged, thanks. Seems you read my thoughts :)

PR #2241, opened 2022-04-29
Add support for List-Id completion (list:<List-Id>)

tsdh:
As the title suggests, this commit adds completion support for the known List-Id values after the search keyword `list:` just like we already support email address completion after `to:`, `from:`, and friends.


djcb:
Merged, thanks.

PR #2242, opened 2022-04-29
Allow a function as mu4e-view-open-program

tsdh:
Allow `mu4e-view-open-program` to be set to a function instead of a string naming some external application.  In the function case, it must be a function receiving one single argument, the filename.


djcb:
Merged, thanks.

PR #2251, opened 2022-05-09
Add support for flag:personal & fix mu4e-contacts-info

tsdh:
I guess the change in `mu4e-search.el` is pretty self-explanatory.

Wrt. `mu4e-contacts-info`: `contacts` is a list with entries of the form `(rank
. full-mail-address)` where `full-mail-address` is a string like `"Name"
<mail@address.com>` and not a contact plist.


djcb:

PR #2253, opened 2022-05-11
Simplify mu4e-view-mode-map creation and fix `-` binding

tsdh:
`mu4e-view-mode-map` was initialized from a sparse keymap and eventually passed to `suppress-keymap` in order to remove bindings to `self-insert-command` which obviously make no sense in read-only buffers.  However, `suppress-keymap` also remaps all digits and the minus sign to make prefix args, i.e., pressing 8 would be equal to `C-u 8` and `-` got bound to `negative-argument` instead of `mu4e-view-mark-for-unflag` as defined.

While `suppress-keymap` has a `NODIGITS` arg to suppress the rebinding of digits and minus sign (which would also mean that pressing a digit would try inserting that digit and lead to an error), I chose the more straight-forward solution and start from an empty rather than a sparse keymap and added the `digit-argument` bindings for 0..9 explicitly.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -862,21 +862,20 @@ This is useful for advising some Gnus-functionality that does not work in mu4e."
     (apply func args)))
 
 (defvar mu4e-view-mode-map
-  (let ((map (make-sparse-keymap)))
+  (let ((map (make-keymap)))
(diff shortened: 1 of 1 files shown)


djcb:
Ah, good catch! Thanks, merged.

PR #2254, opened 2022-05-11
Make mu4e~headers-from-or-to respect its documentation

Chris00:
The variable `mu4e-headers-from-or-to-prefix` was not defined nor used, contrarily to what the documentation says.  (I like to customize the prefix to a non-textual symbol as I find it increases legibility for me.)

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -193,6 +193,13 @@ chronologically (`:date') by the newest message in the thread.")
   "Direction to sort by; a symbol either `descending' (sorting
   Z->A) or `ascending' (sorting A->Z).")
 
+(defcustom mu4e-headers-from-or-to-prefix '("" . "To ")
+  "Prefix for the :from-or-to field when it is showing,
+  respectively, From: or To:.  It is a cons cell with the car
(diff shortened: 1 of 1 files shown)


djcb:
It's more like the documentation wasn't updated for the new implementation... so there was (since 2012!) a 
```
(defconst mu4e-headers-from-or-to-prefix '("" . "To ")
```
which is not meant to be user-settable.

We can re-introduce that, but then please with the cons-pair as before, and as a `defcustom`.

Chris00:
> (defconst mu4e-headers-from-or-to-prefix '("" . "To ")
>
> which is not meant to be user-settable.
>
> We can re-introduce that, but then please with the cons-pair as before, and as a defcustom.

Sure, I'll do that when I have a bit of time.  May you remind me the purpose of the `car` of the pair?  Is it to have two possibilities according to `mu4e-use-fancy-chars`??

Chris00:
> May you remind me the purpose of the `car` of the pair?

Never mind, I found it in the Git history.

Chris00:
Done.  (Better now than much later! 😜)

djcb:
Looks good :) Merged, thanks.

Issue #2257, opened 2022-05-12
mu4e oldview 


scfrank-pri:
Thanks for providing the oldview mode. Since 1.6.10, with the new gnus article view, I have problems to see the content of my mails. 
The summary looks ok, the headers look ok - except for a number of nasty ^M, but there is no message content (maybe a problem of base64 encoding?) Clicking on the encoding, I see: 

```
no-conversion (alias: binary)
Do no conversion.
When you visit a file with this coding, the file is read into a
unibyte buffer as is, thus each byte of a file is treated as a
character.
Type: raw-text (text with random binary characters)
EOL type: LF
```

If the new mu4e requires additional gnus config, it would be really helpful to learn about it.
I have been using mu4e for years, without any problems. But this transition to gnus article simply knocks me out. No idea what else to report (gnu emacs 27.1).
Thanks for thinking of the normal user.

djcb:
Hmm, it suggest some emacs setup issue, but it's unclear. In any case, please attach an example message where this happens.

scfrank-pri:
I attach a short sample message (in a zip file) which renders OK in mu4e oldview, except for the ^M at each line end (which have never been visible before). The same message renders OK also in thunderbird. 
But in mu4e with the gnu article view I only see the headers, nothing else. 

Thanks for taking a look at it. 
[test.zip](https://github.com/djcb/mu/files/8688279/test.zip)

scfrank-pri:
I should add that in mu4e oldview the coding of the message is shown: U -- utf-8 (alias: mule-utf-8 cp65001)
whereas in mu4e with gnus article view it is, as reported above,
no-conversion (alias: binary) 
Do no conversion.

When you visit a file with this coding, the file is read into a
unibyte buffer as is, thus each byte of a file is treated as a
character.
Type: raw-text (text with random binary characters)
EOL type: LF 

But I have no idea how and where to fix this. If I do it directly in the emacs buffer (changing to UTF-8), this has no effect, except that I see U instead of =

My init.el already contains: 
(set-language-environment "UTF-8")
(prefer-coding-system 'utf-8)

Chris00:
[...]

PR #2258, opened 2022-05-13
Normalize newlines on messages

Chris00:
Fixes https://github.com/djcb/mu/issues/2257

This is just a suggestion.  There may be a better solution.


tsdh:
I'd rather not change the buffer contents unconditionally.  But you could add something to your setup.
```elisp
(add-hook 'mu4e-view-rendered-hook #'gnus-article-remove-cr)
```

djcb:
Thanks! This change has been included (a while ago), closing PR.

PR #2262, opened 2022-05-20
mu4e-org-store-link: prevent error when no message at point

progfolio:
The error can cause org-store-link to short circuit before it should (e.g. when org-store-link is called by org-babel-tangle-file).

### DIFF
--- a/mu4e/mu4e-org.el
+++ b/mu4e/mu4e-org.el
@@ -104,15 +104,12 @@ It links to the last known query when in `mu4e-headers-mode' with
 `mu4e-org-link-query-in-headers-mode' set; otherwise it links to
 a specific message, based on its message-id, so that links stay
 valid even after moving the message around."
-  (let ((view-mode-p (mu4e-is-mode-or-derived-p 'mu4e-view-mode))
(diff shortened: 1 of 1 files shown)


djcb:
Hmm, when you try to store a link in view/headers mode when there's no header, an error seems to be appropriate?

EuklidAlexandria:
The problem is that `org-store-link` loops over all org-store-link functions via `org-store-link-functions` regardless of the mode. Hence, `mu4e-message-at-point` is also called for non-mu4e buffers and returns an error. Maybe it should only be called if `view-mode-p` or `headers-mode-p` are non-nil?

progfolio:
> Hmm, when you try to store a link in view/headers mode when there's no header, an error seems to be appropriate?

Agreed. 

The main problem, as @EuklidAlexandria correctly noted, is the user-error signaled by `mu4e-message-at-point' short-circuits the calling function.

I've cleaned up the logic a bit, too.
If I understand correctly, `mu4e--org-store-link-message` calls `mu4e-message-at-point` itself. If there is no message, the error will be signaled there. No need to check ahead of time by calling `mu4e-message-at-point`.

djcb:
Ah, yes, good point.  Thanks, merged!

Issue #2265, opened 2022-05-28
[mu4e bug]mu4e-message-field should return a simple list


Chris00:
**Describe the bug**

Evaluating `(mu4e-message-field (mu4e-message-at-point) :to)` on a message form the mailing list returns `((:email "mudiscuss@googlegroups.com"))`.  I believe there is a list that needs to be removed.  Indeed, with this `mu4e-contact-email` returns `nil` on such output and so `mu4e-message-contact-field-matches` always fails.

**Environment**
Emacs: 27.1
Mu:  commit ffcb59bf

**Checklist**
- [x] you are running vanilla emacs (i.e. without Doom, Evil, Spacemacs etc.) (otherwise, please try to reproduce without those
- [x] you are running mu4e without any third-party extensions (otherwise, please try to reproduce without those)
- [x] you are running either the latest 1.6.x release, or a 1.7.x development release (otherwise, please upgrade).

djcb:
All contact-fields are _lists_ of contact-plists; and `mu4e-contact-email` takes one such plist. That works with `mu4e-message-contact-field-matches` as well, or?

i.e. with the message I got from Github:
```
M-:  (mu4e-message-contact-field-matches (mu4e-message-at-point) :from "Troestler") RET
(:name "Christophe Troestler" :email "notifications@github.com")
```
Does that not work for you?

Chris00:
﻿On 28 May 2022 at 09:57 -07, "Dirk-Jan C. Binnema" ***@***.***> wrote:
> All contact-fields are lists of contact-plists; and mu4e-contact-email takes one such plist. That works with mu4e-message-contact-field-matches as well, or?
>
> i.e. with the message I got from Github:
>
> M-: (mu4e-message-contact-field-matches (mu4e-message-at-point) :from "Troestler") RET (:name "Christophe Troestler" :email ***@***.***")
>
> Does that not work for you?

No, I get `nil`.

Chris00:
With the message from Github, `M-: (mu4e-message-contact-field-matches (mu4e-message-at-point) :to ".*@googlegroups.com")` works but `M-: (mu4e-message-contact-field-matches (mu4e-message-at-point) '(:from :to) ".*@googlegroups.com")` return `nil`.

Chris00:
With
```diff
@@ -139,10 +139,11 @@ address) regular expressions RX. If there is a match, return
 non-nil; otherwise return nil. RX can also be a list of regular
 expressions, in which case any of those are tried for a match."
[...]

PR #2267, opened 2022-05-28
Make mu4e-message-contact-field-matches accept a list of fields

Chris00:
Fixes https://github.com/djcb/mu/issues/2265


djcb:
Looks good, thanks!

PR #2291, opened 2022-07-04
Add optional name for maildir shortcuts

manufactory:
Useful to nicely display maildirs with long paths such as `lists/project/project_X`. If no `:name` is specified, keep the original behavior and use `:maildir` as name.


manufactory:
More dependent patches coming in

PR #2292, opened 2022-07-04
More customizable maildir shortcuts

manufactory:
1. Add a `:name` attribute to `mu4e-maildir-shortcuts` to display maildirs with long paths nicely
2. Add support for maildir shortcuts without a `:key`. Alternatively, allow `:key` to be a custom string.
I find a custom string useful for hinting at hydras.
This implements [#2110], which has conflicts with the current code.

I also added support for separator lines with titles to categorizes maildirs. Something like
```
* -- Private --
* [jp]  /private/inbox                                         (n/N)
* -- Work --
* [jw]  /work/inbox                                            (n/N)
```
I didn't include it since it's quite hackish, but I'd be prepared to clean it up and submit it if interested.

Sorry for opening/closing two other pull requests. I'm not too used to the Github workflow :)

### DIFF
--- a/mu4e/mu4e-main.el
+++ b/mu4e/mu4e-main.el
@@ -212,7 +212,7 @@ clicked."
            with longest = (mu4e--longest-of-maildirs-and-bookmarks)
            with queries = (plist-get mu4e--server-props :queries)
            for m in mds
-           for key = (string (plist-get m :key))
+           for key = (plist-get m :key)
            for name = (plist-get m :name)
            for query = (plist-get m :query)
            for qcounts = (and (stringp query)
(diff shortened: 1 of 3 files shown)


djcb:
Nice work.... I'll get back to this once the 1.9.x work starts. Thank you.

mu4e/mu4e-main.el
```diff
@@ -228,9 +228,12 @@ clicked."
            when (not (and mu4e-main-hide-fully-read (eq unread 0)))
            concat (concat
                    ;; menu entry
-                   (mu4e--main-action-str
-                    (concat "\t* [j" key "] " name)
-                    (concat "j" key))
```
llmII, line 233: Should wrap in `mu4e--main-action-str` (since at the least it'll make it look like the rest of mu4e's maildir shortcuts that are specified manually with `mu4e-maildir-shortcuts`)

llmII:
There are changes needed for `mu4e-ask-maildir` (like previously in my prior attempt at a PR). Otherwise you'll get `(wrong-type-argument characterp nil)` for maildirs where the `:key` is omitted.

mu4e/mu4e-main.el
```diff
@@ -228,9 +228,12 @@ clicked."
...
+		   (cond ((characterp key)
+			  (mu4e--main-action-str
+			   (concat "\t* [j" (string key) "] " name)
+			   (concat "j" (string key))))
+			 ((stringp key) (concat "\t* " key "  " name))
+			 (t (concat "\t*      " name)))
```
llmII, line 236: Why so many spaces after `\t*`? I'm only asking as the space amount seems arbitrary, but perhaps for you it gives you the specific way you want it to look or something (for me it doesn't look as great if I were to take 2 of the spaces out).

djcb:
Merging this, thanks!

@llmII: it looks good for me, but surely we can tweak the details later.

PR #2295, opened 2022-07-06
Add a custom maildir list to show in the maildir section.

i02sopop:
I have a bunch of maildirs that I don't want to have with a shortcut defined, but I want to be shown in the `Maildirs` section, so I'm adding a new variable, `mu4e-maildir-custom-list`, to define all the maildirs I want to be shown in the `Maildirs` section. It's my first PR, so comments are welcome :).

Signed-off-by: Ritho <palvarez@ritho.net>

### DIFF
--- a/mu4e/mu4e-bookmarks.el
+++ b/mu4e/mu4e-bookmarks.el
@@ -44,7 +44,6 @@
       :key ?t)
     ( :name "Last 7 days"
       :query "date:7d..now"
-      :hide-unread t
       :key ?w)
     ( :name "Messages with images"
       :query "mime:image/*"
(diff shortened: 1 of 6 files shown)


djcb:
Thank you for your work!

A number of people have expressed interest in a bit more control over maildirs / other shortcuts, so that's something to consider for 1.9.x.  I'd probably do so by adding some a property to the `maildir-shortcuts``.

i02sopop:
> Thank you for your work!
> 
> A number of people have expressed interest in a bit more control over maildirs / other shortcuts, so that's something to consider for 1.9.x. I'd probably do so by adding some a property to the `maildir-shortcuts``.

Nice, this is a very basic implementation that just works for my needs, but it probably can be improved a lot, so get from here whatever you need. Also, if you want me to improve this solution please share your ideas, since I'm not so fluent on lisp and I'd probably can make the code better with some indications :).

PR #2313, opened 2022-08-06
mu4e: Enable autoload generation in meson.

cireu:
Create autoload file for mu4e. So Emacs can autoload some entry function of mu4e


cireu:
Autoload generation will try to use the relative path to source directories and it's not relocatable. Trying to fix this.

djcb:
This has been implemented / merged now, so closing this one -- thanks!

Issue #2305, opened 2022-07-23
mu4e fails to recognize any calendar invitations


averter:
## Expected or desired behavior
mu4e correctly displaying calendar invitations and showing the corresponding buttons (accept, decline, etc). This used to work until I have recently changed email accounts +  moved to version 1.8.x.

## Actual behavior

Similarly to issue #1660 Calendar Invitations in vcal format are shown as regular attachments. This happens always. 

## Steps to reproduce

This is the relevant part of my mu4e configuration

```` lisp
;; Use gnus article view
(setq mu4e-view-use-gnus t)
(require 'mu4e-icalendar)
(mu4e-icalendar-setup)

(require 'org-agenda)
(setq gnus-icalendar-org-capture-file "~/org/NOTES.org")
(setq gnus-icalendar-org-capture-headline '("Calendar"))
(gnus-icalendar-org-setup)

````
## Versions of mu, mu4e/emacs, operating system etc.

I am using mu4e 1.8.6 and emacs 27.1. 

## Any other detail

I am using the gnus-based message view and have tried all the solutions suggested in #1660 .

**Comment by djcb** [comment]
@Chris00, any idea?

(Note, `mu4e-view-use-gnus` is long gone, no need to set it.)

**Comment by Chris00** [comment]
An example of message exhibiting the problem (an attached file, bit a copy/paste) would be good.  In the past, as I remember it, the problem was that messages were retrieved with Thunderbird which was adding `^M` at the end of lines.

**Comment by averter** [comment]
> An example of message exhibiting the problem (an attached file, bit a copy/paste) would be good.

Thanks for the quick reply. I've created a gist of an edited email available [here](https://gist.github.com/averter/e0d98f740c53cac3e85842b522978167)

The message might look a bit weird/distorted, since it has been sent to my work email and then I've used a thunderbird filter to automatically move it to my protonmail account.

**Comment by Chris00** [comment]
Your attachment is declared with the type
```
Content-Transfer-Encoding: base64
Content-Disposition: attachment; filename=calendar.ics
Content-Type: text/calendar; name=calendar.ics
X-Pm-Content-Encryption: on-import
```
It is not a request to know whether you are attending or not, contrarily to the following *calendar invitation*
```
Content-Type: text/calendar; charset="utf-8"; method=REQUEST
Content-Transfer-Encoding: base64
```
[...]

PR #2325, opened 2022-08-25
Fix empty 'To:' header on reply to icalendar invitation

mrvdb:
On reply to an ical invitation the To: header stayed empty, this fixes that.

The 'Reply-To' data is set if it is not there and if the emaill was different from the Organizer so
that the msg can be further treated by the draft construct functions.

This issue is briefly mentioned in issue #2305


djcb:
@Chris00 ^^ does this look okay?

mrvdb:
Just to make sure, is anything needed for this to get merged? Seemed like a fairly trivial change to me at the time. (I've been using it since then)

djcb:
Looks fine to me, but I'm using icalendar. @Chris00: is there merge-ready?

djcb:
No complaints from @Chris00, so merging this (well, cherry picked the change). Thank you!

PR #2329, opened 2022-09-03
mu4e-view: use `save-excursion' around toggling html

seanfarley:
Previously, the cursor would be moved to the first html attachment which
would break using mu4e keybindings (e.g. delete). With this change, the
cursor isn't moved at all, presumably the expected behavior.

I've targeted this for the 1.8 release branch because it seems like a bug but if
you want to apply to master instead, that's fine.


djcb:
Merged -- thanks. I'll merge it to master now, then cherry-pick it to 1.8.

seanfarley:
I see that 1.8.10 was just cut. Was this changeset cherry-picked?

djcb:
It was.

seanfarley:
Ah, I must have missed it; sorry for the noise!

PR #2330, opened 2022-09-04
mu4e-main: automatically update the mail counts after indexing

seanfarley:
It always struck me as a bit odd to force the user to refresh the main
view after updating and indexing their mail. This change creates a hook
that will simply call `revert-buffer' after an indexing completes.


djcb:
Merged; thanks!

PR #2332, opened 2022-09-07
m4e: remove disruptive characters from header fields

remvee:
Sometimes fields (especially subject fields) contain disruptive characters like new lines which will mess up the mu4e-headers view. This changeset introduces `mu4e~headers-sanitize-field-value` to replace all control characters by spaces.

Note, Unicode Left to Right Overrides do not cause any problems so they are ignored.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -653,12 +653,17 @@ space propertized with a `display' text property which expands to
       (mu4e~headers-truncate-field-precise field val width)
     (mu4e~headers-truncate-field-fast val width)))
 
+(defsubst mu4e~headers-sanitize-field-value (val)
+  "Replace control characters in VAL by regular spaces."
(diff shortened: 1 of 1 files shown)


remvee:
To reproduce the problem try the following raw subject header::
```
Subject: test=?utf-8?Q?=0A?=
```

seanfarley:
Ah! I ran into this issue as well, thanks so much for pinpointing it and submitting a fix!

djcb:
Merged; thanks!

djcb:
Hmmm had to revert this PR because it broke the threading display.

remvee:
Oeps sorry :flushed:, dunno how that happened.  I've forced pushed the right version just now.  Problem was the `mu4e~headers-truncate-field` function needs to go last because it adds padding which `mu4e~headers-sanitize-field-value` destroys..

@djcb can you reopen this PR and reconsider with the updated branch?  Or should I make a new PR?

PR #2343, opened 2022-10-02
build: add 'native-comp' option, defaulting to false

jian-lin:


jian-lin:
Elisp manual says
> compiling Lisp code natively always produces its byte code as well

I guess if `native-comp` is enabled, there is no need to do an extra byte compilation. But my meson knowledge is limited, so I leave this improvement to future contributors.

mu4e/meson.build
```diff
@@ -89,6 +89,21 @@ foreach src : mu4e_srcs
                           '--funcall', 'batch-byte-compile', '@INPUT@'])
 endforeach
 
+if get_option('native-comp')
+  target_func = '(setq native-comp-eln-load-path (list "' + mu4e_nativelispdir + '"))'
+  custom_target('native-comp',
+                build_by_default: true,
+                input: mu4e_srcs,
+                output: 'native-comp',
```
eli-schwartz, line 97: This produces an output file named literally "native-comp"? Seems odd.

What's the goal here, to produce compiled outputs in the build directory or to have them be installed? What happens if the user doesn't have root access during build, only during install? What happens if someone wants to use this but in conjunction with `$DESTDIR`?

jian-lin: > This produces an output file named literally "native-comp"?

No, output is not used here. I specify it because it is mandatory by meson syntax. The output eln files are directly put into mu4e_nativelispdir, which is named `native-lisp` and  is in the same dir as mu4e_lispdirdir, by emacs.

> What happens if the user doesn't have root access during build, only during install?

I only consider the usecase of installing using a package manager, have not considered manually building and installing. I have test it with nix, a system package manager like apt, and it works well. I think I have merged building and installing, will try to learn more meson and work on that later.

> What happens if someone wants to use this but in conjunction with $DESTDIR

I do not think this will be an issue since eln files are installed together with elc files. Assume elc files are installed correctly in this case, so do eln files.

jian-lin: Since my meson knowledge is limited, guides on how to seperate the building phase and installation phase is welcome.

theophilusx:
Eli Schwartz ***@***.***> writes:

> @eli-schwartz commented on this pull request.
>
> -------------------------------------------------------------------------------------------------------------------------
>
> In mu4e/meson.build:
>
>> @@ -89,6 +89,21 @@ foreach src : mu4e_srcs
>                            '--funcall', 'batch-byte-compile', ***@***.***@'])
>  endforeach
>  
> +if get_option('native-comp')
> +  target_func = '(setq native-comp-eln-load-path (list "' + mu4e_nativelispdir + '"))'
> +  custom_target('native-comp',
> +                build_by_default: true,
> +                input: mu4e_srcs,
> +                output: 'native-comp',
>
> This produces an output file named literally "native-comp"? Seems odd.
>
> What's the goal here, to produce compiled outputs in the build directory or to have them be installed? What happens if
> the user doesn't have root access during build, only during install? What happens if someone wants to use this but in
> conjunction with $DESTDIR?
>

Yes, I'm not sure about this as well. 

From what I've read in the threads on emacs-devel, all the native comp
data is put into a 'eln' directory within .emacs.d (or wherever
individual users have their init data). If your using an Emacs with
native comp enabled, Emacs will automatically natively compile any file
it loads where there isn't a native comp version.

I'm not sure there is anything which has to be done here. I recently
enabled native comp (I previously removed it because there were too many
issues and no visible performance improvement). Initial run generated
quite a bit of output as things were natively compiled and I got some
good (useful0 compilation warnings (mainly with my own extensions
etc). Since then, all has worked fine (though, I still notice next to no
performance improvement).

Personally, as native compilation is still not on by default (not even
in Emacs 29), I would leave this for AOT compilation and make no
change. Once native compilation is enabled by default in a released
version of Emacs, then revisit this topic. There is still stuff being
sorted out regarding native comp (for example, long thread this week on
enabling individual users to disable native comp on systems where
globally it is enabled)i, so the dust has not settled yet.

jian-lin:
> all the native comp data is put into a 'eln' directory within .emacs.d

Not all, only JIT eln files are in .emacs.d. This pr is mainly for AOT.

> I recently enabled native comp

Again, that is JIT and this pr is mainly for AOT.

> as native compilation is still not on by default

Many distributions have enabled native comp by default or provided an native comp version of emacs.

> I would leave this for AOT compilation and make no
change. 

I am a bit confused by you. With no change, there is no AOT for mu4e. This pr is to enable AOT for mu4e.

> long thread this week on
enabling individual users to disable native comp on systems where
globally it is enabled

I think setting `native-comp-deferred-compilation` to nil can disable JIT. No?

mu4e/meson.build
jian-lin: On splitting build and install phases, I can put build results in a temp dir. My question is how to move or copy a dir in install phase with custom_target.

I do not think we can specify the file name of an eln file.

eli-schwartz: > I only consider the usecase of installing using a package manager, have not considered manually building and installing.

No, you have considered a variety of use cases *other* than a package manager. :)

> I have test it with nix, a system package manager like apt

Nix is not a package manager like apt. Nix is a package manager unique for not doing things like any other Linux package manager (I am one of the major authors of a Linux package manager, I know how this works).

> I do not think this will be an issue since eln files are installed together with elc files. Assume elc files are installed correctly in this case, so do eln files.

Thereby demonstrating exactly why Nix is not a package manager like the usual ones (the ones I'm concerned about).

Nix doesn't package to DESTDIR, it packages directly to "manually building and installing into a custom --prefix".

That is the *only* reason why it works for Nix.

> since eln files are installed together with elc files

They very explicitly are not. https://github.com/djcb/mu/blob/4cceded947e9924b880ea35d79cec1095022f210/mu4e/meson.build#L71-L90

This (elc handling) is *extremely* different from your PR to do eln handling. The elc files are written into the ***build*** directory, the eln files are proposed in this PR to be written to the ***install*** directory.

It is of the utmost importance, regardless of build system (it has *nothing* to do with Meson) that files are *compiled* and produced in the build directory, and only installed by a separate install step. If it works correctly in that case (the case of "basically all Linux distros"), then it will work correctly for Nix as well, but the reverse is NOT true.

theophilusx:
Jian Lin ***@***.***> writes:

>  all the native comp data is put into a 'eln' directory within .emacs.d
>
> Not all, only JIT eln files are in .emacs.d. This pr is mainly for AOT.
>
>  I recently enabled native comp
>
> Again, that is JIT and this pr is mainly for AOT.
>
>  as native compilation is still not on by default
>
> Many distributions have enabled native comp by default or provided an native comp version of emacs.

Many might be a stretch. Quite a few are still bundling Emacs 27 and few
have changed the config from the default or provided a native comp
version that I have seen. There are 'user' level repositories and
snap/flatpak repositories which have native comp versions, but I don't
consider these part of a distro. 

>
>  I would leave this for AOT compilation and make no
>  change.
>
> I am a bit confused by you. With no change, there is no AOT for mu4e. This pr is to enable AOT for mu4e.

Sorry, meant JIT

>
>  long thread this week on
>  enabling individual users to disable native comp on systems where
>  globally it is enabled
>
> I think setting native-comp-deferred-compilation to nil can disable JIT. No?
>
It has been a very long thread and I'm really not across all of
it. However, it seems to have started because a) that setting was not
intuitive and b) it had some unexpected edge cases.

WRT the build, as far as I'm aware, most people either use a linux
pacakge manage or build from sources. There was a recent thread
regarding problems encountered by Debian with generating native comp
files - something to do with running as root and needing to write data
into root's home directory I think. I always build from the git
sources. I do the build as a normal user and then use sudo to do the
install. Quite a few people don't do the install and just run 'in place'
as you really only need to do install to make it available to multiple
users.

I imagine any PR willl need to work with normal build and install steps.

mu4e/meson.build
jian-lin: > What happens if the user doesn't have root access during build, only during install? What happens if someone wants to use this but in conjunction with $DESTDIR?

I have separated the build and install phases. As a result, your initial concerns should all be addressed.

meson.build
```diff
@@ -132,7 +136,12 @@ guile_dep = dependency('guile-3.0', required: get_option('guile'))
...
 emacs_name=get_option('emacs')
-emacs=find_program([emacs_name], version: '>=25.3', required:false)
+if get_option('native-comp')
+  emacs_min_version = '28.1'
+else
+  emacs_min_version = '25.3'
+endif
+emacs=find_program([emacs_name], version: '>=' + emacs_min_version, required:false)
 if not emacs.found()
   message('emacs not found; not pre-compiling mu4e sources')
```
jian-lin, line 146: If `native-comp=true` and emacs is not >= 28.1, is it better to exit?

jian-lin:
I convert it to draft because there is an issue: the path hash part of the eln file name (for mu4e-17b2ec84-f30a0d36.eln, "17b2ec84" is the path_hash and "f30a0d36" is content hash) outputted at the build of mu4e does not match the one emacs looks up when using mu4e, which makes this AOT native compilation useless.

I am digging into `comp-el-to-eln-rel-filename`.

Help are welcome.

jian-lin:
My initial goal is to enable AOT native compilation for NixOS. After digging into `comp-el-to-eln-rel-filename`, I figure out that it's better to do AOT native compilation in the post install phase instead of the build phase of mu4e to match eln filenames.

Thanks for your review.

Closing.

Thaodan:
> My initial goal is to enable AOT native compilation for NixOS. After digging into `comp-el-to-eln-rel-filename`, I figure out that it's better to do AOT native compilation in the post install phase instead of the build phase of mu4e to match eln filenames.
> 
> Thanks for your review.
> 
> Closing.

How did you do it in NIxOS? I want do something similar for pdf-tools on Arch.

jian-lin:
Here is my [pr][1].

[1]: https://github.com/NixOS/nixpkgs/pull/194281/files

Thaodan:
> Here is my [pr][1].
> 
> [1]: https://github.com/NixOS/nixpkgs/pull/194281/files
> 

Thanks

PR #2346, opened 2022-10-11
The "*Article*" buffers are now called "*mu4e-article*"

dkogan:
Now all mu4e buffer names begin with "*mu4e", which makes various external scripts, tools work better

### DIFF
--- a/mu4e/mu4e-helpers.el
+++ b/mu4e/mu4e-helpers.el
@@ -115,7 +115,7 @@ buffer list.")
   "Name of the buffer for message headers.")
 (defconst mu4e-embedded-buffer-name " *mu4e-embedded*"
   "Name for the embedded message view buffer.")
-(defconst mu4e-view-buffer-name "*Article*"
+(defconst mu4e-view-buffer-name "*mu4e-article*"
   "Name of the view buffer.")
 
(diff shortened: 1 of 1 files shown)


dkogan:
Hi. Can we make all mu4e buffer names begin with "*mu4e" ? I had some external scripts that broke when the article buffer was renamed to not do that.

Thanks!

djcb:
IIRC, the `gnus` code we're using depending on that buffer having the name it has.

dkogan:
Hi. FWIW I've been running mu4e like this for a while, and haven't hit
any obvious issues. Do you know what could break with this buffer rename?

djcb:
I remember some of the gnus code relying on it, is been a while though, can't remember the details. In any case, I 'd recommend using `mu4e-view-buffer-name` in scripts as well, so it'll work no matter what value it has.

As a last resort, note that you can actually _change_ the value of a `defconst` in your local setup with `setq`.

dkogan:
I can't use mu4e-view-buffer-name: these are external scripts that look at the X window titles. They don't run in emacs and aren't written in elisp. I can setq, but if you could merge this in the future to make the names consistent again, that'd be awesome. Thanks.

PR #2353, opened 2022-10-23
Use fake Message-ID when empty

dcolascione:
Previously, mu generated a fake message ID for messages without a Message-ID header.  This fake message ID allows these messages to show in an --include-related query. However, if a message contained a Message-ID header with the value equal to the empty string, we did not generate a fake message ID in the index, and consequently, these messages failed to appear in an --include-related query.  This change uses a fake message ID when the Message-ID header is absent _or_ empty.


djcb:
Thanks! Merged locally (with unit test).

PR #2356, opened 2022-10-28
mu4e: temporary reset browse-url-handlers in view in xwidget action

axeoman:
If `browse-url-handlers` is set `mu4e-action-view-in-xwidget` function tries to open url in the handler first instead of xwidget. This is happening because `browse-url` first checks `browse-url-handlers` and `browse-url-browser-function` after.


djcb:
Merged, thanks!

PR #2366, opened 2022-11-14
Avoid word-splitting regular expression matches

dcolascione:
Previously, we would conduct regular expression searches by enumerating all values of a given term, manually regex-matching each one against our search regular expression, remember all the term values that matched our regular expression, then do a big Xapian OR-query that matched any of those term values. In constructing this OR-query, however, we would split each term value on space and add a separate Xapian phrase search term for each resulting word. This approach worked fine most of the time, beacuse when we index a term, we index both each word in a term and the whole term by itself.

This word splitting produced false negatives in some matches, however, because Xapian and the Mu-level word splitting code do word splitting slightly differently and apply different transformations to the text while splitting.  (For example, Xapian transforms fancy Unicode apostrophes to ASCII apostrophes.)

This patch avoids the problem by not word splitting when constructing the big Xapian OR-query for finding the results of regular expression matching.

### DIFF
--- a/lib/mu-parser.cc
+++ b/lib/mu-parser.cc
@@ -267,7 +267,7 @@ Parser::Private::regex(const FieldInfoVec& fields,
 		for (const auto& field : fields) {
 			const auto terms = process_regex(field.field, rx);
 			for (const auto& term : terms) {
-				tree.add_child(Tree({Node::Type::Value,
+				tree.add_child(Tree({Node::Type::ValueAtomic,
 							FieldValue{field.id, term}}));
 			}
 		}
--- a/lib/mu-tree.hh
+++ b/lib/mu-tree.hh
@@ -80,6 +80,7 @@ struct Node {
 		OpAndNot,
 		OpNot,
 		Value,
+                ValueAtomic,
 		Range,
 		Invalid
 	};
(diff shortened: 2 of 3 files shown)


djcb:
Thanks, that looks useful. If merged/pushed (which Github sadly doesn't detect) and added some unit tests. So it seems
`subject:/foo’s/` matches now, but `subject:/foo’s bar/` still doesn't... is that expected?

dcolascione:
> Thanks, that looks useful. If merged/pushed (which Github sadly doesn't detect) and added some unit tests. 

Thanks!

> So it seems subject:/foo’s/ matches now, but subject:/foo’s bar/ still doesn't... is that expected?

Are you sure the apostrophes in the latter match exactly, byte-for-byte?

The single-word terms getting indexed are being normalized by Xapian's word generator, but the whole-field term we index manually is being indexed literally. Ideally we'd apply the same normalization to both kinds of indexed terms (probably in gx_utf8_flatten), but this patch doesn't do that.

djcb:
These are the current tests: https://github.com/djcb/mu/blob/master/lib/tests/test-mu-store-query.cc#L629

Guess we need the normalization at some point.

PR #2371, opened 2022-11-28
Teach mu4e-copy-thing-at-point about SHR links

vermiculus:
This allows one to copy the link behind a link-widget instead of using RET to open it.

I've been having good success with this patch on my system for the last several days. It's very handy -- I currently cannot open links directly into a browser with my setup.


djcb:
Merged - thanks!

PR #2376, opened 2022-12-12
Add variable for initial input to maildir selection

markk:
"/" was hardcoded as initial input to `mu4e-completing-read-function' when selecting 'other' maildirs. This can now be set by the user.


djcb:
Merged, thanks!

PR #2377, opened 2022-12-12
Preserve `mu4e~headers-view-win' when a search is (re-)executed

mickeynp:
Triggering a search -- directly, or indirectly through reindexing -- calls `mu4e-headers-mode` which kills all buffer-local variables including `mu4e~headers-view-win`, which causes any live view windows to lose their 'connection' to the now-updated headers buffer.


djcb:
Merged, thanks.

PR #2378, opened 2022-12-13
Bugfix/fix window and buffer issues

mickeynp:
WIP to fix the rough edges around window and buffer management.


mickeynp:
I believe this is also ready to go.

djcb:
Okay, merged. Thanks!

PR #2379, opened 2022-12-14
bug-reference: Add clarifying comment / checked-headers as variable

tsdh:
It just occurred to me that my bug-reference setup didn't work anymore with the
current master.  I activate `bug-reference-mode` in `mu4e-view-mode-hook`.  That
doesn't work anymore because the mode is activated now when the buffer is still
empty.  It works again when activating in `mu4e-after-view-message-hook`.  So I
added a clarifying comment.

Also I've made the headers whose values are checked to figure out a suitable
bug-reference setup a variable and added Reply-To there.

Well, I guess I'm the only user of that feature anyway... ;-)

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -1311,24 +1311,33 @@ the third MIME-part."
 
 ;;; Bug Reference mode support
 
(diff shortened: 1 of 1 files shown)


djcb:
Merged - thanks.

PR #2380, opened 2022-12-14
Re-add missing argument as `define-mail-user-agent' expects it

mickeynp:
`C-x m` with the user agent set to mu4e is broken. This undoes a fix I'd done earlier due to a byte compiler warning.

@djcb, can you think of a better way of handling this issue?

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -796,7 +796,7 @@ draft message."
 
 ;;;###autoload
 (defun mu4e~compose-mail (&optional to subject other-headers _continue
-                                    yank-action _send-actions _return-action)
+                                    _ yank-action _send-actions _return-action)
   "This is mu4e's implementation of `compose-mail'.
 Quoting its docstring:
 


djcb:
Basically your fix, but.... this should work?

```
@@ -796,7 +796,8 @@ draft message."
 
 ;;;###autoload
 (defun mu4e~compose-mail (&optional to subject other-headers _continue
-                                    yank-action _send-actions _return-action)
+                                    _switch-function yank-action _send-actions
+                                    _return-action)
   "This is mu4e's implementation of `compose-mail'.
 Quoting its docstring:
 
@@ -812,6 +813,9 @@ HEADER and VALUE are strings.
 CONTINUE, if non-nil, says to continue editing a message already
 being composed.  Interactively, CONTINUE is the prefix argument.
 
+SWITCH-FUNCTION, if non-nil, is a function to use to
+switch to and display the buffer used for mail composition.
+
 YANK-ACTION, if non-nil, is an action to perform, if and when
 necessary, to insert the raw text of the message being replied
 to. It has the form (FUNCTION . ARGS). The user agent will apply
```

mickeynp:
Ah yes, that works for me.

mickeynp:
Should be ready to go.

djcb:
Merged, thanks.

PR #2383, opened 2022-12-17
Change how *mu4e-main* is displayed and default to full frame

### LINKED ISSUE #2382: mu4e--main-view forces full frame window
I missed this change, but `mu4e--main-view` now calls `delete-other-windows` --- which completely decimates the existing window layout. It is not possible to configure it with `display-buffer-alist` and any customisation a user may have is ignored.

mickeynp:
Fixes #2382 by making `*mu4e-main*` use `display-buffer` actions to prefer a full frame.

### DIFF
--- a/mu4e/mu4e-window.el
+++ b/mu4e/mu4e-window.el
@@ -27,10 +27,8 @@
 (defconst mu4e--sexp-buffer-name "*mu4e-sexp-at-point*"
   "Buffer name for sexp buffers.")
 
-(defvar mu4e-main-buffer-name " *mu4e-main*"
-  "Name of the mu4e main buffer.
-The default name starts with SPC and therefore is not visible in
-buffer list.")
+(defvar mu4e-main-buffer-name "*mu4e-main*"
+  "Name of the mu4e main buffer.")
 
 (defvar mu4e-embedded-buffer-name " *mu4e-embedded*"
   "Name for the embedded message view buffer.")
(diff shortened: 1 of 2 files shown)


djcb:
Merged; thanks

mu4e/mu4e-window.el
```diff
@@ -267,6 +265,7 @@ for BUFFER-OR-NAME to be displayed in."
             ('(view . vertical) '((window-min-width . fit-window-to-buffer)))
             (`(,_ . t) nil)))
          (window-action (cond
+                         ((eq buffer-type 'main) '(display-buffer-full-frame))
                          ((and (eq buffer-type 'compose) mu4e-compose-in-new-frame)
```
djcb, line 269: Note, `display-buffer-full-frame` is emacs-29, but we should support emacs 26+... what can we use instead?

mickeynp: Backport it maybe:

```
(defun display-buffer-full-frame (buffer alist)
  "Display BUFFER in the current frame, taking the entire frame.
ALIST is an association list of action symbols and values.  See
Info node `(elisp) Buffer Display Action Alists' for details of
such alists.

This is an action function for buffer display, see Info
node `(elisp) Buffer Display Action Functions'.  It should be
called only by `display-buffer' or a function directly or
indirectly called by the latter."
  (when-let ((window (or (display-buffer-reuse-window buffer alist)
                         (display-buffer-same-window buffer alist)
                         (display-buffer-pop-up-window buffer alist)
                         (display-buffer-use-some-window buffer alist))))
    (delete-other-windows window)
    window))
```

djcb: Yeah, guess that's the best way. Pushed; thanks.

mickeynp: Yeah it's not perfect but the code is simple enough and it preserves the behaviour.

PR #2395, opened 2023-01-02
mu4e-view: don't delete window on view msg in single-window mode

bjcohen:
Don’t delete the existing buffer window when viewing a new message when `mu4e-split-view` is `single-window` or `nil`. This was causing the window layout to be trashed when navigating to a new message with `n` or `p` with a message open when in single-window mode.


djcb:
merged -- thanks!

PR #2401, opened 2023-01-04
Update mu4e.texi

sje30:
Suggestion to mention how to change contexts manually.

Feel free to ignore if you think it not relevant.  But I smiled when I found `C-c ;' in *Draft* buffers and think this should be noted somewhere.


djcb:
Merged, thanks.

PR #2408, opened 2023-01-12
mu4e: fix typo

progfolio:
s/mose/most

Also, Thanks for the recent modeline support. It obviates the need for the mu4e-alert package for me (which I had been maintaining a personal fork due to a lack of maintenance).


djcb:
Didn't apply anymore after latest changes, but I'll include it manually; thanks!

PR #2413, opened 2023-01-15
Multiword maildirs are not quoted when autocomplete is used (fixes #2…

### LINKED ISSUE #2389: [mu4e bug] Multiword maildirs are not quoted when autocomplete is used.
**Describe the bug**

In mu4e-search, the autocomplete for maildir does not quote/escape maildirs with a space. For example, it will autocomplete to `maildir:/path/to dir` instead of `maildir:"/path/to dir"` which is the desired search expression.

I think a potential solution is adding another exit-function [here](https://github.com/djcb/mu/blob/master/mu4e/mu4e-search.el#L388).

**How to Reproduce**

1. Open mu4e
2. `s` for mu4e-search
3. type in maildir: then `<tab>` for autocomplete
4. Select a multi-word maildir. A good example is the Sent Mail or All Mail folders from Gmail inboxes
5. Observed: No matching messages found because the maildir was not quoted.

**Environment**
Linux, mu/mu4e 1.8.11, Emacs 28.2 with Doom

**Checklist**
- [X] you are running either an 1.8.x release or the latest 1.9.x development release
- [X] you can reproduce the problem without 3rd party extensions (including Doom/Evil etc.)

tsdh:
This fixes #2389, i.e., now the mu4e search query completion offers maildirs with whitespace in their name quoted as `maildir:"Foobar/Junk Mail"`.

### DIFF
--- a/mu4e/mu4e-search.el
+++ b/mu4e/mu4e-search.el
@@ -389,7 +389,13 @@ status, STATUS."
    ((looking-back "maildir:\\([a-zA-Z0-9/.]*\\)" nil)
     (list (match-beginning 1)
           (match-end 1)
-          (mu4e-get-maildirs)))
+          (mapcar (lambda (dir)
+                    ;; Quote maildirs with whitespace in their name, e.g.,
+                    ;; maildir:"Foobar/Junk Mail".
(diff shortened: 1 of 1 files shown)


djcb:
Merged, thanks!

PR #2414, opened 2023-01-15
fix batch-size option in sub_init

derek-zhou:
I use mu on a resource constrained VPS, so I need to reduce batch-size from the default value, thus I found this bug.


djcb:
Oh good catch! Merged, thanks.

PR #2422, opened 2023-01-22
remove noexcept

husainaloos:
This change removes noexcept from MessagePart::looks_like_attachment and MessagePart::mime_type because the method content_type throws an exception.


djcb:
Hmmm, the `noexcept` doesn't mean that you can't have exceptions, just that you terminate when that happens.
But, perhaps we could make that clearer with `g_error`.

PR #2423, opened 2023-01-25
really allow flags to be empty in move command

derek-zhou:
Without this patch, I see the following error:

```
;; mu> (move :docid 1 :maildir "/.Archive")
[2b](:error 32818 :message "invalid flags ''")
```

flags should not default to "", which is not a valid flags string.


djcb:
Merged... this subtly changes the behavior, but I suppose it's the intended one.

PR #2426, opened 2023-01-31
mu4e: Allow completion for symbolic links in maildir

metaspace:
This patch allows mu4e to autocomplete when `mu4e-root-maildor` contains symbolic links to other maildirs. Without this patch mu4e can still work when directories under `mu4e-root-maildor` are symlinks, but it will not autocomplete properly when using `mu4e-search-maildir`.

Signed-off-by: Andreas Hindborg <nmi@metaspace.dk>


djcb:
Looks useful; merged. Thanks.

PR #2428, opened 2023-02-03
Only add bug-reference setup function after the feature was loaded

leotaku:
This PR aims to improve the support for bug-reference-mode introduced in #2021.

Currently, we might call `add-hook` on `bug-reference-auto-setup-functions` before `bug-reference` is loaded, which results in overriding the default value. I think it makes much more sense to add our hook to the list of hooks, which is what this PR aims to achieve.

@tsdh You might be interested in this change.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -1328,8 +1328,9 @@ GROUP-REGEXP and each header value against HEADER-REGEXP in
        (mail-fetch-field "maildir")
        header-values))))
 
-(add-hook 'bug-reference-auto-setup-functions
-          #'mu4e--view-try-setup-bug-reference-mode)
+(with-eval-after-load 'bug-reference
+  (add-hook 'bug-reference-auto-setup-functions
+            #'mu4e--view-try-setup-bug-reference-mode))
 
 
 (provide 'mu4e-view)


tsdh:
@leotaku Hi Leo, yes, I'm interested.  You are absolutely right and the patch looks good.  The issue never occurred to me because I require `bug-reference` in my `user-init-file`.

djcb:
Looks useful.. .thanks, merged.

thierryvolpiatto:
Avoiding the use of `with-eval-after-load` in mu4e source files would be better IMHO.
As this hook run after regular `bug-reference-mode-hook` probably something like below would work and be cleaner (not tested though):

```
diff --git a/mu4e/mu4e-view.el b/mu4e/mu4e-view.el
index 776f71fa..d08a2e27 100644
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -1328,10 +1328,12 @@ GROUP-REGEXP and each header value against HEADER-REGEXP in
        (mail-fetch-field "maildir")
        header-values))))
 
-(with-eval-after-load 'bug-reference
+(defun mu4e--view-add-bug-reference-hook ()
   (add-hook 'bug-reference-auto-setup-functions
             #'mu4e--view-try-setup-bug-reference-mode))
 
+(add-hook 'bug-reference-mode-hook 'mu4e--view-add-bug-reference-hook)
+
 
 (provide 'mu4e-view)
 ;;; mu4e-view.el ends here
```

tsdh:
Thierry Volpiatto ***@***.***> writes:

Hi Thierry,

> Avoiding the use of `with-eval-after-load` in mu4e source files would
> be better IMHO.

Why?  Because of this sentence in (info "(elisp) Hooks for Loading")?

     Normally, well-designed Lisp programs should not use
  ‘with-eval-after-load’.  If you need to examine and set the variables
  defined in another library (those meant for outside use), you can do
  it immediately—there is no need to wait until the library is loaded.

Well, the above only applies to variables whose default value is nil,
unfortunately.

Hm, I could change bug-reference.el so that
bug-reference-auto-setup-functions is defined like this:

--8<---------------cut here---------------start------------->8---
(defvar bug-reference-auto-setup-functions nil
  "Functions trying to auto-setup `bug-reference-mode'.
  ...)

(dolist (fn (list #'bug-reference-try-setup-from-vc
                  #'bug-reference-try-setup-from-gnus
                  #'bug-reference-try-setup-from-rmail
                  #'bug-reference-try-setup-from-rcirc
                  #'bug-reference-try-setup-from-erc))
  (add-hook 'bug-reference-auto-setup-functions fn))
--8<---------------cut here---------------end--------------->8---

Then it wouldn't matter to add-hook to it before bug-reference is
loaded...

> As this hook run after regular `bug-reference-mode-hook` probably
> something like below would work and be cleaner (not tested though):
>
> ```
> diff --git a/mu4e/mu4e-view.el b/mu4e/mu4e-view.el
> index 776f71fa..d08a2e27 100644
> --- a/mu4e/mu4e-view.el
> +++ b/mu4e/mu4e-view.el
> @@ -1328,10 +1328,12 @@ GROUP-REGEXP and each header value against HEADER-REGEXP in
>         (mail-fetch-field "maildir")
>         header-values))))
>  
> -(with-eval-after-load 'bug-reference
> +(defun mu4e--view-add-bug-reference-hook ()
>    (add-hook 'bug-reference-auto-setup-functions
>              #'mu4e--view-try-setup-bug-reference-mode))
>  
> +(add-hook 'bug-reference-mode-hook 'mu4e--view-add-bug-reference-hook)
> +
>  
>  (provide 'mu4e-view)
>  ;;; mu4e-view.el ends here
> ```

Indeed, that should also do the trick though it will try to add the
setup function whenever bug-reference-mode is enabled in a buffer, i.e.,
it adds per buffer and not only once which would suffice given that
bug-reference-auto-setup-functions is a global variable, at least by
default.  So right now, I prefer the with-eval-after-load version.

Bye,
Tassilo

tsdh:
Tassilo Horn ***@***.***> writes:

> Hm, I could change bug-reference.el so that
> bug-reference-auto-setup-functions is defined like this:
>
> (defvar bug-reference-auto-setup-functions nil
>   "Functions trying to auto-setup `bug-reference-mode'.
>   ...)
>
> (dolist (fn (list #'bug-reference-try-setup-from-vc
>                   #'bug-reference-try-setup-from-gnus
>                   #'bug-reference-try-setup-from-rmail
>                   #'bug-reference-try-setup-from-rcirc
>                   #'bug-reference-try-setup-from-erc))
>   (add-hook 'bug-reference-auto-setup-functions fn))
>
> Then it wouldn't matter to add-hook to it before bug-reference is
> loaded...

I've done that now on the emacs-29 branch but unfortunately that won't
help us because immediately until mu4e drops emacs-28 support.

Bye,
Tassilo

thierryvolpiatto:
Hello Tassilo,

Tassilo Horn ***@***.***> writes:

> 1.  ( ) multipart/mixed     (*) text/html           
>
> Thierry Volpiatto ***@***.***> writes:
>
> Hi Thierry,
>
>> Avoiding the use of `with-eval-after-load` in mu4e source files would
>> be better IMHO.
>
> Why? Because of this sentence in (info "(elisp) Hooks for Loading")?

Yes, mostly.

> Normally, well-designed Lisp programs should not use
> ‘with-eval-after-load’. If you need to examine and set the variables
> defined in another library (those meant for outside use), you can do
> it immediately—there is no need to wait until the library is loaded.
>
> Well, the above only applies to variables whose default value is nil,
> unfortunately.
>
> Hm, I could change bug-reference.el so that
> bug-reference-auto-setup-functions is defined like this:
>
> --8<---------------cut here---------------start------------->8---
> (defvar bug-reference-auto-setup-functions nil
> "Functions trying to auto-setup `bug-reference-mode'.
> ...)
>
> (dolist (fn (list #'bug-reference-try-setup-from-vc
> #'bug-reference-try-setup-from-gnus
> #'bug-reference-try-setup-from-rmail
> #'bug-reference-try-setup-from-rcirc
> #'bug-reference-try-setup-from-erc))
> (add-hook 'bug-reference-auto-setup-functions fn))
> --8<---------------cut here---------------end--------------->8---
>
> Then it wouldn't matter to add-hook to it before bug-reference is
> loaded...

Yes, nice. 

>> As this hook run after regular `bug-reference-mode-hook` probably
>> something like below would work and be cleaner (not tested though):
>>
>> ```
>> diff --git a/mu4e/mu4e-view.el b/mu4e/mu4e-view.el
>> index 776f71fa..d08a2e27 100644
>> --- a/mu4e/mu4e-view.el
>> +++ b/mu4e/mu4e-view.el
>> @@ -1328,10 +1328,12 @@ GROUP-REGEXP and each header value against HEADER-REGEXP in
>> (mail-fetch-field "maildir")
>> header-values))))
>>
>> -(with-eval-after-load 'bug-reference
>> +(defun mu4e--view-add-bug-reference-hook ()
>> (add-hook 'bug-reference-auto-setup-functions
>> #'mu4e--view-try-setup-bug-reference-mode))
>>
>> +(add-hook 'bug-reference-mode-hook 'mu4e--view-add-bug-reference-hook)
>> +
>>
>> (provide 'mu4e-view)
>> ;;; mu4e-view.el ends here
>> ```
>
> Indeed, that should also do the trick though it will try to add the
> setup function whenever bug-reference-mode is enabled in a buffer, i.e.,
> it adds per buffer and not only once which would suffice given that
> bug-reference-auto-setup-functions is a global variable, at least by
> default. So right now, I prefer the with-eval-after-load version.

Also what is possible is requiring bug-reference on top of
mu4e-view (bug-reference is small and load fast) and then adding the hook normally.

> Bye,
> Tassilo
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you commented.*Message ID: ***@***.***>


-- 
Thierry

thierryvolpiatto:
Tassilo Horn ***@***.***> writes:

> 1.  ( ) text/plain          (*) text/html           
>
> Tassilo Horn ***@***.***> writes:
>
>> Hm, I could change bug-reference.el so that
>> bug-reference-auto-setup-functions is defined like this:
>>
>> (defvar bug-reference-auto-setup-functions nil
>> "Functions trying to auto-setup `bug-reference-mode'.
>> ...)
>>
>> (dolist (fn (list #'bug-reference-try-setup-from-vc
>> #'bug-reference-try-setup-from-gnus
>> #'bug-reference-try-setup-from-rmail
>> #'bug-reference-try-setup-from-rcirc
>> #'bug-reference-try-setup-from-erc))
>> (add-hook 'bug-reference-auto-setup-functions fn))
>>
>> Then it wouldn't matter to add-hook to it before bug-reference is
>> loaded...
>
> I've done that now on the emacs-29 branch but unfortunately that won't
> help us because immediately until mu4e drops emacs-28 support.

Great, thanks.

> Bye,
> Tassilo
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you commented.*Message ID: ***@***.***>


-- 
Thierry

leotaku:
> Tassilo Horn ***@***.***> writes:
> Hm, I could change bug-reference.el so that bug-reference-auto-setup-functions is defined like this: (defvar bug-reference-auto-setup-functions nil "Functions trying to auto-setup `bug-reference-mode'. ...) (dolist (fn (list #'bug-reference-try-setup-from-vc #'bug-reference-try-setup-from-gnus #'bug-reference-try-setup-from-rmail #'bug-reference-try-setup-from-rcirc #'bug-reference-try-setup-from-erc)) (add-hook 'bug-reference-auto-setup-functions fn)) Then it wouldn't matter to add-hook to it before bug-reference is loaded...
> I've done that now on the emacs-29 branch but unfortunately that won't help us because immediately until mu4e drops emacs-28 support. Bye, Tassilo

@tsdh I personally am unsure if this is the best way of solving the issue. I fear that with this way of adding the default hooks, users might be confused when adding their own custom hook before loading `bug-reference`, because, as I understand it, the default hooks will be inserted in front of them when `bug-reference` is loaded.

My personal preferred way of solving this issue would be having a `defconst` variable called something like `bug-reference-default-auto-setup-functions` that would be tried only after any functions in `bug-reference-auto-setup-functions`, which would be `nil` per default. I've seen this pattern in other packages like e.g. `multiple-cursors`.

An alternative solution would be passing a high `DEPTH` parameter to the `add-hook` calls for the default functions, but in my opinion this could also be confusing for users.

tsdh:
@leotaku The `foo-default-functions` + `foo-functions` approach is usually taken for stuff where the former is a `defvar` or `defconst` and the latter a `defcustom`.  In our case, there is no `defcustom` involved and `bug-reference-auto-setup-functions` is basically only intended for packages.  And the order of functions shouldn't make a difference since each function should only trigger if it is responsible for this buffer, i.e., by checking `derived-mode-p` or something alike.  I think, I'll add an explaining paragraph to its documentation string.

If a user wants to implement his own setup, he should do so in `bug-reference-mode-hook` or by setting the variables using `.dir-locals.el` or a local variables section in their files.  If `bug-reference-url-format` and `bug-reference-bug-regexp` are set after running this hook, the auto-setup is inhibited anyhow.

PR #2440, opened 2023-03-07
mu4e: Add optional message plist arg to mu4e--org-store-link-message

johnhamelink:
This change makes it possible to use the function in an mu4e action, like so:

```
(defun jjh/mu4e-action-store-and-capture (msg)
  "Store the link to MSG, then run `org-capture'."
  (mu4e--org-store-link-message msg)
  (org-capture))
```


djcb:
Merged; thanks.

PR #2442, opened 2023-03-15
Don't die if the headers buffer isn't in a window

vermiculus:
If the headers buffer isn't in a live window (e.g., mu4e-split-view is 'single-window), get-buffer-window will return nil. In turn, with-selected-window will die if fed a nil window, so default to using the current window in this case.

This fixes a regression from c23dad70586bbb54891c506629f2ce2ed8e463d2 that prevented using n/p from a view buffer.


djcb:
Merged; thanks!

PR #2443, opened 2023-03-15
Change letter mark for mailing-list messages to "L"

farrokhi:
Fixed an issue where messages from mailing-list ("list" flag) were shown using "s" in header view, which is also used for "signed" messages. Therefore changing it to "L".


djcb:
Yeah, the 's' seems like a copy-paste error. However the character should be a lower-case 'l', for consistency with the other content-flags (and `mu fields`)... can you do that? Thanks!

djcb:
Oh, I've done already & merged :)

farrokhi:
I saw this a bit late. Thanks for merging it!

PR #2449, opened 2023-03-28
[ux bug] mu4e: create random name for newly detached views

seanfarley:
Without creating some kind of a temporary name (which could then be
trivially renamed by the user as the current instructions say) a new
user trying to learn how to detach a view will run into this error when
switching to a new message after detaching:

[mu4e] Detached view buffer from *mu4e-headers*
funcall-interactively: No buffer named *mu4e-article*<2>

It seems like a good default and less friction to create a random name
automatically when detaching.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -327,11 +327,14 @@ Add this function to `mu4e-view-mode-hook' to enable this feature."
   (unless mu4e-linked-headers-buffer
     (mu4e-error "This view buffer is already detached."))
   (mu4e-message "Detached view buffer from %s"
-                (prog1 mu4e-linked-headers-buffer
+                (progn mu4e-linked-headers-buffer
                   (with-current-buffer mu4e-linked-headers-buffer
                     (when (eq (selected-window) mu4e~headers-view-win)
                       (setq mu4e~headers-view-win nil)))
-                  (setq mu4e-linked-headers-buffer nil))))
+                  (setq mu4e-linked-headers-buffer nil)
+                  ;; automatically rename mu4e-view-article buffer when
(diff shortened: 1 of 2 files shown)


seanfarley:
I might be totally off here but I couldn't think of a reason why a user would want the old `*mu4e-article*` name. Just to note, reattaching a view still works with this change suggestion.

djcb:
@mickeynp: thoughts? Thanks!

mickeynp:
Sounds good -- but I recommend keeping the earmuffs for the buffer names to keep it consistent with other parts of mu4e.

seanfarley:
Took me a minute to figure out what "earmuffs" meant but I see now. Makes sense!

djcb:
Can we use `uniquify` for this?

mickeynp:
Ideally that'd be best. However I recommend we have use prefix naming (in keeping with mu4e tradition) to make it easy to override in `display-buffer-alist` without having to resort to custom functions.

mickeynp:
I recommend you follow the Emacs/uniquify style guide and append the name instead of trying to force it in between the stars: `*mu4e-article*<whatever>`.

seanfarley:
> I recommend you follow the Emacs/uniquify style guide and append the name instead of trying to force it in between the stars: `*mu4e-article*<whatever>`.

Oh really? I can easily do that but I thought people (such as myself) would have `^*mu4e-.*$` type regexs? Admittedly, I only know enough about `display-buffer-alist` to hurt myself so I'm learning a lot here.

seanfarley:
I think (hope?) the new code is ready for a new review now 😬

djcb:
Merged, thanks.

PR #2457, opened 2023-04-01
Fix defcustom type definitions

RuijieYu:
So that the type definitions reflects the docstrings more closely.  In additon, fixed warnings from flymake about docstring.

* mu4e/mu4e-draft.el mu4e-compose-keep-self-cc: should be a comma

* mu4e/mu4e-folders.el mu4e-maildir-shortcuts: recognize new (plist) format

* mu4e/mu4e-helpers.el mu4e-debug: removed unnecessary double space in docstring

* mu4e/mu4e-query-items.el mu4e-query-rewrite-function: prepend "\\=" before column 0 "(" in docstring

* mu4e/mu4e-search.el mu4e-search-sort-field mu4e-search-sort-direction: used radio in types

* mu4e/mu4e-server.el mu4e-mu-binary: "mu" executable must exist

* mu4e/mu4e-view.el mu4e-view-fields: fixed type
mu4e-view-actions: for function refs, prefer "#'" rather than "'"

### DIFF
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -60,7 +60,7 @@ Useful when this is not equal to the From: address."
   :group 'mu4e-compose)
 
 (defcustom mu4e-compose-keep-self-cc nil
-  "When non-nil. keep your e-mail address in Cc: when replying."
+  "When non-nil, keep your e-mail address in Cc: when replying."
   :type 'boolean
   :group 'mu4e-compose)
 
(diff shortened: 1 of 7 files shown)


RuijieYu:
This PR is inspired by a bunch of warnings showing up when I do a `(setopt mu4e-maildir-shortcuts '((:maildir "Root" :key ??) ...))` which uses the new format.  On closer inspection it seemed to me that the type definition for this variable had not been updated, so I decided to grep for defcustom and look at each of the type definitions.

djcb:
Ah, nice work.

PR #2464, opened 2023-04-03
don't use display-buffer-reuse-window when composing in frame

### LINKED ISSUE #2458: `display-buffer-reuse-window` is set unconditionally
**Describe the bug**

When creating a new, second e-mail draft and `mu4e-compose-in-new-frame` is set then a new frame should be created instead of the existing reused.

**How to Reproduce**

Create two e-mail drafts with `mu4e-compose-in-new-frame`  set.

The root cause is that `display-buffer-reuse-window` and `display-buffer-reuse-mode-window` is set unconditionally, but it should not be used when creating new frames for composition.

**Environment**
Please describe the versions of OS, Emacs, mu/mu4e etc. you are using.

Debian, Emacs 29, mu/mu4e 1.10

**Checklist**
- [X] I am running 1.10
- [X] you can reproduce the problem without 3rd party extensions (including Doom/Evil etc.)

malb:
fixes #2458

### DIFF
--- a/mu4e/mu4e-window.el
+++ b/mu4e/mu4e-window.el
@@ -297,22 +297,28 @@ for BUFFER-OR-NAME to be displayed in."
             ('(view . vertical) '((window-min-width . fit-window-to-buffer)))
             (`(,_ . t) nil)))
          (window-action (cond
-                         ((eq buffer-type 'main) '(display-buffer-full-frame))
+                         ((eq buffer-type 'main) '(display-buffer-reuse-window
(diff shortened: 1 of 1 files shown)


djcb:
Thanks, looks good to me... any thoughts, @mickeynp?

mickeynp:
Nice one, malb. Looks good to me if it looks good to you, Dirk-Jan.

djcb:
Merged; thanks!

PR #2468, opened 2023-04-05
mu4e: allow to disable the global mode-line items

### LINKED ISSUE #2463: [mu4e rfe] Ability to disable the global mode-line tweaks
**Is your feature request related to a problem? Please describe.**

Whereas I enjoy having some segments in the mode-line of the headers buffer I'd rather mu4e to not to maintain any "global mode-line". This is because I have my own new mail indicators already. I checked in the manual (sections 13.2.x) and it does not seem possible to disable the global mode-line. It'd be nice if this was configurable.

**Describe the solution you'd like**

A defcustom in `mu4e-modeline.el` to prevent mu4e from touching the global mode-line.

nbarrientos:
This patch permits configuring mu4e so only buffer-specific modelines are displayed.

Contrary to the solution I initially suggested in <https://github.com/djcb/mu/issues/2463>, not registering the modeline is not enough. This is because somebody could have `mu4e-modeline-show-global` set to `t`, start `mu4e` (the global modeline is registered), then set it to `nil` and re-start `mu4e`. As there's nothing cleaning up `mu4e--modeline-global-items`, changing `mu4e-modeline-show-global` during an Emacs session that followed the path described wouldn't work.

The new way of fixing this is more resilient, allowing the global modeline to be registered but only adding the global bits to the return value of `mu4e--modeline-string` when configured to do so. This approach allows as well changing `mu4e-modeline-show-global` on the fly.

It has been tested with mu4e 1.10.1 + this patch.

Fixes #2463

### DIFF
--- a/mu4e/mu4e-modeline.el
+++ b/mu4e/mu4e-modeline.el
@@ -48,6 +48,14 @@ display the bookmark name rather than the query."
   :type 'boolean
   :group 'mu4e-modeline)
 
+(defcustom mu4e-modeline-show-global t
+  "Whether to populate global modeline segments.
+
+If non-nil, show both buffer-specific and global modeline items,
+otherwise only present buffer-specific information."
+  :type 'boolean
+  :group 'mu4e-modeline)
+
 (defvar-local mu4e--modeline-buffer-items nil
   "List of buffer-local items for the mu4e modeline.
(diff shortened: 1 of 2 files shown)


djcb:
Merged -- thanks!

PR #2484, opened 2023-05-04
Allow on-the-fly updates to `mu4e-user-mailing-lists' (Bug#2483)

### LINKED ISSUE #2483: [mu4e bug] Customizing `mu4e-user-mailing-lists` has no effect with existing (and even quitted) mu4e sessions
**Describe the bug**
Please provide a clear and concise description of what you expected to happen
and what actually happened.

I tried to modify `mu4e-user-mailing-lists` but the modification didn't take effect even after restarting `mu4e`. 

I have an idea of how to fix it, so I'll send a PR within a few days.

Steps:

1. Open a mu4e session. 
2. Run `(setopt mu4e-user-mailing-lists '(("bug-gnu-emacs.gnu.org" . "EmacsBug")))`.  Equivalent actions within customize view also suffices. 
3. Notice that nothing changes in the mail listings.
4. Restart mu4e and try again, notice (3) again. 
5. Try restarting Emacs. 
6. (Didn't try, but after (5) everything should work as expected.)

RuijieYu:
DRAFT.  Fixes #2483.

Note that there is a TODO item: I made a rename from `mu4e--lists-hash` to `mu4e-lists--hash`, but noticed that it is also used in `mu4e-search.el`, so I wonder if I should rename it there as well.

Comments welcome.

### DIFF
--- a/mu4e/mu4e-lists.el
+++ b/mu4e/mu4e-lists.el
@@ -28,7 +28,7 @@
 ;;; Code:
 
 ;;; Configuration
-(defvar mu4e-mailing-lists
+(defconst mu4e-mailing-lists
   '( ("bbdb-info.lists.sourceforge.net"                       . "BBDB")
      ("boost-announce.lists.boost.org"                        . "BoostA")
      ("boost-interest.lists.boost.org"                        . "BoostI")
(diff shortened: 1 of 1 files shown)


djcb:
Thanks! I change it in a slightly different way, there's `mu4e-mailing-list-info-refresh` now.

PR #2488, opened 2023-05-10
Updates to mu4e-contacts.el

RuijieYu:
* mu4e/mu4e-contacts.el
 
(mu4e-personal-address-p): Fix docstring error (return nil *only if* the address does not match any of the personal addresses, not the other way around).  Unwrap unnecessary (if COND t nil) to just COND because its caller `seq-find` only distinguishes between nil and non-nil.

(mu4e-personal-address): Add space in defun between name and args. Ensure double space after a sentence ends.  Convert a duplicated call of (mu4e-server-properties) into using when-let and reusing its result.


djcb:
Merged; thanks.

PR #2489, opened 2023-05-10
* mu4e/mu4e-contacts.el (mu4e-compose-complete-max): fix type

RuijieYu:
Should `mu4e-compose-complete-max` take a string or integer?  I can't deduce it from the docs.


djcb:
> Should `mu4e-compose-complete-max` take a string or integer? I can't deduce it from the docs.

It should be a number (or `nil`)

djcb:
Merged; thanks.

PR #2491, opened 2023-05-10
Ensure mu4e-thread-mode is only enabled in headers buffer

andersjohansson:
This was run outside of the `with-current-buffer (mu4e-get-headers-buffer)` in `mu4e~headers-found-handler`, which could be anywhere.

I noticed it when `mu4e-thread-mode` was suddenly active in other buffers and shadowed my `C-tab` binding.


djcb:
Good catch, thanks!

PR #2492, opened 2023-05-12
Fix more defcustom type errors and reworded docs

RuijieYu:
* mu4e/mu4e-draft.el (mu4e-compose-reply-recipients): (mu4e-compose-reply-to-address):
* mu4e/mu4e-thread.el (mu4e-thread-fold-single-children): Clarify docstrings and fixed type definitions.

(Wrote a command that checks types for all options in a group, so all types should be good now.)


RuijieYu:
Also, shouldn't `mu4e-thread-fold-single-children` be `...-child` instead, singular form?

djcb:
Merged, thanks.

PR #2494, opened 2023-05-26
Fix asking maildir reading completing candidate is text-property

stardiviner:
Some packages like `vertico-prescient` adding text-property to vertico candidates will break `mu4e-ask-maildir` response.

Here is the screenshot when text-property candidates caused error:

Screenshot 2023-05-26 at 09 33 38@2x


djcb:
Merged, thanks

PR #2495, opened 2023-05-28
Add a custom variable to decide if we can edit any message or just the draft messages.

i02sopop:
In some situations you might want to edit any mail, i.e., when there's a patch in the mail and you want to edit that patch before applying it.

This PR allows the user to indicate if (s)he wants to be able to edit just the draft messages (which is the default behavior) or if it's allowed to edit any message by adding a new custom variable to the configuration.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -81,6 +81,15 @@
 ;;; Configuration
 ;; see mu4e-drafts.el
 
+(defcustom mu4e-allow-edit-any-message nil
+  "Whether to allow to edit any message or just the draft messages.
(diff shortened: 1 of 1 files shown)


djcb:
Thanks for the effort, but this is not really a capability I want to add to `mu4e`; you can edit draft messages but no other ones (you can of course just open the file and change it directly -- but that's outside `mu4e`).

i02sopop:
> Thanks for the effort, but this is not really a capability I want to add to `mu4e`; you can edit draft messages but no other ones (you can of course just open the file and change it directly -- but that's outside `mu4e`).

Ok, I'll close the PR and I'll keep it in my local copy, since it's something I want to have in my email workflow, thanks for the input.

PR #2496, opened 2023-05-28
Add a new maildir section in the homepage.

i02sopop:
Since not all the maildirs have a shortcut defined, and the user might want to see more than the maildirs with a shortcut in their home, we add a new maildir section in the homepage to see all the defined maildirs to show. The list of maildirs to be shown are defined in the new custm mu4e-maildirs variable.

### DIFF
--- a/mu4e/mu4e-bookmarks.el
+++ b/mu4e/mu4e-bookmarks.el
@@ -44,7 +44,6 @@
       :key ?t)
     ( :name "Last 7 days"
       :query "date:7d..now"
-      :hide-unread t
       :key ?w)
     ( :name "Messages with images"
       :query "mime:image/*"
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -91,7 +91,14 @@ This multiplexes the `message-mode' hooks `message-send-actions',
   :type 'hook
   :group 'mu4e-compose)
 
+(defcustom mu4e-allow-edit-any-message nil
+  "Whether to allow to edit any message or just the draft messages.
+Setting this to t allows any message to be edited, not just the draft messages.
 
+Changes to this value only take effect after (re)starting the mu
+session."
+  :type 'boolean
+  :group 'mu4e)
 
 (defvar mu4e-captured-message)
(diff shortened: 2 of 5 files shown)


djcb:
Thank you for the effort/PR! 

However, I don't want an extra maildir-section in the main view. TBH, I don't see much value in having maildir shortcuts _without_ a shortcut; even if you don't use them, you can still add some shortcut. You could even give all of the same binding, e.g., `:key ?_`. Or?

_Maybe_ we could support key-less shortcuts directly in the normal Maildirs section, but not convinced it's worth the extra code.

i02sopop:
> Thank you for the effort/PR!
> 
> However, I don't want an extra maildir-section in the main view. TBH, I don't see much value in having maildir shortcuts _without_ a shortcut; even if you don't use them, you can still add some shortcut. You could even give all of the same binding, e.g., `:key ?_`. Or?
> 
> _Maybe_ we could support key-less shortcuts directly in the normal Maildirs section, but not convinced it's worth the extra code.

What I was looking for is to have one section in the home view with all the maildirs I have, even if they don't have shortcuts, because I want to know which maildirs have new emails even if I don't have defined a shortcut. I have over 100 maildirs that I don't check daily, and it seems kind of hacky to assign the same key binding to all of them, apart that it's clearer for me to have a first section of my daily maildirs and a second section with the rest of maildirs.

If you don't want to add this new section to the home view is ok, I can maintain it in my local changes without problem :).

llmII:
Unless things have changed, assigning the same shortcut to every maildir results in the jumplist shown in the minibuffer being very ugly. Something like:
mu4e-with-a-ridiculous-jump-list

I'm not too picky with regards to if it should be a separate section or inline with another section, but I think this would probably be yet another time someone has wanted this and @djcb doesn't. I'm among the number wanting something like this for similar purposes, displaying a count of read/unread and path basically.

Keep in mind the image below is from my own "patched" version and it's likely my hack is uglier than this PR, but something like the below where you have hundreds of maildirs makes it easier to see which boxes have new/unread:
mu4e-with-hacks

Issue #2466, opened 2023-04-04
[mu4e bug] address autocompletion


pmenair:
autocompletion creates address that Microsoft will not accept

where there is no name, instead of ```name <address@domain.com>``` or just ```address@domain.com``` I get ```address@domain.com <address@domain.com>```

MSFT refuses to send this

**Environment**
Up to date Emacs and MacOS

**Checklist**
- [X] you are running either an 1.8.x/1.10.x release or `master` (otherwise please upgrade)
- [X] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)
- [X] you have read all of the above

djcb:
Yeah, that's an invalid address, so that's my it's rejected.  

Anyway, does this happen with _all_ such addresses, or one in particular? Please attach an email with the address where this happens. You can remove any information and/or change the email as long as the problem is still reproducible. 

Also, what mu version?

pmenair:
version is 1.8.14

Not sure I can give you anything that will reproduce it. The problem is in wherever autocomplete is pulling addresses from, so it must be something arising from the email the address was originally located in, not the one I'm trying to insert it in, right?

It only happens with some addresses, it's rare enough that I forget to look for it. But it's not just one address.

djcb:
Indeed; the addresses are harvested from the messages you received. You can get their "display names" (use for completion) through `M-x mu4e-contacts-info` (which is meant for such debugging). Once you've located a problem case, please find the message where it originated from, and attach it here, thanks.

It must be something special / rare, or it would affect more people.

pmenair:
problem is I'm not comfortable attaching email, it's from a law firm

When I run M-x mu4e-contacts-info it shows ```name@domain.com <name@domain.com>```. When I look at email with that sender/recipient, it shows BOTH ```"name@domain.com" <name@domain.com>``` and ```"fname lname" <name@domain.com>```, but there is only one entry in the contacts info output. This is programmatically generated on the fly, so there's no "address book" to edit, right?

djcb:
I'd never ask to attach anything contain private information, but as mentioned,  you can remove any information and/or change the email as long as the problem is still reproducible. 

Anyway, this may be moot since apparently the correct data is mu4e's contact info.

So you have 
   `Foo Bar <foo@example.com>`
in `mu4e-contacts-info`, yet when you try to auto-complete that, the final `>` is not included? And this happens _always_ what this particular address and not with others?

pmenair:
[...]

PR #2501, opened 2023-06-09
Quote Display Names which contain special characters

jaknel:
RFC 822 and RFC 5322 define characters which require an atom be quoted when included as a word in a mailbox.

### DIFF
--- a/lib/message/mu-contact.cc
+++ b/lib/message/mu-contact.cc
@@ -29,13 +29,20 @@
 
 using namespace Mu;
 
+const std::string rfc822SpecialCharacters  = "()<>@,;:\\\".[]";
+
 std::string
 Contact::display_name() const
 {
 	auto needs_quoting= [](const std::string& n) {
-		for (auto& c: n)
-			if (c == ',' || c == '"')
+		for (auto& c: n){
+			// RFC 822 3.3 and RFC 5322 3.2.3 defines special characters, Ascii
+			// Control (CTL) characters (0-31), DEL (127), and space as requiring
+			// quoting.
+			if (c <=31 || c >= 127 || c == ' ' || rfc822SpecialCharacters.find(c) != std::string::npos){
 				return true;
+			}
+		}
 		return false;
 	};
 
--- a/lib/message/mu-contact.cc
+++ b/lib/message/mu-contact.cc
@@ -29,17 +29,16 @@
 
 using namespace Mu;
 
-const std::string rfc822SpecialCharacters  = "()<>@,;:\\\".[]";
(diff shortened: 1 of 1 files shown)


jaknel:
**Describe the bug**
An email address which ends up in the contact list database as `test@example.org <test@example.org>` needs to be quoted as `"test@example.org" <test@example.org>` when authoring an email in Mu4e

**To Reproduce**
```
From: test@example.org <test@example.org>
Date: 2023-06-03

Test
```
Index the above email using mu index on a Maildir. Then autocomplete the address while authoring an email in Mu4e.  If sent to an Office365 account the email will bounces as undeliverable (The date header above is included only to deal with variable `mu4e-compose-complete-only-after`)

**Environment**
Emacs and mu/mu4e all built from master on Ubuntu linux.

**Checklist**
- [ x] you are running either the latest 1.8.x/1.10.x release or `master` (otherwise, please upgrade).

jaknel:
I believe this also resolves  #2466

djcb:
Hmmm, the PR seems to do too much (and breaks the unit-tests); in practice, seems the RFCs are not followed and spaces, non-ascii UTF-8 are used without any problem. The Office365 may be a special case here; can't we fix that case by just adding '@' to the characters `needs_quoting` must test for?

jaknel:
Patch 96e16034c70d549eac243b00579fe1e0ed1a22a5 added a check for ','  My first attempt added  '@' on top of that `if (c == ',' || c == '"' || c == '@')` and would resolve my specific issue. I'd agree the issue would be resolved with just that change.

This PR came from looking at the RFCs and trying to figure out what the Right Thing™ to do is.   I thought a more general fix could benefit others.  For example, an email with the mailbox: `first<last  <first.last@example.org>`  Mu will index the address, return the unquoted candidate to mu4e using auto complete, but then message-mode (gnus?) will complain `Email address first<last <first.last@example.org> looks invalid; send anyway? (y or n)` when sending the email.  However, if auto complete returned `"first<last"  <first.last@example.org>`  the email is sent without an issue.   

I've pushed a new commit which points to RFC5322 and that does not mark spaces, periods, or extended ascii as need quoting.  `make test` is also now successful.  But again, I would be ok with just '@' making it into the codebase.

djcb:
This is the display-name, and gnus will still process it (e.g.,utf-8 encoding); so `display_name()` is not necessarily RFC-compliant. I've just pushed a change that adds the '@' , as suggested -- thanks!

(I'll think about those other cases)

djcb:
No longer needed, so closing this; thanks!

PR #2521, opened 2023-07-18
mu4e-notifications: add default action

mssdvd:
This PR adds a default notification action.

This is what the freedesktop.org specification says:

>  The default action (usually invoked by clicking the notification) should have a key named "default". The name can be anything, though implementations are free not to display it.


djcb:
Merged, thanks!

PR #2522, opened 2023-07-19
Install Guile extension into Guile's extension directory

a3a3el:


a3a3el:
Forward porting this fix: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1041271

djcb:
Merged, thanks!

Issue #2524, opened 2023-07-23
[mu4e bug] Incorrect TO and CC lists when reply-all to a message I sent


danielfleischer:
**Describe the bug**
Given an email I previously sent with **TO** and **CC** lists, I want to reply all in order to "continue" the thread, choosing replying to all recipients. 

What I expect: new email draft with the **same** TO: and CC: lists. 

What I get: **all** recipients are now in the TO: list. 

**Environment**
Mac, Emacs 29.0.2. Mu Version 1.11.8. `mu4e-compose-dont-reply-to-self` is set to t. 

**Comments**
TO and CC lists are reconstructed correctly when I am not the original sender so it has something to do with the fact that the sender is self.

djcb:
Merged: https://github.com/djcb/mu/pull/2530
Thanks.

PR #2530, opened 2023-08-06
To/Cc Lists fix

danielfleischer:
Related to #2524. 

Modified code to achieve consistent "reply-all" behavior.

**Case 1: I'm not the sender**

> From: sender
> To: Me, group 1
> CC: group 2

results in 

> From: Me
> To: sender, group 1
> CC: group 2

**Case 2: I'm the sender**

> From: me
> To: group 1
> CC: group 2

results in:

> From: me
> To: group 1
> CC: group 2


djcb:
Merged -- thanks!

Issue #2535, opened 2023-08-16
[mu4e bug] Reply to sender (only) does not work


Chris00:
**Describe the bug & How to Reproduce**

Reply to any message with “reply to sender only” and all addresses end up in `To`.

In the function
```elisp
(defun mu4e~draft-create-to-lst (origmsg)
  "Create a list of address for the To: in a new message.
This is based on the original message ORIGMSG. If the Reply-To
address is set, use that, otherwise use the From address. Note,
whatever was in the To: field before, goes to the Cc:-list (if
we're doing a reply-to-all). Special case: if we were the sender
of the original, we simple copy the list form the original."
  (let* ((reply-to
         (append
          (plist-get origmsg :to)
          (or (plist-get origmsg :reply-to) (plist-get origmsg :from))))
         (reply-to (cl-delete-duplicates reply-to
                                         :test #'mu4e~draft-address-cell-equal)))
    (if mu4e-compose-dont-reply-to-self
        (cl-delete-if
         (lambda (to-cell)
           (mu4e-personal-address-p (mu4e-contact-email to-cell)))
         reply-to)
      reply-to)))
```
I do not understand the `(plist-get origmsg :to)` because it will put all addresses in `To` regardless of whether one wants to replay to all or not.

**Environment**
Debian,
mu commit 6168d776

**Checklist**
- [x] you are running either an 1.8.x/1.10.x release or `master` (otherwise please upgrade)
- [x] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)
- [x] you have read all of the above

PR #2536, opened 2023-08-16
Don't use the original message `To` for a reply `To` field

Chris00:
Fixes https://github.com/djcb/mu/issues/2535


djcb:
Merged -- thanks!

PR #2542, opened 2023-08-23
mu4e: make it conform with `package.el`

jian-lin:
# Problem

`mu4e` does not conform with `pakcage.el`, which makes autoload commands, such as `(mu4e)`, not work.  Users have to `(require 'mu4e)` first.

# Cause

`(package-activate-all)` of `package.el` is responsible for loading `NAME-autoloads.el` of all packages when Emacs starts.  See `(elisp) Startup Summary`.

- The default installation dir does not conform with `package.el`.  `package.el` looks for packages in `package-user-dir` and `package-directory-list`.  Check the source of `package-load-all-descriptors`.
- Missing `mu4e-pkg.el` description file.  See https://www.gnu.org/software/emacs/manual/html_node/elisp/Multi_002dfile-Packages.html.
- Missing `mu4e-autoloads.el`.  If a package is installed using `package.el`, this file will be generated.  Since we do not use `package.el` to install `mu4e`, we generate it manully.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -194,7 +194,8 @@ endif
 
 # emacs -- needed for mu4e compilation
 emacs_name=get_option('emacs')
-emacs=find_program([emacs_name], version: '>=26.3', required:false)
+emacs_min_version='26.3'
+emacs=find_program([emacs_name], version: '>='+emacs_min_version, required:false)
 if not emacs.found()
   message('emacs not found; not pre-compiling mu4e sources')
 endif
--- a/mu4e/meson.build
+++ b/mu4e/meson.build
@@ -26,6 +26,16 @@ mu4e_meta = configure_file(
       'MU_DOC_DIR'      : join_paths(datadir, 'doc', 'mu'),
     })
 
+mu4e_pkg_desc = configure_file(
+  input: 'mu4e-pkg.el.in',
+  output: 'mu4e-pkg.el',
+  install: true,
+  install_dir: mu4e_lispdir,
+  configuration: {
+      'VERSION'           : meson.project_version(),
+      'EMACS_MIN_VERSION' : emacs_min_version,
+    })
+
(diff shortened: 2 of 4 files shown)


jian-lin:
This PR includes https://github.com/djcb/mu/pull/2313 by @cireu.

djcb:
Not sure about this; mu4e is not a "package" (for elpa etc.): you need to build it yourself; don't see that changing any time soon. I don't want to change the default installation dir, people depend on that.

Appreciate the effort, but how would this improve mu4e?

jian-lin:
> mu4e is not a "package" (for elpa etc.): you need to build it yourself

`mu4e` does not have to be a package on GNU ELPA or MELPA and this PR is not about that.

This PR is about the compatibility between **installed** `mu4e`, either manually compiled or from a package manager (many distributions have packaged `mu4e` together with `mu` or separately), and the builtin package manager of Emacs (written in `package.el`).

The functionalities of Emacs builtin package manager can be divided into two groups:
1. downloading and installing a package
2. activating packages when Emacs starts
   - loading packages' NAME-autoloads.el
   - adding packages to `load-path`
   - adding packages to `Info-directory-list`

This PR does not cover the 1, instead it makes 2 work out of box.

> how would this improve mu4e

This PR makes defined autoload commands using [autoload cookies][1] in `mu4e` work without extra config.

> I don't want to change the default installation dir, people depend on that.

Sure, I can remove that commit https://github.com/djcb/mu/pull/2542/commits/ebaa908432da2e348b34ce9dace3a3049993e983.  People can configure `lispdir` themselves to achieve the same result.

[1]: https://github.com/djcb/mu/blob/28b779efb89d613ffe8750a595cafe09a2c426db/mu4e/mu4e.el#L58

djcb:
Thanks - yes the rest looks good. Merged.

PR #2547, opened 2023-08-31
mu4e: use special-mode for mu4e-last-update-buffer

progfolio:
This buffer is not intended to be edited.
Special mode also allows quickly dismissing via the quit-window binding.


djcb:
Merged, thanks.

PR #2552, opened 2023-09-11
mu: Fix "expected command" server error report

dme:


lib/utils/mu-command-handler.hh
```diff
@@ -171,7 +171,7 @@ private:
 		if (!listp() || empty() || !cbegin()->symbolp() ||
 		    !plistp(cbegin() + 1, cend()))
 			throw Error(Error::Code::Command,
-				    "expected command, got '%s'", to_string().c_str());
+				    "expected command, got '{}'", to_string().c_str());
 	}
```
djcb, line 175: Ah, good catch. Can you remove the `.c_str()` as well? It's not needed with fmt; thanks.

dme: 👍🏻

djcb:
Merged -- thanks!

PR #2555, opened 2023-09-17
Connect `mu4e` on windows to `mu` in WSL

samvidmistry:
This PR tries to make `mu4e` on Windows work with `mu` running on WSL. I am connecting `mu4e` to an instance of `mu` running inside WSL. Basic navigation and reading emails works. Marking and deleting emails _kinda_ works. I keep running into this error.
Screenshot 2023-09-17 140200
I tried looking around in the code to see who is calling this command so that I can add another `/` in front, but I couldn't find anything in `mu4e`. It may be being called internally from `mu`, but I'm not sure. I'm positive that what I've done may not be the best way to implement this feature and there are many other scenarios that are probably not working with this setup. Are you guys aware of this and is this feature on your roadmap? I'll be happy to help implement this feature if I receive guidance from someone who knows the codebase. 😄 I'm an Elisp newbie BTW.

### DIFF
--- a/mu4e/mu4e-headers.el
+++ b/mu4e/mu4e-headers.el
@@ -1397,24 +1397,25 @@ descendants."
 
 
 (defun mu4e-headers-view-message ()
-  "View message at point."
-  (interactive)
-  (unless (eq major-mode 'mu4e-headers-mode)
-    (mu4e-error "Must be in mu4e-headers-mode (%S)" major-mode))
-  (let* ((msg (mu4e-message-at-point))
-         (path (mu4e-message-field msg :path))
-         (_exists (or (file-readable-p  path)
-                      (mu4e-warn "No message at %s" path)))
-         (docid (or (mu4e-message-field msg :docid)
-                    (mu4e-warn "No message at point")))
-         (mark-as-read
-          (if (functionp mu4e-view-auto-mark-as-read)
-              (funcall mu4e-view-auto-mark-as-read msg)
-            mu4e-view-auto-mark-as-read)))
-    (when-let ((buf (mu4e-get-view-buffer (current-buffer) nil)))
-      (with-current-buffer buf
-        (mu4e-loading-mode 1)))
-    (mu4e--server-view docid mark-as-read)))
+       "View message at point."
+       (interactive)
+       (unless (eq major-mode 'mu4e-headers-mode)
+               (mu4e-error "Must be in mu4e-headers-mode (%S)" major-mode))
+       (let* ((msg (mu4e-message-at-point))
+              (path (mu4e-message-field msg :path))
+              (_exists ;; (or (file-readable-p  path)
+               ;;     (mu4e-warn "No message at %s" path))
+               t)
+              (docid (or (mu4e-message-field msg :docid)
+                         (mu4e-warn "No message at point")))
+              (mark-as-read
+               (if (functionp mu4e-view-auto-mark-as-read)
+                 (funcall mu4e-view-auto-mark-as-read msg)
+                 mu4e-view-auto-mark-as-read)))
+                 (when-let ((buf (mu4e-get-view-buffer (current-buffer) nil)))
+                           (with-current-buffer buf
+                             (mu4e-loading-mode 1)))
+                 (mu4e--server-view docid mark-as-read)))
 
 (defvar-local mu4e-headers-open-after-move t
   "If set to non-nil, open message after `mu4e-headers-next' and
--- a/mu4e/mu4e-helpers.el
+++ b/mu4e/mu4e-helpers.el
@@ -567,21 +567,25 @@ Or go to the top level if there is none."
 
 ;;; Misc
 (defun mu4e-join-paths (directory &rest components)
-  "Append COMPONENTS to DIRECTORY and return the resulting string.
+       "Append COMPONENTS to DIRECTORY and return the resulting string.
 
 This is mu4e's version of Emacs 28's `file-name-concat' with the
 difference it also handles slashes at the beginning of
 COMPONENTS."
-  (replace-regexp-in-string
-   "//+" "/"
-   (mapconcat (lambda (part) (if (stringp part) part ""))
-              (cons directory components) "/")))
+       (let ((double-slash (string-match "^//[^/]" directory))
+             (path (replace-regexp-in-string
+                    "//+" "/"
+                    (mapconcat (lambda (part) (if (stringp part) part ""))
+                               (cons directory components) "/"))))
+         (if double-slash
+             (concat "/" path)
+           path)))
 
 (defun mu4e-string-replace (from-string to-string in-string)
-  "Replace FROM-STRING with TO-STRING in IN-STRING each time it occurs.
+       "Replace FROM-STRING with TO-STRING in IN-STRING each time it occurs.
 Mu4e version of emacs 28's string-replace."
-  (replace-regexp-in-string (regexp-quote from-string)
-                            to-string in-string nil 'literal))
+       (replace-regexp-in-string (regexp-quote from-string)
+                                 to-string in-string nil 'literal))
 
 (defun mu4e-plistp (object)
   "Non-nil if and only if OBJECT is a valid plist.
--- a/mu4e/mu4e-message.el
+++ b/mu4e/mu4e-message.el
@@ -187,18 +187,21 @@ A part would look something like:
 (defalias 'mu4e-msg-field 'mu4e-message-field)
 
 (defun mu4e-field-at-point (field)
-  "Get FIELD for the message at point.
+       "Get FIELD for the message at point.
 Either in the headers buffer or the view buffer. Field is a
 symbol, see `mu4e-header-info'."
-  (plist-get (mu4e-message-at-point) field))
+       (plist-get (mu4e-message-at-point) field))
 
 (defun mu4e-message-readable-path (&optional msg)
-  "Get a readable path to MSG or raise an error.
+       "Get a readable path to MSG or raise an error.
 If MSG is nil, use `mu4e-message-at-point'."
-  (let ((path (plist-get (or msg (mu4e-message-at-point)) :path)))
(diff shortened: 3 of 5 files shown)


samvidmistry:
Along with the changes in PR, I have made the following tweaks

```elisp
(setq mu4e-get-mail-command (if (is-windows)
				  "wsl offlineimap"
				"offlineimap"))
(if (is-windows)
      (setq mu4e-mu-binary "wsl mu"
	    mu4e-mu-through-wsl t))
```

djcb:
People have been using mu/mu4e with WSL (and Cygwin I think) with various levels of success. I'd be fine with small tweaks, but not separate code-paths that I cannot test. 

So, I wouldn't want such `(if mu4e-mu-through-wsl ...)` blocks. It seems mostly about adding paths; but
can't you use the 'WSL' paths by passing them to `mu init --maildir / --muhome`? Either directly, or by using symlinks? The fewer changes to mu/mu4e, the better.

djcb:
> People have been using mu/mu4e with WSL (and Cygwin I think) with various levels of success. I'd be fine with small tweaks, but not separate code-paths that I cannot test.
> 
> So, I wouldn't want such `(if mu4e-mu-through-wsl ...)` blocks. It seems mostly about adding paths; but can't you use the 'WSL' paths by passing them to `mu init --maildir / --muhome`? Either directly, or by using symlinks? The fewer changes to mu/mu4e, the better.

Any update on this? Thanks!

samvidmistry:
Hey @djcb, sorry I don't check GitHub often. But I will from now on.

> People have been using mu/mu4e with WSL (and Cygwin I think) with various levels of success.

In my search, I couldn't find of any resources explaining how to connect mu running in Emacs on Windows to mu-server running in WSL. It's mostly when running emacs and mu **both** in WSL (or more frequently, Cygwin).

> It seems mostly about adding paths; but can't you use the 'WSL' paths by passing them to mu init --maildir / --muhome

I'm not sure what you mean here. `mu` executable is already inside WSL. Unfortunately, I am not able to setup my dev environment to be able to compile mu4e properly so can only test limited functionality. But the idea of setting the mu-home to a path inside WSL made sense. So I tried setting the `mu4e-mu-home` variable to `//wsl.localhost/Ubuntu/samvid/.mail`. This is possible because windows Emacs supports Unix style `/`. However, the function `mu4e-join-paths` removes the first `/`, making the path `/wsl.localhost/Ubuntu/samvid/.mail` which isn't accessible. You can see that I have modified that function in this PR. I'm not sure what the function does and why, so I can't modify it to keep the first `/` untouched in case of WSL paths. It'll be great if you can explain more about why it is removing the first `/` and whether it is possible to fix the function to work consistently with WSL paths.

djcb:
Ah, I see: https://stackoverflow.com/questions/20690828/what-does-double-slash-in-cd-mean-in-linux
So perhaps we can update `mu4e-join-paths` for that (and a few similar changes elsewhere). 

That's the kind of small changes we can do.  I'd like to avoid special WSL code paths.

samvidmistry:
@djcb Thanks for the link.

I have updated the PR with new changes. There are 3 major changes.

- Removing validations, because the paths are not windows.
- Update `mu4e-join-paths` to preserve one or two slashes, shrink any more sequences of slashes.
- One fork for WSL
```elisp
(if mu4e-mu-through-wsl
                (concat "//wsl.localhost/Ubuntu" path)
              path)))
```
This is because the path starts at `/home...`.

With this, I can properly read emails in windows. Let me know how best we can incorporate the changes.

samvidmistry:
For the third point, I was wondering if we can introduce another variable, `mu4e-mu-wsl-root` and we check `file-exists-p` for paths with and without `mu4e-mu-wsl-root`. I understand that you probably won't like it. But I'm an elisp newbie so I don't know what kind of abstractions are available to handle this scenario. Maybe we can use something like `eieio` to abstract away wsl paths from most of the code?

djcb:
> @djcb Thanks for the link.
> 
> I have updated the PR with new changes. There are 3 major changes.
> 
>     * Removing validations, because the paths are not windows.

The basic requirement would be that we use the paths that do exist in windows. Why can't that work?
Of course, the validations must stay in place.
 
>     * Update `mu4e-join-paths` to preserve one or two slashes, shrink any more sequences of slashes.

Saw that, thanks... can you make it work by just updating the regexp?
 
>     * One fork for WSL
> 
> 
> ```emacs-lisp
> (if mu4e-mu-through-wsl
>                 (concat "//wsl.localhost/Ubuntu" path)
>               path)))
> ```
> 
> This is because the path starts at `/home...`.
> 
> With this, I can properly read emails in windows. Let me know how best we can incorporate the changes.

But I think we should not use such special paths. Both `mu` and `mu4e` should use the same paths. e.g. something like
`mu init --muhome=//wsl.localhost/Ubuntu/home/user/.mu  --maildir=//wsl.localhost/home/user/Maildir` 

> For the third point, I was wondering if we can introduce another variable, `mu4e-mu-wsl-root` and we check `file-exists-p` for paths with and without `mu4e-mu-wsl-root`. I understand that you probably won't like it. But I'm an elisp newbie so I don't know what kind of abstractions are available to handle this scenario. Maybe we can use something like `eieio` to abstract away wsl paths from most of the code?

I still don't understand why that would be needed; I don't want special wsl code in mu, hopefully we can solve it without.

PS: best to configure your editor to use the same indentation/whitespace settings that mu4e does, that makes it easier to read your diffs.

samvidmistry:
> Of course, the validations must stay in place.

Of course. The validations were just getting in the way. 😄

> The basic requirement would be that we use the paths that do exist in windows. Why can't that work?

Explained below.

> But I think we should not use such special paths. Both mu and mu4e should use the same paths. e.g. something like
mu init --muhome=//wsl.localhost/Ubuntu/home/user/.mu  --maildir=//wsl.localhost/home/user/Maildir

I understand your reasoning. The reason why `mu init --muhome=//wsl.localhost/Ubuntu/home/user/.mu  --maildir=//wsl.localhost/home/user/Maildir` cannot work is because `mu` is running in WSL. In WSL, `//wsl.localhost` doesn't exist. `//wsl.localhost/Ubuntu` is where WSL filesystem is mounted on Windows.

Essentially, as long as we are talking with `mu` over the process connection, `/home/...` paths are fine since it is running in WSL. As soon as we mix emacs native functions which interact with Windows filesystem, we have to add `//wsl.localhost/Ubuntu` to the path. In case of viewing the email, we are using `insert-file-contents-literally` which interacts with windows filesystem. I took a cursory look at the commands in `mu server`. I couldn't find anything that can give us the actual contents of the email file. If we can ask `mu` for the email content, we can then simply use the `/home/...` paths. I hope I was able to explain what the issue is.

samvidmistry:
> PS: best to configure your editor to use the same indentation/whitespace settings that mu4e does, that makes it easier to read your diffs.

Welp, I think there's something wrong with the files in this repo. When I open the file in emacs on windows, the default line endings I get are `LF`, which shouldn't be because I cloned this on Windows using git for windows. At the end of each line, I get `^M`, which is the carriage return character. That makes sense because I cloned this repo on windows. It puzzles me why Emacs is not properly detecting the line endings. A quick search on google told me that this happens when files have mixed line endings. I didn't dig deeper. I assume that's why the indentation is completely off.

djcb:
> > Of course, the validations must stay in place.
> 
> Of course. The validations were just getting in the way. 😄
> 
> > The basic requirement would be that we use the paths that do exist in windows. Why can't that work?
> 
> Explained below.
> 
> > But I think we should not use such special paths. Both mu and mu4e should use the same paths. e.g. something like
> > mu init --muhome=//wsl.localhost/Ubuntu/home/user/.mu  --maildir=//wsl.localhost/home/user/Maildir
> 
> I understand your reasoning. The reason why `mu init --muhome=//wsl.localhost/Ubuntu/home/user/.mu --maildir=//wsl.localhost/home/user/Maildir` cannot work is because `mu` is running in WSL. In WSL, `//wsl.localhost` doesn't exist. `//wsl.localhost/Ubuntu` is where WSL filesystem is mounted on Windows.
> 
> Essentially, as long as we are talking with `mu` over the process connection, `/home/...` paths are fine since it is running in WSL. As soon as we mix emacs native functions which interact with Windows filesystem, we have to add `//wsl.localhost/Ubuntu` to the path. In case of viewing the email, we are using `insert-file-contents-literally` which interacts with windows filesystem. I took a cursory look at the commands in `mu server`. I couldn't find anything that can give us the actual contents of the email file. If we can ask `mu` for the email content, we can then simply use the `/home/...` paths. I hope I was able to explain what the issue is.

Ah, thanks for the explanation. So `mu4e` really depends on the real paths in many places, it's not realistic to add special code is all of those places. 

So I would suggest:
- perhaps run emacs inside WSL as well? That should work (or get you a lot closer at least)
- running Emacs in window vs mu in WSL is not something `mu` / `mu4e` are designed for; the file name thing is one, but the way mu4e seeks the binary is another; there are probably a lot more problems. Perhaps you make the file names match by some good uses of mounts / symlinks etc. so that the path to the messages look the same in either environment

But overall, I'd suggest just using Emacs inside WSL.

djcb:
> > PS: best to configure your editor to use the same indentation/whitespace settings that mu4e does, that makes it easier to read your diffs.
> 
> Welp, I think there's something wrong with the files in this repo. When I open the file in emacs on windows, the default line endings I get are `LF`, which shouldn't be because I cloned this on Windows using git for windows. At the end of each line, I get `^M`, which is the carriage return character. That makes sense because I cloned this repo on windows. It puzzles me why Emacs is not properly detecting the line endings. A quick search on google told me that this happens when files have mixed line endings. I didn't dig deeper. I assume that's why the indentation is completely off.

Can't really help you with that, sorry!

samvidmistry:
Interesting discussion @djcb. I'm going to close this PR for now. It seems I can temporarily stop using Windows. If needed, I'll create a fork and add a translation layer between mu4e and mu-server to translate paths. Thanks!

PR #2560, opened 2023-09-28
Fix defcustom :type of mu4e-headers-leave-behavior

tsdh:
The `const` values of the choice had the wrong order `(const :tag <value> <text>)`.
The correct order is `(const :tag <text> <value>)`.

This resulted in the following warning when loading mu4e:

```
⛔ Warning (emacs): Value ‘apply’ does not match type (choice (const :tag ask ask user whether to ignore marks) (const :tag apply apply marks without asking) (const :tag ignore ignore marks without asking))
```


djcb:
Merged; thanks!

PR #2569, opened 2023-10-11
Avoid claiming copyright for future years

bmwiedemann:
Avoid claiming copyright for future years

When compiling today's version in 2039,
the man-pages would state

Copyright © 2022-2039 Dirk-Jan C. Binnema. License GPLv3+: GNU GPL version 3 or later

which cannot be correct, since nobody did any copyrightable work in 2039 in that version.

Additionally, a copyright from 2023 will usually be valid for 70 years
after death of the author, so even if you forget to update it later,
there is no harm.

This patch was done while working on reproducible builds for openSUSE.

### DIFF
--- a/man/copyright.inc.in
+++ b/man/copyright.inc.in
@@ -2,7 +2,7 @@
 
 This manpage is part of ~mu~ @VERSION@.
 
-Copyright © 2008-@YEAR@ Dirk-Jan C. Binnema. License GPLv3+: GNU GPL version 3
+Copyright © 2008-2023 Dirk-Jan C. Binnema. License GPLv3+: GNU GPL version 3
 or later <https://gnu.org/licenses/gpl.html>. This is free software: you are
 free to change and redistribute it. There is NO WARRANTY, to the extent
 permitted by law.


djcb:
A copyright notice in the future is no more incorrect than compiling in the future...  

I can see it helps with reproducibility, at a (small) extra maintenance cost

bmwiedemann:
"compiling in the future" is just "compiling after waiting" - I would not call that incorrect.
"Today" is yesterday's "Tomorrow". And I bet, someone will still recompile today's maildir-utils in 3 months when 2024 started.

djcb:
If you "compile after wait", it's not the "copyright for future years" anymore... 

Anyway, I guess reproducibility is the real reason. For that, this PR is too limited, since it only changes it in one place; a full mu build also includes texinfo doc generation (under mu4e/ and guile/), which uses 'dynamic' dates as well. So I guess the way to do it is to specify the date in the top-level meson.build, and use that throughout.

djcb:
...and that's what I have done, pushed to `master`; so closing this PR -- thanks for your work!

PR #2576, opened 2023-10-28
Fix installation dir of mu.scm

bauermann:
Hello,

This fixes loading of the `(mu)` module in Guile for me.

Guile expects the `(mu)` module to be installed in `.../share/guile/site/3.0/mu.scm`, but before this patch Meson installed it in `.../share/guile/site/3.0/mu/mu.scm` instead.

If you accept this pull request and if more releases will be made from the 1.10 branch, can I kindly ask that this commit be cherry-picked there as well?

### DIFF
--- a/guile/meson.build
+++ b/guile/meson.build
@@ -93,9 +93,12 @@ if makeinfo.found()
   endif
 endif
 
-guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0', 'mu')
-install_data(['mu.scm','mu/script.scm', 'mu/message.scm', 'mu/stats.scm', 'mu/plot.scm'],
-             install_dir: guile_scm_dir)
+guile_scm_root_dir=join_paths(datadir, 'guile', 'site', '3.0',)
+install_data(['mu.scm'], install_dir: guile_scm_root_dir)
+
+guile_scm_mu_dir=join_paths(datadir, 'guile', 'site', '3.0', 'mu')
+install_data(['mu/script.scm', 'mu/message.scm', 'mu/stats.scm', 'mu/plot.scm'],
+             install_dir: guile_scm_mu_dir)
 
 
 mu_guile_scripts=[


djcb:
So currently everything goes under `/usr/local/share/guile/site/3.0/mu/`; but you want `/usr/local/share/guile/site/3.0/mu.scm` and e.g `/usr/local/share/guile/site/3.0/mu/mu/mu.scm`? (Note the `mu/mu/`, if I read the code correctly).

The first parts seems okay (but do you have some guile reference / policy / recommendation to do it like that?); the `mu/mu` part seems off.

Something like this perhaps?

```
-guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0', 'mu')
-install_data(['mu.scm','mu/script.scm', 'mu/message.scm', 'mu/stats.scm', 'mu/plot.scm'],
-             install_dir: guile_scm_dir)
-
+guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0')
+install_data(['mu.scm'], install_dir: guile_scm_dir)
+guile_scm_mu_dir=join_paths(guile_scm_dir 'mu')
+install_data(['script.scm', 'message.scm', 'stats.scm', 'plot.scm'],
+             install_dir: guile_scm_mu_dir)
```

stsquad:
I've been building with:

```
-Dguile=disabled -Dprefix=/home/alex/src/emacs/install
```

because the guile part didn't seem to follow prefix and kept trying to install outside of my home dir.

bauermann:
> So currently everything goes under `/usr/local/share/guile/site/3.0/mu/`; but you want `/usr/local/share/guile/site/3.0/mu.scm` and e.g `/usr/local/share/guile/site/3.0/mu/mu/mu.scm`? (Note the `mu/mu/`, if I read the code correctly).

Yes for `/usr/local/share/guile/site/3.0/mu.scm`, bu no for `/usr/local/share/guile/site/3.0/mu/mu/mu.scm`.

`mu.scm` doesn't need to be in both places, just the former.
 
> The first parts seems okay (but do you have some guile reference / policy / recommendation to do it like that?);

My understanding is that it isn't a policy recommendation, but a requirement for the interpreter to find the module. The Guile manual [says](https://www.gnu.org/software/guile/manual/html_node/Modules-and-the-File-System.html):

> In this case, loading `(ice-9 popen)` will eventually cause Guile to run `(primitive-load-path "ice-9/popen")`. primitive-load-path will search for a file `ice-9/popen` in the `%load-path (see [Load Paths](https://www.gnu.org/software/guile/manual/html_node/Load-Paths.html)).

And as expected from the above, without this pull request the only way I could get Guile to load the `mu` module was to load it as `(mu mu)`:

```
$ guile                               
GNU Guile 3.0.9
Copyright (C) 1995-2023 Free Software Foundation, Inc.

Guile comes with ABSOLUTELY NO WARRANTY; for details type `,show w'.
This program is free software, and you are welcome to redistribute it
under certain conditions; type `,show c' for details.

Enter `,help' for help.
scheme@(guile-user)> (use-modules (mu))
While compiling expression:
no code for module (mu)
scheme@(guile-user)> (use-modules (mu mu))
;;; note: auto-compilation is enabled, set GUILE_AUTO_COMPILE=0
;;;       or pass the --no-auto-compile argument to disable.
;;; compiling /gnu/store/1jm8kpgigq6ifspd9knnzg4y4d6i3zk2-profile/share/guile/site/3.0/mu/mu.scm
;;; compiled /home/bauermann/.cache/guile/ccache/3.0-LE-8-4.6/gnu/store/yls4gy5qv8822a0y6ayhziv2jxqdvh16-mu-with-guile-1.10.7/share/guile/site/3.0/mu/mu.scm.go
While compiling expression:
no code for module (mu mu)
scheme@(guile-user)>
```

I believe the error `no code for module (mu mu)` is because `mu.scm` defines the module `(mu)` and not `(mu mu)`, but I'm not sure about that.

Another evidence is the Debian package, which also installs with the [layout](https://packages.debian.org/sid/amd64/maildir-utils/filelist) I'm proposing.

> the `mu/mu` part seems off.

Agreed.

> 
> Something like this perhaps?
> 
> ```
> -guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0', 'mu')
> -install_data(['mu.scm','mu/script.scm', 'mu/message.scm', 'mu/stats.scm', 'mu/plot.scm'],
> -             install_dir: guile_scm_dir)
> -
> +guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0')
> +install_data(['mu.scm'], install_dir: guile_scm_dir)
> +guile_scm_mu_dir=join_paths(guile_scm_dir 'mu')
> +install_data(['script.scm', 'message.scm', 'stats.scm', 'plot.scm'],
> +             install_dir: guile_scm_mu_dir)
> ```

Yes. That will have the same end-result as my patch.

bauermann:
> I've been building with:
> 
> ```
> -Dguile=disabled -Dprefix=/home/alex/src/emacs/install
> ```
> 
> because the guile part didn't seem to follow prefix and kept trying to install outside of my home dir.

If you want to install the Guile extension, you can specify the path where it will be installed with:

`-Dguile-extension-dir=/some/path/lib`

That said, I don't know whether the fact that the `guile-extension-dir` isn't automatically put inside the `prefix`is a bug or not.

djcb:
> > So currently everything goes under `/usr/local/share/guile/site/3.0/mu/`; but you want `/usr/local/share/guile/site/3.0/mu.scm` and e.g `/usr/local/share/guile/site/3.0/mu/mu/mu.scm`? (Note the `mu/mu/`, if I read the code correctly).
> 
> Yes for `/usr/local/share/guile/site/3.0/mu.scm`, bu no for `/usr/local/share/guile/site/3.0/mu/mu/mu.scm`.
> 
> `mu.scm` doesn't need to be in both places, just the former.

Yes, my bad, I meant to write e.g. `plot.scm`  instead of that second `mu.scm`.
 
> > The first parts seems okay (but do you have some guile reference / policy / recommendation to do it like that?);
> 
> My understanding is that it isn't a policy recommendation, but a requirement for the interpreter to find the module. The Guile manual [says](https://www.gnu.org/software/guile/manual/html_node/Modules-and-the-File-System.html):

[snip]

Thanks for digging that up, and it's also in the mu guile ref; so yes.

> > Something like this perhaps?
> > ```
> > -guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0', 'mu')
> > -install_data(['mu.scm','mu/script.scm', 'mu/message.scm', 'mu/stats.scm', 'mu/plot.scm'],
> > -             install_dir: guile_scm_dir)
> > -
> > +guile_scm_dir=join_paths(datadir, 'guile', 'site', '3.0')
> > +install_data(['mu.scm'], install_dir: guile_scm_dir)
> > +guile_scm_mu_dir=join_paths(guile_scm_dir 'mu')
> > +install_data(['script.scm', 'message.scm', 'stats.scm', 'plot.scm'],
> > +             install_dir: guile_scm_mu_dir)
> > ```
> 
> Yes. That will have the same end-result as my patch.

After a few more tweaks at least ;), I've pushed the change to master, and I'll add it to the release branch too.
Thank you for your help!

djcb:
> > I've pushed the change to master, and I'll add it to the release branch too. Thank you for your help!
> 
> Thank you! Sorry for the extra tweaks you had to do.

Ah, the tweaks were for my too hastely written changes, no need to be sorry :-)

Anyway, closing this now, thanks for resolving this!

PR #2582, opened 2023-11-02
build: Add a new "tests" option to disable building tests

wavexx:
I have a script that rebuilds mu a few times, and during one of the stages I don't run tests, so I'd like to skip building them too. I'm a complete noob with meson: I kind of expected to be able to find a built-in way to skipping the build of tests, but I couldn't find any.

This adds a new "tests" feature option, which defaults to auto. When explicitly set to "disabled" it will skip the test definitions. It's kind of ugly IMHO, essentially moving all test definitions inside if blocks...

It was just an experiment on my side so I don't think much of this. Still, might be useful for packaging mu.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -94,74 +94,76 @@ list_maildirs = executable('list-maildirs', 'mu-scanner.cc',
 # unit tests
 #
 
-test('test-threads',
-     executable('test-threads',
-                'mu-query-threads.cc',
-                install: false,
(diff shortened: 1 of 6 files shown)


progfolio:
Agree that it would be nice to have more options to easily disable aspects of build script (e.g. tests, byte-compilation, etc).

djcb:
I took the idea here and added a slightly different implementation (using sub directories); so you can build with ` -Dtests=disabled`

Issue #2574, opened 2023-10-19
[mu4e bug] mu4e-action-retag-message may inadvertently capitalise tags


creichen:
**Describe the bug**
Calling `mu4e-action-retag-message` with a spec of `+foo` may result in the message being tagged with `+Foo` instead. 
This is due to `mu4e--replace-first-line-matching` not enforcing case in `replace-match`.

**How to reproduce**
I am not completely sure about the circumstances yet-- this happens in an as-of-yet unpublished experimental/WIP tool of mine when I run `mu4e-action-retag-message` after `mu4e-headers-next` (but not after `mu4e-headers-prev`), but only in Emacs 29.1.
My current argument for this being a bug is that the use of `replace-match` in `mu4e--replace-first-line-matching` clearly does not enforce preserving capitalisation (according to the docs), and that this mechanism is what `mu4e-action-retag-message` relies on for adding tags, so even if it weren't observable, it would be (needlessly) risky behaviour.  Please let me know if you find it necessary to gather more explicit dynamic evidence.

**Environment**
Debian/trixie, Emacs 29.1, mu4e 1.8.14

**Proposed Fix**
- Add `&rest` parameter to `mu4e--replace-first-line-matching`
- Use that `&rest` parameter to control the second parameter to `replace-match` in `mu4e--replace-first-line-matching` (https://github.com/djcb/mu/blob/2a28b4efbbdae72f468a249e7c88b853b02a8bc7/mu4e/mu4e-actions.el#L172)
- Pass `t` as this optional fourth parameter to `mu4e-replace-first-line-matching` in _both_ calls to that function within `mu4e-action-retag-message` (currently L226 / L230)

It may also be appropriate to always set `t` to the second parameter to `replace-match`, but I haven't had time to investigate that option.

If either approach seems reasonable to you, I can send a pull request later this week, if you'd like.

**Checklist**
- [x] you are running either an 1.8.x/1.10.x release or `master` (otherwise please upgrade)
- [x] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)
- [x] you have read all of the above

djcb:
Yes, would be useful to have specific steps where this not have the desired result. If there is, a PR is welcome too :-)

[ all that tagging code is really old, I guess nowadays it'd be better to use message-mode to do the header manipulations ]

creichen:
Sorry, took longer than planned for me to get back to it, and I cannot reproduce it any longer.
[...]

PR #2592, opened 2023-11-12
When re-tagging, force `replace-match' to preserve case

creichen:
Sets the second parameter to `replace-match' to t, thus forcing it to not attempt to auto-adjust the case of any tags.

Addresses #2574 .


djcb:
Merged, thanks.

PR #2599, opened 2023-11-19
man: fix unreproducible copyright years

a3a3el:
967b724e7546 ("build: avoid dynamic dates for reproducibility") introduced the `mu_date` variable to hard-code a build-date.  This is used for the dates embedded in the texinfo documentation, but `date` is still called to set the copyright years in the man-pages.  Use `mu_date` there too.

### DIFF
--- a/man/meson.build
+++ b/man/meson.build
@@ -17,10 +17,9 @@
 #
 # generate org include files
 #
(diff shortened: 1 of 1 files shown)


djcb:

Issue #2601, opened 2023-11-22
[mu4e bug] mu4e-compose-reply assertion failure


progfolio:
**Describe the bug**

Invoking the command `mu4e-compose-reply` signals the following error:

```emacs-lisp
mu4e-compose-reply: Assertion failed: (when reply-type (member reply-type '(wide supersede)))
```

The `REPLY-TYPE` argument is declared optional, and the `when` condition in the assertion does not account for a nil value.

**To Reproduce**

`M-x mu4e-compose-reply` from a `*mu4e-article*` buffer.

**Environment**
Arch Linux
GNU Emacs 30.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.38, cairo version 1.18.0) of 2023-11-16
mu4e @ commit: fa0248a3

**Checklist**
- [x] you are running either the latest 1.8.x/1.10.x release or `master` (otherwise, please upgrade).

djcb:
Fix merged -- thanks!

PR #2602, opened 2023-11-22
mu4e-compose-reply: account for nil REPLY-TYPE in assertion

progfolio:
Previous assertion would fail when REPLY-TYPE is nil, but nil is a valid value for the argument.

See: https://github.com/djcb/mu/issues/2601


djcb:
Oops, yes. Thanks

Issue #2608, opened 2023-12-04
mu4e--get-current-buffer-type does not return correct value for  view type buffers


progfolio:
`derived-mode-p` declared the modes as a `&rest` parameter up until 11 days ago on Emacs master.
Passing the argument as a single list here:

https://github.com/djcb/mu/blob/730bd13ea20243b48d43d494be3bb1448ae83247/mu4e/mu4e-window.el#L155

will break not work on any Emacs version prior to that change.
The old calling convention is  still supported according to:

https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=e6556db4200ccf28bd9bb033be0d5ce3cd2316a9

djcb:
Fixed already. Thanks!

PR #2609, opened 2023-12-04
mu4e: mu4e--get-current-buffer-type: fix view type

progfolio:
see: https://github.com/djcb/mu/issues/2608


djcb:
Thanks, but this brings back a byte-compiler warning (on emacs 30.x). I've fixed it earlier in slightly different way. Regardless, thank you!

PR #2612, opened 2023-12-18
mu4e: add support for macOS notifications

seanfarley:
Add support for alert.el to be used on systems that don't have native dbus functionality.

Question: is this something that can be applied to the 1.10 branch?

IMHO, this should be on 1.10 just to make 1.10 the release "where notifications work by default." Apologies for not catching this sooner.


djcb:
Looks useful, merged. Thank!

Issue #2583, opened 2023-11-02
[mu4e] Update icalendar


Chris00:
Following [your ptompt](https://github.com/djcb/mu/issues/2580#issuecomment-1789555575), I open this issue not to forget.  I did a quick check and it seems to basically work but I'll check more carefully later.

djcb:
Great! Your help is appreciated :)

djcb:
@Chris00, any updates here? Thanks!

Chris00:
> @Chris00, any updates here? Thanks!

Sorry, I'm quite busy ATM.  I did some tests and it seems to work.  However, I'd like to do more testing before calling it done.

PR #2617, opened 2024-01-03
mu4e-icalendar: Fix getting the organizer from the message

Chris00:
I did additional tests and this is the only problem I found.

Fixes https://github.com/djcb/mu/issues/2583.


djcb:
Merged; thanks!

PR #2632, opened 2024-02-25
build: add cld2 option

MatthewGentoo:
Add an option for builders to explicitly disable cld2, instead of relying on the automatic detection.


djcb:
Merged, thanks.

Issue #2631, opened 2024-02-25
[mu bug] 1.12.0 test failures


manphiz:
**Describe the bug**
3 out of 46 tests fail when building in Debian's chroot environment (sbuild).  Though the failures are deterministic, I suspect that this may be due to that the build environment produces different output compared with the Developers.  It would be good if the tests are made more robust so that they won't be affected by environment differences.

- 34: test-cmd-view
```
----------------------------------- stderr -----------------------------------
**
ERROR:../mu/mu-cmd-view.cc:334:void test_mu_view_01(): assertion failed (res->standard_out.size() == 364): (388 == 364)
==============================================================================
```
- 36: test-cmd-cfind
```
----------------------------------- stderr -----------------------------------
**
ERROR:../mu/mu-cmd-cfind.cc:336:void test_mu_cfind_plain(): assertion failed (s1s__.c_str() == s2s__.c_str()): ("M\\303\\274 testmu@testmu.xx\nHelmut Kr\\303\\266ger hk@testmu.xxx\n" == "M\303\274 testmu@testmu.xx\nHelmut Kr\303\266ger hk@testmu.xxx\n")
==============================================================================
```
- 38: test-mu-guile
```
----------------------------------- stderr -----------------------------------
guile: warning: failed to install locale
Backtrace:
           1 (primitive-load "/<<BUILDDIR>>/maildir-ut?")
           0 (setlocale 6 "")

ERROR: In procedure setlocale:
In procedure setlocale: Invalid argument
Error: something went wrong
**
ERROR:../guile/tests/test-mu-guile.cc:89:void test_something(const char*): assertion failed: (0)
==============================================================================
```

Full build log: https://gist.github.com/manphiz/76fe2b46cc685736fc9ce7bbefee3593

**Environment**
Current Debian sid/unstable

**Checklist**
- [X] you are running either the latest 1.8.x/1.10.x release or `master` (otherwise, please upgrade).
   Running 1.12.0

djcb:
Yes, we test locally and in CI (Ubuntu & MacOS): 
https://github.com/djcb/mu/actions/runs/8031147033/job/21939183124

I suspect the `sbuild` environment is so bare-bones it is missing some locale-related things. Can you find what? And what is the exact sequence of commands you use to setup / build `mu` in the sbuild environment? Thanks.

manphiz:
[...]

PR #2635, opened 2024-02-26
Fix guile_load_path to include build path

manphiz:
When meson is run in a build path that is different from source root, the generated shared library cannot be found under "<source_root>/guile".  This patch adds the corresponding path under build root to guile_root_path so that they continue to work.

This should fix https://github.com/djcb/mu/issues/2631.

### DIFF
--- a/guile/tests/meson.build
+++ b/guile/tests/meson.build
@@ -21,6 +21,7 @@
 if get_option('b_sanitize') == 'none'
   guile_load_path=':'.join([
     join_paths(meson.project_source_root(), 'guile'),
+    join_paths(meson.project_build_root(), 'guile'),
     meson.current_build_dir()])
 
   test('test-mu-guile',


guile/tests/meson.build
```diff
@@ -21,6 +21,7 @@
 if get_option('b_sanitize') == 'none'
   guile_load_path=':'.join([
     join_paths(meson.project_source_root(), 'guile'),
```
manphiz, line 23: Actually for the current status this line can be dropped, though not sure whether there will be future use of this, so I didn't do it yet.

djcb: Hmm, `GUILE_LOAD_PATH` is for Scheme source files IIRC, and we don't have any in the build-dir.

manphiz: Ack, then it's better to leave this intact.

manphiz:
Friendly ping.  This should fix the remaining issues in https://github.com/djcb/mu/issues/2631. Any concerns on merging this?

djcb:
I've merged with a small follow-up to separate the load & extension paths. Thanks!

manphiz:
Ah indeed https://github.com/djcb/mu/commit/fdce40af926884c31d6a8a821808061153a53750 makes it clearer.  Thanks!

PR #2641, opened 2024-02-29
return buf from mu4e--compose-setup

malb:
This makes using `mu4e-compose-foo` easier to use programmatically


djcb:
Thanks, I did the same already ('tho slightly different).

Issue #2636, opened 2024-02-26
[mu4e rfe] Let `display-buffer-alist` dictate compose buffers


malb:
**Is your feature request related to a problem? Please describe.**

At present, I don't see how I can pass e.g. frame parameters to the mu4e-compose frame created by mu4e.

**Describe the solution you'd like**

An option to let `display-buffer-alist` do its thing. I suggest an additional option for `mu4e-compose-switch`: `display-buffer-alist` which translates to `(mu4e--compose-switch-function)` becoming a noop. In addition, I think `switch-to-buffer` in `mu4e--compose-setup` should be replaced by `display-buffer`.

**Describe alternatives you've considered**

I've patched this behaviour as follows:

**Additional context**
```diff
diff --git a/mu4e/mu4e-compose.el b/mu4e/mu4e-compose.el
index 082665a0..2830f9c9 100644
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -776,13 +776,13 @@ Optionally, SWITCH determines how to find a buffer for the message
     (mu4e--context-autoswitch parent mu4e-compose-context-policy)
     (with-current-buffer
         (mu4e--compose-setup-buffer compose-type compose-func parent)
-      (funcall (or switch (mu4e--compose-switch-function)) (current-buffer))
+      ;; (funcall (or switch (mu4e--compose-switch-function)) (current-buffer))
       (unless (eq compose-type 'edit)
         (set-visited-file-name ;; make it a draft file
          (mu4e--draft-message-path (mu4e--message-basename) parent)))
       (mu4e--compose-setup-post compose-type parent)
       (setq buf (current-buffer))
-      (switch-to-buffer buf)
+      (display-buffer buf)
       (let* ((msgframe (selected-frame))
              (actions (list
                        (lambda () ;; kill frame when it was created for this
```

I then use:

```elisp
(defun malb/frame-dispatch (buffer alist)
  "Assigning buffers to frames."
  (cond
   ;; a mu4e compose buffer
   ((eq (buffer-local-value 'major-mode buffer) 'mu4e-compose-mode)
    (malb/display-mu4e-draft-buffer buffer alist)
    nil)

   ((eq (buffer-local-value 'major-mode (current-buffer)) 'mu4e-compose-mode)
    (malb/switch-away-from-special-and-tab (string-match-p ".*helm.*" (buffer-name buffer)))
    nil)

   ;; an mu4e buffer
[...]

PR #2643, opened 2024-02-29
allow mu4e--compose-switch-function to do nothing

malb:
This is to address #2636.


djcb:
Thanks; I've handled this slightly different; instead we now set a `display-buffer' value so we'll use `display-buffer` to, well, display the buffer.

Perhaps this could be reworked a bit, but I'd like to keep the changes small for now, in the 12.x branch.

PR #2652, opened 2024-03-02
More man page fixes

a3a3el:
Fixes for some warnings and formatting errors

### DIFF
--- a/man/meson.build
+++ b/man/meson.build
@@ -55,6 +55,7 @@ man_orgs=[
   'mu-info.1.org',
   'mu-init.1.org',
   'mu-mkdir.1.org',
+  'mu-move.1.org',
   'mu-query.7.org',
   'mu-remove.1.org',
   'mu-server.1.org',
--- a/man/meson.build
+++ b/man/meson.build
@@ -82,6 +82,7 @@ foreach src : man_orgs
   expr_tmpl = ''.join([
     '(progn',
     '  (require \'ox-man)',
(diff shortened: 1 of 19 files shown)


djcb:
Thanks, but changing to e.g., `mu_add` and rebuilding gives me (`man -l build/man/mu-add.1`) 
```
NAME
       muadd - add one or more messages to the database
```
am I missing something?

a3a3el:
Have you applied the full series?  In particular: ["_man: fix formatting of underscores_"](https://github.com/djcb/mu/pull/2652/commits/7f98345d575f1cab6fa002181e2b6655bfe6c5cd).

djcb:
Ah, testing the whole branch, I get e.g. `mu_server` in `man`.
Looking at the generated text:
```
.TH "MU-SERVER" "1"

.SH "NAME"
.PP
mu_server - the mu backend for the mu4e e-mail client
```
the underscore is unwanted.

djcb:
Oh, the `_` was by design? Yeah, I'm not a fan of that :-) And  parsing by some tools is secondary to the human reader I think.

But, a hyphen is acceptable though, since it's a bit less ugly, and since reflects the name of  the manpage (i.e., `man mu-add`); probably also why those other tools use it.

a3a3el:
Cool.  I will go with the hyphens.

a3a3el:
I've also tweaked a couple of commit messages and order the commits to put the short ones first.

PR #2657, opened 2024-03-04
Respect `message-cite-reply-position` customization

### LINKED ISSUE #2656: [mu4e bug] `message-cite-reply-position` is not respected in 1.12
When I hit `R` in the message buffer, I expect that cursor is placed according to `message-cite-reply-position` customization. Currently it's always in the beginning of the message.

**How to Reproduce**
1. `(setopt message-cite-reply-position 'below)`
2. `(setopt message-cite-function #'message-cite-original-without-signature)`
3. Open any message and hit `R` to open `mu4e-compose-mode` buffer with reply.

**Environment**
- MacOS 14
- Emacs 29.2
- mu 1.12

**Checklist**
- [x] you are running either an 1.8.x/1.10.x release or `master` (otherwise please upgrade)
- [x] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)
- [x] you have read all of the above

rrudakov:
Close #2656 

Restore some code from [previous version](https://github.com/djcb/mu/blob/v1.10.8/mu4e/mu4e-compose.el#L587-L589)

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -731,7 +731,9 @@ Is this address yours?"
       (message-goto-to)
     (if (not (message-field-value "Subject"))
         (message-goto-subject)
-      (message-goto-body)))
+      (pcase message-cite-reply-position
+        ((or 'above 'traditional) (message-goto-body))
+        (_ (when (message-goto-signature) (forward-line -2))))))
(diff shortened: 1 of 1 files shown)


djcb:
Merged, thanks.

Issue #2666, opened 2024-03-18
Calendar invite stopped working


joukeHijlkema:
**Describe the bug**
Emails with a calendar invite give the error when clicking on Accept. This worked until I upgraded to the latest version (1.12.0)

`Debugger entered--Lisp error: (error "This is a pseudo-article")
  signal(error ("This is a pseudo-article"))
  error("This is a pseudo-article")
  gnus-summary-select-article()
  gnus-summary-reply()
  gnus-icalendar-send-buffer-by-mail("*CAL*" "Accepted: Lunar Islands - Discussion on Test Defin..." "xxx.xxx@xxx.xx")
  gnus-icalendar-reply(((#<buffer  *mm*-736088> ("text/calendar" (method . "REQUEST") (charset . "utf-8")) base64 nil nil nil nil nil) accepted #<gnus-icalendar-event-request gnus-icalendar-event-request-17a852db3894>))
  gnus-article-push-button((mouse-2 (#<window 13 on *mu4e-article*> 761 (53 . 355) 849592659 nil 761 (6 . 19) nil (5 . 10) (8 . 18))))
  funcall-interactively(gnus-article-push-button (mouse-2 (#<window 13 on *mu4e-article*> 761 (53 . 355) 849592659 nil 761 (6 . 19) nil (5 . 10) (8 . 18))))
  call-interactively(gnus-article-push-button nil nil)
  command-execute(gnus-article-push-button)
`
**How to Reproduce**
open an email with an inviter and click [Accept]
[error.txt](https://github.com/djcb/mu/files/14637016/error.txt)

**Environment**
- endeauvourOS
- GNU Emacs 29.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.40, cairo version 1.18.0)
- mu/mu4e 1.12.0

**Checklist**
- [x] you are running either an 1.8.x/1.10.x release or `master` (otherwise please upgrade)
- [x] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)
- [x] you have read all of the above

Please make sure you all items in the checklist are set/met before filing the ticket. Thank you!

djcb:
Does this happen with _all_  invitation or only some? And do the ones failing still work with 1.10?

djcb:
@Chris00 any idea?

joukeHijlkema:
I'll check asap.

Chris00:
[...]

PR #2668, opened 2024-03-19
Update install instructions for icalendar

Chris00:
Update install instructions for icalendar. See https://github.com/djcb/mu/issues/2666


djcb:
Merged; thanks!

PR #2674, opened 2024-03-28
man: Escape Org heading-like lines in example

alphapapa:
The asterisks at the beginning of these lines in this example block look like headings to Org; they need to be escaped with commas.


djcb:
Merged, thanks!

PR #2675, opened 2024-03-31
mu4e-compose: add new line while replying above the cited message

VitoVan:
I think this pull request could be a mistake.

Because when I was a child, when I was reading an email with mu4e, and pressed `R`, the buffer would turn into something like this:

```markdown
To: "Jack" <jack@example.com>
Subject: Re: Jack's Nonsense
From: Vito Van <mememe@example.com>
Date: Sun, 31 Mar 2024 08:42:50 +0200
Message-ID: <m2h0lnm7sx.fsf@example.com>
--text follows this line--

"Jack" <jack@example.com> writes:

> this is Jack's nonsense
>
> [4. application/x-zip-compressed; nonsense.zip]...
``` 

and the cursor would be located just under the line of `--text follows this line--`, which is fantastic.

But after the latest update, which I didn't update anything on my computer unless it's more than two years old or I was feeling very lucky, after I pressed the magic reply key `R`, the buffer turned into something like this:

```markdown
To: "Jack" <jack@example.com>
Subject: Re: Jack's Nonsense
From: Vito Van <mememe@example.com>
Date: Sun, 31 Mar 2024 08:42:50 +0200
Message-ID: <m2h0lnm7sx.fsf@example.com>
--text follows this line--
"Jack" <jack@example.com> writes:

> this is Jack's nonsense
>
> [4. application/x-zip-compressed; nonsense.zip]...
``` 

**No new line anymore!**

And I got many letters to reply, I have to press `C-o` every time I pressed `R`, this delivered some huge mental and muscle burden. Oh, I like mu4e so! I can't switch to any other email client, please don't do this to me. I said. But after I dived into my messed-up-but-still-working Emacs configuration and git-blamed the source code of mu4e-compose.el, I gave up, and added this damn `open-line` here.

Please tell me that I'm wrong, dear djcb.

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -846,7 +846,7 @@ replied to or forwarded, etc."
     (if (not (message-field-value "Subject"))
         (message-goto-subject)
       (pcase message-cite-reply-position
-        ((or 'above 'traditional) (message-goto-body))
+        ((or 'above 'traditional) (message-goto-body) (open-line 1))
         (_ (when (message-goto-signature) (forward-line -2))))))
   ;; buffer is not user-modified yet
   (set-buffer-modified-p nil)


VitoVan:
I'm working on Emacs 29.1, btw.

djcb:
Appreciate the effort, but  the empty line is superfluous I think; I  can see the "cosmetic" improvement when editing, but only because we have the `--text follows this line--` which the recipients won't get.

If you still want the extra newline, I'd recommend using `mu4e-compose-mode-hook` in your configuration to do that -- thanks!

VitoVan:
@djcb Thank you!

I tried `mu4e-compose-mode-hook` before, but failed to add a newline at the place I wanted.

Now I tried a little harder, it worked:

```elisp
(defun v/mu4e-add-new-while-composing (&rest r)
  (goto-char (point-min))
  (search-forward-regexp mail-header-separator nil t)
  (open-line 1))

(add-hook 'mu4e-compose-mode-hook 'v/mu4e-add-new-while-composing)
```

Thank you for making mu4e, it is wonderful.

djcb:
Doesn't `message-goto-body` work?

VitoVan:
> Doesn't `message-goto-body` work?

🤦 

Yes, it worked.

I didn't know of its existence until you told me.

PR #2677, opened 2024-04-01
mu-utils: Fix build with musl

MatthewGentoo:
In musl, `stdout` is a macro that expands to `(stdout)`, and `::(stdout)` is not valid C++.

```
../mu-1.12.2/lib/utils/mu-utils.hh:268:32: error: expected id-expression before '(' token
 268 |                              ::stdout);
     |                                ^~~~~~
```

Nothing in the Mu namespace is named stdout, so it is safe to drop the `::`.

Bug: https://bugs.gentoo.org/928361


MatthewGentoo:
Would you accept a change to add Alpine Linux, or Gentoo Linux w/ musl to the CI workflow?

djcb:
Merged, thanks.

djcb:
> Would you accept a change to add Alpine Linux, or Gentoo Linux w/ musl to the CI workflow?

Sure, assuming after this PR that would work.

PR #2696, opened 2024-04-19
mu-move: remove debug print

fbenkstein:
The `mu move` command always prints messages to `stderr` which seem like they were left over from some debugging session. These are hard to filter out without also potentially suppressing error messages. They're also present when giving the `--quiet` option. I don't think these serve any purpose at the moment, so I suggest to just remove them.


djcb:
Yeah, that was some leftover debugging stuff. Merged, thanks.

PR #2698, opened 2024-04-22
Fix: Put sent mail in correct location when `mu4e-sent-messages-behavior` is a function

josephmturner:
The second commit is just a stylistic change.  If `pcase-exhaustive` is not preferred in this case, I'd be curious to hear why.

Thank you for all your work on `mu4e`!!!

Joseph


djcb:
Ah, good catch!

mu4e/mu4e-compose.el
```diff
@@ -330,12 +330,18 @@ Returns the path for the sent message, either in the sent or
...
   (when-let ((sent-dir
-          (pcase mu4e-sent-messages-behavior
-            ('delete nil)
-            ('trash (mu4e-get-trash-folder parent))
-            ('sent (mu4e-get-sent-folder parent))
-            ((pred functionp) (funcall mu4e-sent-messages-behavior))
-            (_ (mu4e-error "Error in `mu4e-sent-messages-behavior'")))))
+              ;; We can't call `functionp' since at least 'delete is a function.
+              (pcase mu4e-sent-messages-behavior
+                ('delete nil)
```
djcb, line 335: Would be nice not to duplicate the cases, perhaps first calculate the value?

mu4e/mu4e-compose.el
```diff
@@ -331,17 +331,15 @@ trash folder, or nil if the message should be removed after
...
               ;; We can't call `functionp' since at least 'delete is a function.
-              (pcase mu4e-sent-messages-behavior
+              (pcase-exhaustive mu4e-sent-messages-behavior
                 ('delete nil)
                 ('trash (mu4e-get-trash-folder parent))
                 ('sent (mu4e-get-sent-folder parent))
                 ((pred functionp)
-                 (pcase (funcall mu4e-sent-messages-behavior)
+                 (pcase-exhaustive (funcall mu4e-sent-messages-behavior)
                    ('delete nil)
```
djcb, line 340: No biggie, but In this case, I suppose it's nice to give a  mu4e-specific error. It's more informative than `pcase-exhaustive`...

mu4e/mu4e-compose.el
josephmturner: How about the force-pushed change?  Another alternative is to use a couple of `if`s instead of `pcase` in the first `when-let` clause.

josephmturner:
I've pushed the changes.  Feel free to squash merge if you like.  Thank you!!

djcb:
Okay, I'll merge it now, thanks! I'll tweak it  bit later, but anyway the eternal glory goes to you for fixing the problem!

josephmturner:
Thank you!!

PR #2716, opened 2024-06-17
Ensure mu4e-compose-pre-hook has access to parent message and compose type

### LINKED ISSUE #2715: [mu4e bug] mu4e-compose-pre-hook is broken in v1.12+
**Describe the bug**

Due to an old known bug in date time formatting, my setup relies on preparatory work done for every draft. I expose it in this old issue: https://github.com/djcb/mu/issues/2113#issuecomment-908486670

To make it short, I temporarily store the `mu4e-compose-parent-message` in a variable, to be able to get it again to generate the citation line (historically the parent-message variable was lost while generating the citation line. Maybe this is not the case anymore, but this is another subject).

Please note the documentation for the `mu4e-compose-pre-hook` still say the following:

> Hook run just *before* message composition starts.
>
> If the compose-type is a symbol, either ‘reply’ or ‘forward’, the
> variable ‘mu4e-compose-parent-message’ is the message replied to
> / being forwarded / edited, and ‘mu4e-compose-type’ contains the
> type of message to be composed.

However, in version 1.12.0 through 1.12.5 the mu4e-compose-pre-hook call has disapeared and any functions added to the hook is never called. From https://github.com/djcb/mu/commit/3ff2f9f1ac87369f7f10ffbf769ea3404bae0dec the hook is called again… but too soon in the chain: `mu4e-compose-parent-message` and `mu4e-compose-type` are only set *after* the hook is run in the function `mu4e--draft`

**How to Reproduce**

- Add a function to `mu4e-compose-pre-hook`
- Start a draft

The function is never called, or after https://github.com/djcb/mu/commit/3ff2f9f1ac87369f7f10ffbf769ea3404bae0dec, they are called but `mu4e-compose-parent-message` and `mu4e-compose-type` are nil.

**Possible solution**

If I may, I’d have moved the following block:

```elisp
(cl-assert (member compose-type '(reply forward edit new)))
  (cl-assert (eq (if parent t nil)
                 (if (member compose-type '(reply forward)) t nil)))
  ;; remember some variables, e.g for user hooks. These are permanent-local
  ;; hence survive the mode-switch below (we do this so these useful vars are
  ;; available in mode-hooks.
  (setq-local
   mu4e-compose-parent-message parent
   mu4e-compose-type compose-type)
```

from the function `mu4e--prepare-draft-buffer` to the function `mu4e--prepare-draft`, *before* calling the pre-hook.

However this implies to call `mu4e--prepare-draft` with the compose-type as parameter, what is not possible today, and I’m not sure about the implication of such a move.

**Environment**

Emacs 29.3 with mu 1.12.5

**Checklist**

- [X] you are running either an 1.10.x/1.12.x release or `master` (otherwise please upgrade)
- [X] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)

milouse:
This MR try to fix #2715

### DIFF
--- a/mu4e/mu4e-compose.el
+++ b/mu4e/mu4e-compose.el
@@ -461,7 +461,7 @@ The message is resent as-is, without any editing. See
           "Resend message to address: " mu4e--contacts-set)))
   (let ((msg (mu4e-message-at-point)))
     (with-temp-buffer
-      (mu4e--prepare-draft msg)
+      (mu4e--prepare-draft 'new msg) ;; Resend as fake new?
       (insert-file-contents (mu4e-message-readable-path msg))
       (message-resend address))))
 
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -556,12 +556,14 @@ while in a buffer with the to-be-forwarded/replied-to message."
   "Message headers to hide when composing.
 This is mu4e's version of `message-hidden-headers'.")
 
-(defun mu4e--prepare-draft (&optional parent)
+(defun mu4e--prepare-draft (compose-type &optional parent)
   "Get ready for message composition.
 PARENT is the parent message, if any."
+  (cl-assert (member compose-type '(reply forward edit new)))
+  (cl-assert (eq (if parent t nil)
+                 (if (member compose-type '(reply forward)) t nil)))
   (unless (mu4e-running-p) (mu4e 'background)) ;; start if needed
-  (mu4e--context-autoswitch parent mu4e-compose-context-policy)
-  (run-hooks 'mu4e-compose-pre-hook))
+  (mu4e--context-autoswitch parent mu4e-compose-context-policy))
 
 (defun mu4e--prepare-draft-headers (compose-type)
   "Add extra headers for message based on COMPOSE-TYPE."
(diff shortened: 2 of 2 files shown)


mu4e/mu4e-compose.el
```diff
@@ -461,7 +461,7 @@ The message is resent as-is, without any editing. See
           "Resend message to address: " mu4e--contacts-set)))
   (let ((msg (mu4e-message-at-point)))
     (with-temp-buffer
-      (mu4e--prepare-draft msg)
+      (mu4e--prepare-draft 'new msg) ;; Resend as fake new?
```
milouse, line 464: @djcb This is the point I’m the less sure about. Do not hesitate to comment if this is completely wrong.

mu4e/mu4e-draft.el
```diff
@@ -576,16 +578,6 @@ PARENT is the parent message, if any."
...
-  (cl-assert (member compose-type '(reply forward edit new)))
-  (cl-assert (eq (if parent t nil)
-                 (if (member compose-type '(reply forward)) t nil)))
-  ;; remember some variables, e.g for user hooks. These are permanent-local
-  ;; hence survive the mode-switch below (we do this so these useful vars are
-  ;; available in mode-hooks.
-  (setq-local
-   mu4e-compose-parent-message parent
-   mu4e-compose-type compose-type)
-
```
milouse, line 588: I’ve cut this part in 2 block:

1. move the cl-assert to the `mu4e--prepare-draft` function (to keep the semantic of "preparing" stuff, ensuring everything is ok)
2. move the setq-local statement later in the `mu4e--draft` function to still benefit from being call inside the target buffer (as now they are buffer-local, and `mu4e--prepare-draft` is called before the buffer is created.

djcb:
Thank you! The poroblem this addresses has been solved, so closing this PR.

PR #2720, opened 2024-06-27
fix: readme typo >> clone through ssh error

hgp22:
Most useful pull request.

### DIFF
--- a/README.org
+++ b/README.org
@@ -76,7 +76,7 @@ Help is welcome!
 ** Building
 
 #+begin_example
-$ git clone git@github.com:djcb/mu.git
+$ git clone https://github.com/djcb/mu.git
 $ cd mu
 #+end_example
 
--- a/README.org
+++ b/README.org
@@ -76,7 +76,7 @@ Help is welcome!
 ** Building
 
 #+begin_example
-$ git clone git://github.com/djcb/mu.git
+$ git clone git@github.com:djcb/mu.git
 $ cd mu
 #+end_example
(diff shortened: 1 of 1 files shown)


README.org
```diff
@@ -76,7 +76,7 @@ Help is welcome!
 ** Building
 
 #+begin_example
-$ git clone git://github.com/djcb/mu.git
+$ git clone git@github.com:djcb/mu.git
```
djcb, line 79: Think it should be `https://github.com/djcb/mu.git`, or? Fewer things can go wrong that way

hgp22: I tried to clone the repo, copy & paste from the readme, and it didnt work.
This was the input/output:
```
❯ git clone git://github.com/djcb/mu.git
Cloning into 'mu'...
fatal: unable to connect to github.com:
github.com[0: 140.82.121.4]: errno=Connection timed out
```
I just tried it again and gave me the same error.

djcb: Yeah, `git://` won't work but `https://`:
```
git clone https://github.com/djcb/mu.git
Cloning into 'mu'...
remote: Enumerating objects: 37259, done.
```

hgp22: The link on the readme was not working.
I prefer ssh, but whatever works.
```suggestion
$ git clone https://github.com/djcb/mu.git
```

djcb:
Thanks! Merged. Yeah, prefer ssh as well, but https is more likely to work.

PR #2721, opened 2024-06-28
Mu4e info manual typos fix

Tristan02d:
I think I found two typos in the Mu4e Info manual. Here is a fix.


djcb:
Only two? :-). Thanks & merged.

Tristan02d:
I have not read the entire manual yet. Thank you !

PR #2728, opened 2024-07-22
Try to improve overall manual page typesetting

Tristan02d:
I tried to make the typesetting of the numerous manual pages more consistent and
more compliant to the standards described in man-pages(7). There are quite a lot
of changes I think, hence I tried to scatter them in different commits to make
it easier to figure out what is going on and to allow picking only certain
changes, in case some of them are not wanted.

There is one major thing to notice. In order to get the proper output for
manual pages references, and synopses, I used *zero width spaces* which I
learned about [here](https://orgmode.org/manual/Emphasis-and-Monospace.html).
This is admittedly a little hacky, there might be a better way to get the same
outputs. It renders fine in standard manual pages, but if you plan to export the
Org sources to other formats, I do not know what it may look like.

### DIFF
--- a/man/common-options.inc
+++ b/man/common-options.inc
@@ -1,30 +1,30 @@
 * COMMON OPTIONS
 
 ** -d, --debug
-makes mu generate extra debug information, useful for debugging the program
-itself. By default, debug information goes to the log file, ~/.cache/mu/mu.log.
-It can safely be deleted when mu is not running. When running with --debug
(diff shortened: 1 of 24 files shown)


djcb:
Thank you, nice work! I'm merging this, thanks about the note re: zero-width-space, I'll keep that in mind!

Tristan02d:
Nice, thank you ! Apart from synopses, the only zero-width space is in the `man-link` macro, so you can easily replace it if needed.

PR #2736, opened 2024-08-03
Make toggling search properties easier to discover

Chris00:
For some reason (probably because I don't use it so often), I forget how to toggle the search properties.  Adding a menu entry makes it easier to (re)discover.


djcb:
Merged, thanks.

PR #2738, opened 2024-08-06
Canonicalize init --maildir argument after expanding

rabuu:
Hey,
when I tried to set up `mu` I ran into the following issue:
I ran `mu init --maildir ~/mail` which worked fine but when I tried to do `mu index`
no mails where found. After some investigating I noticed `mu` stores my maildir as `/home/USER//mail` but then disregards mails with the canonical path `/home/USER/mail/...` because they don't have this exact prefix.

I thought it would make sense to store the maildir (which is already ensured to be absolute) as canonical path too and therefore canonicalize the `mu init --maildir` argument after expanding.

This fixes my problem, I hope it doesn't break something else. Feedback is appreciated. :)


djcb:
Looks good, merged. Thanks.

PR #2751, opened 2024-08-28
Ensure `User-Agent' is set

### LINKED ISSUE #2750: [mu4e bug] User-Agent header not set
**Describe the bug**

It looks like mu4e doesn't set `User-Agent` header anymore.  It used to do that when I was using 1.8.x.  I stopped using it for a while and started trying it again since 1.12.5 and I can no longer see `User-Agent` being set.  I suspect this may be happening for earlier versions as well.

**How to Reproduce**

Sending a mail and show all headers, and there is no `User-Agent` set.

**Environment**

Debian Linux stable (Bookworm)
Emacs 29.4
mu4e 1.12.6

**Checklist**

- [x] you are running either an 1.10.x/1.12.x release or `master` (otherwise please upgrade)
- [x] you can reproduce the problem without 3rd party extensions (including Doom/Evil, various extensions etc.)
- [x] you have read all of the above

**Possible Cause**

`'User-Agent` is included in the draft header when `message-newsreader` is set[1].  However, the only place that it's being set is at [2] where it's let-bounded, so it has no effect on the code at [1].

[1] https://github.com/djcb/mu/blob/f101e0d4ca86b46718e0e3d46dee04a6f8b77bc4/mu4e/mu4e-draft.el#L573
[2] https://github.com/djcb/mu/blob/f101e0d4ca86b46718e0e3d46dee04a6f8b77bc4/mu4e/mu4e-draft.el#L517

manphiz:
* Set `message-newsreader' before calling `mu4e--prepare-draft-buffer' so that `User-Agent' is properly set.

* Fixes #2750.

(Though I'm not very certain this is the best place to set this.  Please advice.)

### DIFF
--- a/mu4e/mu4e-draft.el
+++ b/mu4e/mu4e-draft.el
@@ -723,7 +723,8 @@ Returns the new buffer."
   ;; compose-func
   (let ((draft-buffer)
         (oldframe (selected-frame))
-        (oldwinconf (current-window-configuration)))
+        (oldwinconf (current-window-configuration))
+        (message-newsreader mu4e-user-agent-string))
     (with-temp-buffer
       ;; provide a temp buffer so the compose-func can do its thing
       (setq draft-buffer (mu4e--validate-hidden-buffer (funcall compose-func)))


djcb:
Thank you! 

I merged your PR, then slightly moved the place where the change was done in a follow-up commit.

manphiz:
Thanks!  And that's indeed a better way to do it.  TIL!

Issue #2753, opened 2024-09-02
[misc] The zero width space in man-link prevents emacs from referencing related man pages


quaepoena:
**Describe the issue**

The use of a zero width space (0x200B) in the `man-link` macro prevents Emacs' man page viewer from parsing the SEE ALSO section of the mu* man pages, e.g. with `Man-follow-manual-reference`. I understand why it's there, and I don't see a way to fix this issue that involves changing the macro itself.

I was somewhat able to fix this, though, with the following code and then running `C-c C-e M m` or `(org-export-to-buffer 'man "mu.1.org-out")`.

```elisp
;; Replace the ZERO WIDTH SPACE inserted by the man-link macro.
(defun qp-man-filter-remove-zero-width-space (text backend _info)
  (when (org-export-derived-backend-p backend 'man)
    (replace-regexp-in-string "​" "" text)))
(add-to-list 'org-export-filter-plain-text-functions 'qp-man-filter-remove-zero-width-space)
```

When I tried to add that code to `man/meson.build`, however, the resulting files in `build/man/` still have the zero width spaces.

```elisp
expr_tmpl = ''.join([
  '(progn',
  '  (require \'ox-man)',
  '  (setq org-export-with-sub-superscripts \'{})',
  '  (org-export-to-file \'man "@0@")',
  '  (defun qp-man-filter-remove-zero-width-space (text backend _info)',
  '    (when (org-export-derived-backend-p backend \'man)',
  '      (replace-regexp-in-string "​" "" text)))',
  '  (add-to-list \'org-export-filter-plain-text-functions \'qp-man-filter-remove-zero-width-space))'])
```

An example line exported manually:

> \fBmu-add\fP(1),

An example line exported from make:

> \fBmu-add\fP​(1),

I know very little about exporting from org and building with meson, so I don't have any further ideas about how to solve this.

**Environment**

Emacs 29.4
mu 1.12.6

djcb:
@Tristan02d: any thoughts on this? Thanks!

quaepoena:
[Suggested fix](https://github.com/quaepoena/mu/commit/7365b9842e03d66bbac4fa06cc8a237912909c51) using a lambda and no `when`-statement.

Tristan02d:
All these spaces do is deleting an extra visual space when manual pages are
rendered in either the terminal or Emacs. This small enhancement might not be
worth adding too much complexity to the build system.  For now the provided fix
works well and is simple, so I think keeping these spaces and merging it is
[...]

PR #2755, opened 2024-09-04
Replace zero-width spaces in man pages with an empty string.

quaepoena:
The zero-width spaces are added by the `man-link` macro to ensure bold text formatting when exporting from org, but they interfere with man page functionality in Emacs.

Fixes https://github.com/djcb/mu/issues/2753. Let me know if you'd like the comment to be more/less wordy.


djcb:
Merged, thank you!

PR #2757, opened 2024-09-11
Use <DEL> for scroll-down in mu4e-view-mode like in Gnus

manphiz:
* It looks like `<DEL>` does not work in mu4e-view-mode, while it does in mu4e-raw-view-mode which inherits keybindings from view-mode.
* "S-SPC" does not seem to work reliably under terminal either.
* Adding `<DEL>` make it more consistent with existing paging as in info, Gnus, etc.

### DIFF
--- a/mu4e/mu4e-view.el
+++ b/mu4e/mu4e-view.el
@@ -930,6 +930,7 @@ This is useful for advising some Gnus-functionality that does not work in mu4e."
 
     ;; intra-message navigation
     (define-key map (kbd "S-SPC") #'scroll-down)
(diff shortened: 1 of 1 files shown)


manphiz:
Not sure whether this is acceptable but probably worth a try.

Also I'm aware there exists a binding on `<backspace>` for `#'mu4e-scroll-down` but it didn't work for me, probably because I'm mainly under terminal.  If possible, I would suggest for `<DEL>` to be preferred for `scroll-down` to be consistent with other modes.

manphiz:
Understood.  Thanks for clarifying.

PR #2765, opened 2024-09-25
Clarify how the --my-address option works

formbi-kupiec:
Rephrased the fragment about the `--my-address` option and mentioned the example (gotta admit, I didn't think to look there). I think this version is a bit easier to understand.


djcb:
Ah, the beams crossed, I already reworded it in slightly different fashion. Thanks nevertheless!

PR #2773, opened 2024-10-19
doc(mu4e--server-move): typos

zuphinx:
Noticed documentation related typos while inspecting the function mu4e--server-move.

### DIFF
--- a/mu4e/mu4e-server.el
+++ b/mu4e/mu4e-server.el
@@ -664,12 +664,12 @@ The server reports the results for the operation through
 `mu4e-update-func'.
 
 If the variable `mu4e-change-filenames-when-moving' is
-non-nil, moving to a different maildir generates new names forq
+non-nil, moving to a different maildir generates new names for
 the target files; this helps certain tools (such as mbsync).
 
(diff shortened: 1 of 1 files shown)


mu4e/mu4e-server.el
```diff
@@ -643,7 +643,7 @@ PATH must be below the root-maildir."
 Optionally to MAILDIR and optionally setting FLAGS. If MAILDIR is
 nil, message will be moved within the same maildir.
 
-At least one of MAILDIR and FLAGS must be specified. Note that
+At least one of MAILDIR or FLAGS must be specified. Note that
 even when MAILDIR is nil, this is still a filesystem move, since
```
djcb, line 647: I think the "and" is fine here (same below), so let's leave that it is

djcb (changes requested):
Thanks for the contribution! Just some minor changes.

mu4e/mu4e-server.el
zuphinx: Okay. I amended the commit.

djcb:
Merged, thank you!

PR #2786, opened 2024-11-13
mu4e-actions: ensure empty tag removal on message retagging.

mstenek:
Somehow, empty tags creep into the tag list; so now `, tag1, tag2` becomes `tag1, tag2`.


djcb:
Merged, thanks!

PR #2794, opened 2024-12-11
Fix quotation mu4e-contexts example

xuhdev:
The `concat` function should be evaluated in the example.


djcb:
For the `message-signature`, if it's a function or a form, the result is used. I.e, we only evaluate it later. For `concat` the result is the same of course.

djcb:
(see the `message-signature` docstring)

xuhdev:
Got it, thanks for the explanation!

PR #2795, opened 2024-12-11
Fix a potentially misnamed function

xuhdev:
`(add-hook 'mu4e-thread-mode-hook #'mu4e-thread-fold-apply-all)` doesn't seem to fold threads. Did you mean `(add-hook 'mu4e-thread-mode-hook #'mu4e-thread-fold-all)`?


djcb:
Yes, good catch! Merged.

PR #2798, opened 2024-12-21
mu4e-org: autoload org-link-set-parameters

josephmturner:
With this change, mu4e: links now load mu4e if it wasn't loaded already.

### DIFF
--- a/mu4e/mu4e-org.el
+++ b/mu4e/mu4e-org.el
@@ -142,6 +142,7 @@ it with org)."
   (org-capture))
 
 ;; install mu4e-link support.
+;;;###autoload
 (org-link-set-parameters "mu4e"
                          :follow #'mu4e-org-open
                          :store  #'mu4e-org-store-link)


djcb:
Merged, thanks.

wavexx:
This causes a call to (org-link-set-parameters) when loaddefs.el is loaded, which is not what we want..  this is causing an init failure for me.

djcb:
@wavexx disabled the change for now. What exact error did you get?

wavexx:
Attempting to call underfined (org-link-set-parameters) while loading loaddefs.el. Which makes sense, since this cookie is just putting the _call_ into loaddefs, but the definition is not available yet.

mu4e-org.el should require 'ol to make org-link-set-parameters defined instead.

josephmturner:
I am sorry to inconvenience you, @wavexx! Thank you for reporting the issue, and thank you for reverting my mistake so quickly @djcb!!

I am able to reproduce the issue by checking out 62e4b079 (just before this PR was reverted), then generating `mu4e-autoloads.el` by running `M-x package-vc-install-from-checkout`. Then if I run `emacs -Q --eval "(load \"/home/joseph/.local/src/mu/mu4e/mu4e-autoloads.el\")"`, I get the same error `load-with-code-conversion: Symbol’s function definition is void: org-link-set-parameters`.

Adding `(require 'ol)` to `mu4e-org.el` does not solve the issue.  `mu4e-org.el` already requires `org.el`, which in turn requires `ol.el`.

I think the solution is for `org-link-set-parameters` to be autoloaded inside `ol`. I have submitted a patch to the Org mode mailing list.  If it is applied, it will only be available in `org 9.7.21`. @djcb how would you like to handle this feature which will require a bleeding edge of `org`?  Or perhaps you have another solution? :)

Thank you!

Joseph

djcb:
@josephmturner we can add the autoload then when the fix is available in the minimum emacs version supported by mu4e... which is currently 26.x (at least _officially_). 

Or perhaps there's some different way, but it seems like a fairly minor thing (at least for me).

josephmturner:
> we can add the autoload then when the fix is available in the minimum emacs version supported by mu4e... which is currently 26.x (at least _officially_).

Sounds good to me.  Shall I make a new issue as a reminder for then?

josephmturner:
In [this response](https://yhetil.org/orgmode/87o6ys694l.fsf@localhost/T/#m5f3c8073e28f46d14804cbcac100b816420a6963), Ihor points out that autoloading `org-link-set-parameters` cannot be added to Org without causing confusing bugs related to multiple versions of Org being loaded simultaneously.

Let's forget about this issue entirely.  It's no big deal to require `mu4e` before attempting follow `mu4e:` links.  Thank you!!

djcb:
Perhaps it could be done the other way around, adding something to `org-modules`.  But I'll leave that to others :-)

PR #2801, opened 2025-01-03
guile: Fix the mu:body message method

bauermann:
Hello,

I wanted to use the `mu:body-txt` method mentioned in the mu-guile documentation, but it didn't work.

This branch renames it to `mu:body`, removes `mu:body-html` (since the `BodyHtml` field doesn't exist anymore) and updates the test and documentation.

Please let me know if there are any changes or improvements that should be made.


guile/mu.scm
```diff
@@ -38,8 +38,7 @@
     mu:header
     ;; message accessors
     mu:field:bcc
-    mu:field:body-html
-    mu:field:body-txt
+    mu:field:body
     mu:field:cc
```
djcb, line 42: Although the Mu Guile API is deprecated, people may depend on it, so wouldn't want to add backward-incompatibilities if it can be avoided.

So, it's okay to remove body-txt/body-html from the documentation, but keep the functions around; the former as a synonym for `body` and the latter always returning `#f`.

testdata/testdir2/wom_bat/cur/atomic
```diff
@@ -4,8 +4,8 @@ Subject: atoms
 To: "Democritus" <demo@example.com>
 Message-id: <3BE9E6535E302944823E7A1A20D852173@msg.id>
 MIME-version: 1.0
-Content-type: text/plain; charset=us-ascii
-Content-transfer-encoding: 7BIT
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
 Precedence: high
```
djcb, line 10: Is this change needed?

djcb (commented):
Looks good, with a few changes.

guile/mu.scm
bauermann: Makes sense. I'll send an updated version with that change.
Is it ok if I force-push to this branch?

testdata/testdir2/wom_bat/cur/atomic
bauermann: Yes, otherwise the `msg:body` test fails, because with the wrong charset the em dash is skipped.
There are two alternatives:
1. Substitute the em dash with a regular hyphen.
2. Change the test to expect the wrong text, with the em dash omitted.

Is the charset wrong on purpose?

guile/mu.scm
bauermann: I just force-pushed a version implementing the change.

testdata/testdir2/wom_bat/cur/atomic
bauermann: I didn't change this part in the new version.

djcb:
Thanks, looks good! Merged.

PR #2814, opened 2025-02-02
mu4e-resize-linked-headers-window: fix empty when-let* body

progfolio:
Likely a syntax error


djcb:
Ah, you also got the warning from the freshly-built emacs-master? Already had a fix, but thank you nevertheless!

PR #2820, opened 2025-02-22
Use named function, not lambda, for refresh timer

dcolascione:
This change turns the mu4e refresh timer into a named function, making M-x list-timers output display a name having to do with mu4e instead of an inscrutable "anonymous lambda" line.


djcb:
Merged, thanks.

PR #2821, opened 2025-02-22
Improve indexing performance: batch stat()s with io_uring

dcolascione:
Add support for using io_uring to batch stat() calls when scanning maildir directories. This approach reduces the number of individual syscalls by processing stats in batches of up to 16384 files at a time; it also allows the kernel to continue doing kernel-internal stat()s while we do indexing work.

Performance impact is moderate but noticeable on a real-world maildir with ~490k messages:

  Without io_uring: 17.5s real time (0.75s user, 5.6s sys)
  With io_uring:    15.4s real time (0.74s user, 8.4s sys)

The higher sys time with io_uring reflects the batch processing happening in kernel space rather than repeated userspace->kernel transitions.

Only enable io_uring for maildir directories since:
1. They're the only directories likely to be large enough to benefit
2. This ensures the io_uring instance isn't used concurrently

The feature can be disabled at runtime via MU_DISABLE_IO_URING=1 or at build time via -Diouring=disabled. Requires liburing >= 2.3.

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -51,12 +51,15 @@ lib_mu=static_library(
     guile_dep,
     config_h_dep,
     lib_mu_utils_dep,
-    lib_mu_message_dep],
+    lib_mu_message_dep
+  ],
   install: false)
+
+lib_mu_dep_list = [ lib_mu_message_dep, thread_dep ]
 
 lib_mu_dep = declare_dependency(
   link_with: lib_mu,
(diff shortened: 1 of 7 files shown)


djcb:
Looks interesting. Will take a bit longer to review this (I'm not too familiar with io_uring), I don't think before 1.12.9 (which should be in the coming week)

The performance improvements are modest, so we'll have to weigh that against the added complexity. Anyway, thanks for doing this!

PR #2822, opened 2025-02-22
Call file-readable-p only on non-remote files

dcolascione:
Calling file-readable-p before reading a message file is fine when the file is local, but when it's a remote file, doing so incurs an extra network round trip.  It's not really necessary to do so either: if the file is missing, we'll fail in the insert-file-contents below anyway, so we're missing out only on a nicer warning message.


djcb:
Merged, thanks.

PR #2845, opened 2025-04-24
mu4e.texi: fix smart refiling example config

tmuehlbacher:
Noticed that this snippet doesn't work. Other than that I am a very happy new user. Thanks!


djcb:
Welcome :-) Thanks, merged!

PR #2835, opened 2025-04-04
mu4e-notification: close notification once done

matperc:
I will try it for some day to see if this works correctly, I will un-draft it when I am done with my testing.

Thank you for considering this.

Closes #2833

PR #2851, opened 2025-05-25
mu4e-notification: ignore dbus errors

sarg:
notifications protocol says that an empty dbus error message is sent back when the notification to close no longer exists: https://specifications.freedesktop.org/notification-spec/latest/protocol.html#command-close-notification

Some notification daemons don't signal errors, but others do (e.g. EDNC). Adjust the code to be conformant to the spec.

Extends #2835 logic.

### DIFF
--- a/mu4e/mu4e-notification.el
+++ b/mu4e/mu4e-notification.el
@@ -73,9 +73,11 @@ zero."
   (when-let* ((fav (mu4e-bookmark-favorite))
               (delta-unread (plist-get fav :delta-unread)))
     (when (and (fboundp 'notifications-close-notification)
+               (fboundp 'dbus-ignore-errors)
                mu4e--notification-id
                (zerop delta-unread))
-      (notifications-close-notification mu4e--notification-id)
+      (dbus-ignore-errors
+        (notifications-close-notification mu4e--notification-id))
       (setq mu4e--notification-id nil))
     (when (and (> delta-unread 0)
                (not (= delta-unread mu4e--last-delta-unread)))


mu4e/mu4e-notification.el
```diff
@@ -73,9 +73,11 @@ zero."
   (when-let* ((fav (mu4e-bookmark-favorite))
               (delta-unread (plist-get fav :delta-unread)))
     (when (and (fboundp 'notifications-close-notification)
+               (fboundp 'dbus-ignore-errors)
                mu4e--notification-id
                (zerop delta-unread))
```
djcb, line 78: when would `dbus-ignore-errors` _not_ be defined? Some emacs version, or in general when DBUS-support is not available?

mu4e/mu4e-notification.el
```diff
@@ -73,9 +73,11 @@ zero."
...
-      (notifications-close-notification mu4e--notification-id)
+      (dbus-ignore-errors
+        (notifications-close-notification mu4e--notification-id))
       (setq mu4e--notification-id nil))
```
djcb, line 81: Presumably, we should still be able to close notifications, even if `dbus-ignore-errors` is not defined?

sarg:
hi @djcb, is this PR good to merge?

djcb:
Yes, thanks! Merged.

djcb:
Hmm, I'm seeing 
```
mu4e--default-notification-filter: Invalid function: dbus-ignore-errors
```
Is something missing?

sarg:
> Hmm, I'm seeing
> 
> ```
> mu4e--default-notification-filter: Invalid function: dbus-ignore-errors
> ```
> 
> Is something missing?

maybe explicit `(require 'dbus)`? I suppose this happens only when byte-compiling?

PR #2856, opened 2025-06-30
indexer sleep exp backoff

pingz:
improve the performace of indexer with exponential backoff for doubling sleep duration.

I already tested it on my own Linux box. For a small maildir it will significantly improve the user experience.


djcb:
Guess it might make a difference if there's not much to index. I've merged it, with a few tweaks.

PR #2863, opened 2025-08-05
Remove unnecessary leading white-space in buffer names

s10az:
I noticed there are leading white-space in buffer names which is probably unintended so I removed them.


djcb:
The spaces are intentional: 
https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-Names.html

s10az:
Ahh i see, thank you for the explanation. I will close this.

PR #2881, opened 2025-10-22
Add the mu4e-view-mark-and-next function

thomasthiriez:
There was no function to mark a message with a custom mark from the message view.

Now, there is mu4e-view-mark-and-next.


djcb:
Merged, thanks!

PR #2892, opened 2025-12-19
mu4e: save all attachments if none are selected

taingra:
Updates the behavior of `mu4e-view-save-attachments' to allow saving all attachments.  If no files are selected in the completing-read then assume all files should be downloaded to the selected directory.

This is a small change that I wanted for my own setup where I frequently need to download multiple attachments at once.   Please let me know if you'd like any tweaks to this change.


taingra:
Apologies I just noticed I accidentally committed some whitespace changes.  Let me revert and re-commit.

PR #2893, opened 2025-12-20
mu4e: save all attachments if none are selected

taingra:
Updates the behavior of `mu4e-view-save-attachments' to allow saving all attachments.  If no files are selected in the completing-read then assume all files should be downloaded to the selected directory.

This is a small change that I wanted for my own setup where I frequently need to download multiple attachments at once. Please let me know if you'd like any tweaks to this change.


djcb:
Nice idea! Traveling now, but I'll take a closer look soon.

djcb:
Thanks, merged.

taingra:
Thanks!!  Mu is a great tool, thanks for all your work on it!

--
Thomas Ingram
https://taingram.org/

PR #2903, opened 2026-02-17
New incremental cleanup strategy: 63%-88% faster

dcolascione:
This change adds a new cleanup mode that avoids cleanup having re-traverse the directories the index pass just looked at. Additionally, we efficiently query the Xapian database by walking the term list instead of doing multiple point-wise path lookups.

I'd noticed that most of my time in mu's cleanup pass consisted of B-tree lookups in Xapian (one 8KB pread64 at a time).  The point lookups forced Xapian to traverse from the root of the B-tree to the leaf for every single message.  Additionally, in order to join on the message path, we had to do *another* B-tree traversal after locating each message term.  Now we just walk the terms in order, which is much more efficient, as we touch each B-tree node only once.

On my system, with 1371861 total messages, the total time of mu index (no lazy check):

--nocleanup: 3.6s
incremental cleanup: 4.2s (0.6s in cleanup)
legacy cleanup: 5.2s (1.6s in cleanup)

With the new mode, we save 1.0s of the 1.6s cleanup, so we're ~63% faster.

But the incremental cleanup works even better with lazy checking. If I enable --lazy-check, dirty only my INBOX (360778 messages), and run index, I get:

--nocleanup: 0.9s
incremental cleanup: 1.1s (0.2s in cleanup)
legacy cleanup: 2.5s (1.6s in cleanup)

We save 1.4s out of 1.6s for ~88% speedup.

This change also fixes a timestamp bug: we should be storing the *start* time of the index pass in metadata, not the end time, so that on the next index pass, we notice messages that arrived between the two times.

All tests pass.  You can set the environment variable MU_NO_INCREMENTAL_CLEANUP to use the legacy cleanup path instead.

### DIFF
--- a/lib/mu-store.cc
+++ b/lib/mu-store.cc
@@ -477,7 +477,11 @@ Store::remove_messages_by_term(std::span<const std::string> terms,
 		enq.set_docid_order(Xapian::Enquire::ASCENDING);
 		enq.set_query(Xapian::Query{Xapian::Query::OP_OR, qvec.begin(), qvec.end()});
 		auto mset = enq.get_mset(0, xapian_db().size());
-		ids_to_remove.insert(ids_to_remove.end(), mset.begin(), mset.end());
+		// Work around Xapian MSetIterator trait mismatch:
+		// https://trac.xapian.org/ticket/850
+		ids_to_remove.reserve(ids_to_remove.size() + mset.size());
+		for (auto it = mset.begin(); it != mset.end(); ++it)
+			ids_to_remove.push_back(*it);
 	}
 
 	// Sort the IDs to remove to make Xapian tree traversal easier


djcb:
Ah, nice work, thank you! I'll take a closer look in the next days.

djcb (commented):
Overall, looks good!

I'm not too familiar with ranges yet, so a good reason to learn a bit!

Few question (see review).

This also currently does not compile on the MacOS CI build:
   https://github.com/djcb/mu/actions/runs/22084626928/job/63816567686

djcb:
Okay, I have merged this now, thank you!

dcolascione:
Oh, thanks! I was going to get around to making those changes.

PR #2919, opened 2026-04-17
fix typo  'we mu uses....'

jasonblewis:
fix type in man page. 

it said 'we mu uses...' I presume it meant just "mu uses"


djcb:
Merged, thanks.

PR #2930, opened 2026-05-25
Remove unnecessary eager loads of ido

bcc32:
To save a bit of time on startup.

mu4e-actions.el doesn't mention ido.

mu4e-helpers.el does, but `ido-completing-read` is autoloaded anyway.


djcb:
Merged; thanks.

PR #2933, opened 2026-06-08
Fix mu tag parsing from Keywords header

### LINKED ISSUE #2932: [mu bug] Keywords header not parsed as a comma-delimited list
**Describe the bug**

I have emails with multiple tags in the Keywords header. mu parses them as a single tag. 

 This is because mu incorrectly expects the list to be space delimited instead of comma-delimited (see [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.6.5) also #769)

Line 319-321 of lib/message/mu-message.cc:
```
	constexpr std::array<std::pair<const char*, char>, 3> tag_headers = {{
		{"X-Label", ' '}, {"X-Keywords", ','}, {"Keywords", ' '}
	}};
```

**How to Reproduce**

This occurs for any message with a valid Keyword header containing more than one tag in a comma-delimited list.

[mu-keywords-example.txt](https://github.com/user-attachments/files/28727754/mu-keywords-example.txt) contains `Keywords:first,second`:
 
```
$ mu find "tags:first"
no matches for search expression

$ mu find "tags:second"
no matches for search expression

$ mu find "tags:first,second"
Mon Jun  8 13:00:00 2026 me@example.com Tag parsing
```

**Environment**
- Fedora 44 Silverblue
- mu 1.14.1 (from homebrew)

**Checklist**
- [X] Tested against 1.14.1

Thanks!

libanp:
Split tags in the Keywords field on commas instead of spaces

Fixes #2932

### DIFF
--- a/lib/message/mu-message.cc
+++ b/lib/message/mu-message.cc
@@ -317,7 +317,7 @@ static std::vector<std::string>
 extract_tags(const MimeMessage& mime_msg)
 {
 	constexpr std::array<std::pair<const char*, char>, 3> tag_headers = {{
-		{"X-Label", ' '}, {"X-Keywords", ','}, {"Keywords", ' '}
+		{"X-Label", ' '}, {"X-Keywords", ','}, {"Keywords", ','}
 	}};
 
 	std::vector<std::string> tags;


dcolascione:
Wouldn't this need a reindex?

libanp:
It would, yes

dcolascione:
So shouldn't the PR bump the index version?

libanp:
Ah, yes, sorry about that. I've bumped the schema version.

djcb:
Thanks.

This was fixed back in 2016, but then the fix was lost again in some re-factor., a few years back.
Can you roll-back the index-version bump? There's no real need for that, and don't want to inconvenience users.
(I'll add a note to the release notes)

djcb:
Okay, merged the fix (minus the schema bump plus a test). Thanks for the fix!

PR #2942, opened 2026-08-12
mu4e: guard modeline truncate-string-to-width against NS DnD crash

rhgg2:
This apparently pointless fix guards against an insta-crash occasioned by `mu4e-modeline` during drag-and-drop on Darwin.

The control flow in `truncate-string-to-width` involves raising and catching an exception whenever the string is shorter than the target (i.e., the function has nothing to do). Which is generally fine, except that drag-and-drop on Darwin (NS) invokes a re-render while Emacs is idling with `waiting_for_input` set. In particular, this re-renders the modeline installed by `mu4e`, whereupon `truncate-string-to-width` raises an exception, and an exception raised during `waiting_for_input` goes straight to a hard crash. 

This is not really `mu4e`'s problem, but the fix is harmless enough that it seems worth doing: ensure that `truncate-string-to-width` is only called
when it is actually needed.

### DIFF
--- a/mu4e/mu4e-modeline.el
+++ b/mu4e/mu4e-modeline.el
@@ -80,7 +80,9 @@ The string is truncated to fit if its length exceeds
 `mu4e-modeline-max-width'."
   (replace-regexp-in-string
    "%" "%%"
-   (truncate-string-to-width str mu4e-modeline-max-width 0 nil t)))
+   (if (> (string-width str) mu4e-modeline-max-width)
+       (truncate-string-to-width str mu4e-modeline-max-width 0 nil t)
+     str)))
 
 (defvar mu4e--modeline-item nil
   "Mu4e item for the global-mode-line.")


djcb:
Huh, that's a weird problem. So it would still crash if the string is too long?
Is there an upstream ticket for this?

rhgg2:
No it crashes precisely when the string is not too long! (Which is basically always).
The function (in `mule-util.el`) does this:
```
(condition-case nil
    (while (< column end-column)
      (setq last-column column
	    last-idx idx
	    ch (aref str idx)
	    column (+ column (char-width ch))
	    idx (1+ idx)))
  (args-out-of-range (setq idx str-len)))
```
So it iterates the string until it either runs out of budget, or goes off the end of the string, and catches the latter case as an exception.
Stylistically questionable, but perfectly fine code.

After your prodding, I checked a bit more and the fault seems to lie firmly with the core emacs NS module, which calls the redisplay function during a drag-and-drop when emacs is in a mode where it will crash on any exception, including the one above. I'm trying to get my ducks in a row for a bug report, but it looks like the fix already has precedent in the same file. Maybe give me a few days to see if I can make any progress there, silly to guard downstream if it can be fixed at the root!

PR #2943, opened 2026-08-14
mu: sort children of a thread based on their own date only

jbrun3t:
This is a revival of commit c4ccaf0fdbd0:
"mu: sort childs of thread based on the sortfield only", whose effect was apparently lost during the rewrite to C++.

Today when we query a find cmd with the `--threads` option, all the children of each thread are sorted according to the most recent message in the thread.

This patch changes how the children of a thread are sorted. Threads are still sorted according to the thread date, but the children of each thread are now sorted by their own date only.

Here is an example of what happened with the previous sorting:

Example with random kernel thread sorted by date:

[PATCH 0/4] drm/panel: jh057n0090: Add regulators and drop magic value in init
  ┣━▶[PATCH 1/4] MAINTAINERS: Add Purism mail alias as reviewer for their devkit's panel
  ┣━▶[PATCH 2/4] drm/panel: jh057n0090: Don't use magic constant
  ┣━▶[PATCH 3/4] dt-bindings: display/panel: jh057n0090: Document power supply properties
  ┗━▶[PATCH 4/4] drm/panel: jh057n0090: Add regulator support

If someone replies to one of these emails in the middle, its date will bubble-up and the sorting becomes:

[PATCH 0/4] drm/panel: jh057n0090: Add regulators and drop magic value in init
  ┣━▶[PATCH 2/4] drm/panel: jh057n0090: Don't use magic constant
  ┃  ┗━▶ Re: [PATCH 2/4] drm/panel: jh057n0090: Don't use magic constant
  ┣━▶[PATCH 1/4] MAINTAINERS: Add Purism mail alias as reviewer for their devkit's panel
  ┣━▶[PATCH 3/4] dt-bindings: display/panel: jh057n0090: Document power supply properties
  ┗━▶[PATCH 4/4] drm/panel: jh057n0090: Add regulator support

With this patch, we will have the following output:

[PATCH 0/4] drm/panel: jh057n0090: Add regulators and drop magic value in init
  ┣━▶[PATCH 1/4] MAINTAINERS: Add Purism mail alias as reviewer for their devkit's panel
  ┣━▶[PATCH 2/4] drm/panel: jh057n0090: Don't use magic constant
  ┃  ┗━▶ Re: [PATCH 2/4] drm/panel: jh057n0090: Don't use magic constant
  ┣━▶[PATCH 3/4] dt-bindings: display/panel: jh057n0090: Document power supply properties
  ┗━▶[PATCH 4/4] drm/panel: jh057n0090: Add regulator support

Cc: Julien Masson <massonju.eseo@gmail.com>

### DIFF
--- a/lib/mu-query-threads.cc
+++ b/lib/mu-query-threads.cc
@@ -80,6 +80,13 @@ struct Container {
 
 	std::string thread_date_key;
 
+	// The container's own date-key, set once from its query-match and never
+	// overwritten by the bubble-up above. Siblings must sort by this, not by
+	// thread_date_key -- otherwise a subthread gaining a new, later message
+	// (e.g. a reply to one message in a patch series) reorders it among its
+	// siblings, even though its own date didn't change.
+	std::string own_date_key;
+
 	Option<QueryMatch&> query_match;
 	bool                is_nuked{};
 	Container*          parent{};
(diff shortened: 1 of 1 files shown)


djcb:
Thanks! 

Looks good at a quick glance, will hopefully have time for a deeper-dive in the next days.

djcb:
Okay merged, with a few follow-tweaks. Thank you!

jbrun3t:
Thanks

PR #2948, opened 2026-09-10
mu4e: propertize headers before hiding them

mrvdb:
Fix for error: 

```
Cannot syntax-propertize 1..349 because of narrowing!
```

on every compose new message.

Emacs hit the same thing in its own `message-hide-headers` and fixed it in [bug#81035](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=81035) (reported for `C-u m` in a Gnus Group buffer, fixed by Stefan Monnier, commit `cc68545f`). Since mu4e carries its own copy, it never picked up that fix.

This fix is basically the same; propertize up to `end-of-headers` while the buffer is still wide.

Verified with `mu4e-compose-new` as well, on mu 1.14.4-pre2 with Emacs 32.0.50.

### DIFF
--- a/mu4e/mu4e-helpers.el
+++ b/mu4e/mu4e-helpers.el
@@ -713,6 +713,9 @@ which we need to avoid #2661."
                   (save-excursion
                     (goto-char end-of-headers)
                     (insert-before-markers header))))))))
+      ;; `syntax-propertize' can't widen, so make sure it won't need to
+      ;; (Emacs bug#81035).
+      (syntax-propertize end-of-headers)
       (narrow-to-region end-of-headers (point-max)))))
 
 (defun mu4e-key-description (cmd)


djcb:
I have never seen that error.... do you have some minimal steps to reproduce (in mu4e)?

mrvdb:
Just starting a compose, so either 'C' in the main menu or `mu4e-compose-new`

I can't recall when it started, just noticed it today. I found the emacs bug report on it and compared code.


For sanity, I just commented the line in the patched locally here and the error re-appears indeed.

djcb:
Can you reproduce from `emacs -Q` and just some minimal mu4e setup? 
And/or can you share a backtrace of the problem (in mu4e)?

Thanks.

mrvdb:
Reproduced from `emacs -Q` + throwaway mu4e setup based on `master`. Did fresh `mu init`/`index` in tmp maildir).

Note that it is a `message`, not an `error`, so `debug-on-error` won't catch it; `syntax.el` emits it only since commit `e6873462ce1` (Monnier, 2026-04-19, `master`/`emacs-31`).

Correction to my PR text: the Emacs fix is `bbe805e4b14` ("message.el: Fix part of bug#81035"), not the SHA I mentioned

It's guarded by `(> syntax-propertize--done 0)`, so it needs something to have propertized the header area before we narrow. In my setup, that is  probably `(use-hard-newlines t 'always)` in `message-mode-hook` properties from `point-min`, then our `insert-before-markers` flushes `syntax-propertize--done` to 1, then `narrow-to-region end-of-headers` strands it.

Reproducing, on top of a minimal mu4e config:

```elisp
(add-hook 'message-mode-hook (lambda () (use-hard-newlines t 'always)))
(setq debug-on-message "Cannot syntax-propertize")
```

`M-x mu4e-compose-new`:

```
Cannot syntax-propertize 1..124 because of narrowing!
  syntax-propertize(397)
  syntax-ppss(397)          ; here forced; normally fontification/show-paren/sexp motion
```

Swapping only `mu4e-helpers.el` in that sandbox: upstream leaves `syntax-propertize--done` at 1 with `point-min` at 124, hence the message; with this PR it is past the narrowing and nothing is emitted.

djcb:
Thank you! Was able to repro, merging this.

PR #2950, opened 2026-09-11
Fix header buffer mis-alignment line in Emacs 31+

ccseidel:
Emacs 31 introduced new faces 'header-line-active' and 'header-line-inactive'. The fix to header buffer misalignment suggested in the manual remaps 'header-line' only, so headers are misaligned again when the header buffer is inactive. Added a two-liner in comments for Emacs 31+ compatibility regarding header line face remapping..


djcb:
Merged, thanks!

PR #2951, opened 2026-09-12
Use draft path for set-parent-flags in compose-message-sent

thoradam:
When `mu4e-sent-messages-behavior` is set to `delete`, the composed message has no `Fcc` field and `mu4e--compose-message-sent` does not call `mu4e--set-parent-flags` and Replied/Forwarded flags are not set on the parent message. Since 413c03ea drafts are always saved before sending so we can rely on `(buffer-file-name)` instead of `(message-field-value "Fcc")`.


djcb:
Ah, that makes sense.

djcb:
Merged with a small cleanup, thanks!

Issue #2952, opened 2026-09-14
mu4e main-view crashes with CJK bookmark names: wholenump, -4 (length vs string-width)


haifenghan:
### Describe the bug

The mu4e main view crashes on startup with `Wrong type argument: wholenump, -4` when any bookmark (or maildir shortcut) name contains CJK (double-width) characters.

### Environment

- mu / mu4e 1.14.3 (Homebrew, macOS aarch64)
- Emacs 31/32
- Any locale, once a `mu4e-bookmarks` entry has a CJK `:name`

### Reproduce

```elisp
(setq mu4e-bookmarks
      '((:name "未读（全部）"       ;; length 6, string-width 12
         :query "flag:unread AND NOT flag:trashed"
         :key ?u)
        (:name "Gmail 未读"         ;; length 8  -> wins the max-length contest
         :query "maildir:/gmail AND flag:unread"
         :key ?g)))
```

Then `M-x mu4e` → `Wrong type argument: wholenump, -4` (here: `8 - 12 = -4`).

### Root cause

Two different width metrics are used for the same alignment computation in `mu4e-main.el`:

1. `mu4e--main-redraw` (around line 311) computes `max-length` using **`length`** (character count):

```elisp
(max-length (seq-reduce (lambda (a b)
                          (max a (length (plist-get b :name))))
                        (mu4e-query-items) 0))
```

2. `mu4e--main-items` (around line 263) pads using **`string-width`** (display width):

```elisp
(make-string (- max-length (string-width name)) ?\s)
```

For CJK names `string-width` ≈ 2 × `length`, so the padding argument goes
negative as soon as the `length`-based max is smaller than some name's
display width → `make-string` rejects the negative count.

Even when it does not crash, double-width names never align correctly,
since the two metrics disagree.

### Suggested fix (one line)

Use `string-width` consistently in `mu4e--main-redraw`:

```elisp
(max-length (seq-reduce (lambda (a b)
                          (max a (string-width (plist-get b :name))))
                        (mu4e-query-items) 0))
```

For pure-ASCII names the two metrics are identical, so this is a safe
change.

### Workaround currently in use

```elisp
(define-advice mu4e--main-items (:around (fn item-type _max-length) cjk-width)
  (let ((real-max (seq-reduce
                   (lambda (a b) (max a (string-width (plist-get b :name))))
                   (mu4e-query-items) 0)))
    (funcall fn item-type real-max)))
```

Happy to turn this into a PR if that's preferred.

[...]

PR #2953, opened 2026-09-14
mu4e--main-redraw: use string-width for aligning item names

haifenghan:
As discussed in #2952.

One-line fix: compute the maximum item-name width with `string-width` instead of `length`, matching the metric already used for padding in `mu4e--main-items`. Identical behavior for pure-ASCII names; fixes the `wholenump` crash and misalignment for CJK (double-width) bookmark/maildir names.

Byte-compile clean.


djcb:
Merged, thanks!

PR #2955, opened 2026-09-15
Fix end of completion region returned by address CAPF function

### LINKED ISSUE #2946: emacs31 and address completion
I used to run emacs30.2 and TAB-completion of addresses worked well. I've upgraded to emacs31 and 31.1 and address completion now works for some addresses, but not others.

If it works, "fo{TAB}" will jump to "foo@whatever", or show a buffer of completions below.

If it doesn't work, "b{TAB}" inserts a ">" and puts the cursor at location "|" as in "b|>". Further {TAB} key presses insert additional ">" below after the cursor, so 5 tabs and I see "b|>>>>>".

+ mu4e - mu for emacs version 1.14.3
+ emacs: GNU Emacs 31.1.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.49, cairo version 1.18.4) of 2026-09-04
+ Debian Trixie.

I've tested this with the following MWE:

```elisp
(setq package-user-dir (expand-file-name "elpa" "~/.emacs.d/DEBUG"))

(require 'package)
(setq package-enable-at-startup nil)

(add-to-list 'package-archives '("melpa" . "http://melpa.org/packages/") t)
(add-to-list 'package-archives '("org" . "http://orgmode.org/elpa/") t)

(package-initialize)
;; Bootstrap `use-package'
(unless (package-installed-p 'use-package)
  (package-refresh-contents)
  (package-install 'use-package))
(setq package-enable-at-startup nil)

(defvar use-package-verbose t)
(require 'use-package)
(setq use-package-always-ensure t)

  (use-package mu4e
    :load-path "/home/kdm/local/mu/share/emacs/site-lisp/mu4e/"
    :bind (("M-M" . mu4e)
	   ("M-r" . mu4e-update-mail-and-index))  
    :config 
    (setq mu4e-get-mail-command "/usr/bin/mbsync --all")
    (setq mu4e-change-filenames-when-moving t)
    (setq mu4e-index-lazy-check t)
    )
```

catern:
Text after point is part of completion too, it's used for filtering and new text can be added after point.  It is buggy to simply end the completion region at point.

This fixes https://github.com/djcb/mu/issues/2946


djcb:
I don't quite understand, the bug seems to be due to Emacs changing its behavior:

```elisp
 (let ((completion-styles '(substring)))
     (completion-try-completion "a" '("axb" "ayb") nil 1))
   ;; emacs-30.2 => ("a" . 1)     --> as expected
   ;; emacs-31.1 => ("ab" . 1)    --> spurious "b"
   ;; (and current master)
```
so this proposed change works around that?

catern:
The behavior of the substring style changed, but this has always been a possible output from a completion style, which this mu4e code has always had a bug with.

djcb:
Hmm, whether or or text after point is part of the completion or not (and where is that documented?), this is a bug / compatibility break in Emacs; this has worked for years. 

And while I appreciate the PR, Emacs shouldn't break compatibility like that.

djcb:
Merged, thanks, closing this.

PR #2956, opened 2026-09-17
mu4e.texi: fix context-switch keybinding

cheon13:
The composer keybindings table documents `C-c C-;` for "switch the context",
but `mu4e-compose.el` binds `mu4e-compose-context-switch` to `C-c ;`. The
documented sequence is not bound anywhere, so following the manual gives
`C-c C-; is undefined`.

Padding adjusted to keep the table's column alignment.


djcb:
Ah, good catch! Merged, thanks.

